PR #18, opened 2021-11-24
Add Wrap view, tweak Nav layout

Atmosfearful:
- Removed Redeem button from nav to make room for Wrap button
- added wrap and unwrap actions and state
- moved disconnect/connect wallet button to top right on desktop views
- more compact, horizontal nav in mobile views
- slightly increased size of main cards

image


ArchimedesCrypto:
Looks good, LFG

Issue #17, opened 2021-11-23
Trim and cleanup ABI JSON files


Atmosfearful:
These were exported from hardhat or elsewhere and likely don't need all the metadata, AST, legacy-AST and other fields. Trimming these should improve load and build times by a few ms

Export json objects which only contain two properties:
```
// @klimadao/lib/abi/myContract.json
{
"contractName": "myContract",
"abi": [ ... ]
}
```
Update file names so they all match the contract name, except for generics like IERC20.json

ABIs can be copy-pasted from polygonscan.

johnCbogil:
I've opened a pr for this here: https://github.com/KlimaDAO/klimadao/pull/19

Please lmk if there's anything else I can do to support 🙂

Atmosfearful:
Closing this, thank you for your help!
https://github.com/KlimaDAO/klimadao/commit/eb84466d76345c6991efb7c2c8026925fcab4ce1

PR #19, opened 2021-11-26
#17 - Trim and cleanup ABI JSON files

johnCbogil:
### This change addresses issue https://github.com/KlimaDAO/klimadao/issues/17

✅ 

> Update file names so they all match the contract name, except for generics like IERC20.json

I didn't see any filenames to update but please lmk if I missed one

> ABIs can be copy-pasted from polygonscan.

I didn't understand this suggestion. Instead of copying from Polygonscan, I edited each file manually. 

### Notes:
I'm a big noob so please lmk if there's anything else I can do to improve this, I'm eager to help 😄


Atmosfearful:
@johnCbogil This is great, thank you so much for jumping in. Huge reduction of lines of code.
Just going to test a bit before merging later today.

PR #26, opened 2021-12-05
added hover tooltips to labels in the dapp views

### LINKED ISSUE #12: Add hover tooltips to labels in the dapp views
- [ ] Add library, I strongly recommend [tippy.js](https://atomiks.github.io/tippyjs/) for a11y and dev happiness.
- [ ] Create a `TextInfoTooltip` component with Emotion stylesheet, exported from `@klimadao/lib/components`. The TextInfoTooltip component wraps any element and shows the provided text on hover.
- [ ] Add an icon-button and tooltip for any item which merits explanation (see below)
- [ ] Info icon-button likely needs alt text because it doesn't have a visible label
- [ ] Screen-readable aria labels are correctly appended to the target element (tippy should do this out of the box)
- [ ] Double check mobile views (tap to trigger tooltip)
```
Balance: Amount of unstaked KLIMA in your wallet
Staked: Amount of KLIMA tokens already staked by you and receiving rebase rewards 
Time Until Rebase:  Approximate time left before the next round of rebase rewards
Next Rebase:  Percentage increase in your staked KLIMA at the next rebase
5 day ROI:  Percentage increase in (or "Return on Investment" on) your staked KLIMA after 5 days, assuming the current reward rate stays the same over the 5 day period and taking into account the effect of compounding. 
APY:  Percentage increase in your staked KLIMA after 1 year ("Annual Percentage Yield"), assuming the current reward rate stays the same over the 1 year period and taking into account the effect of compounding.  PLEASE NOTE THAT THE CURRENT REWARD RATE MAY CHANGE AND IS NOT GUARANTEED TO STAY THE SAME OVER THE 1 YEAR PERIOD.
Current Index: Amount of staked KLIMA you would have now if you had staked 1 KLIMA from protocol inception.
```

Vienio99:
used react-tooltip instead as there were some problems with tippyjs-react and I was unable to came up with a solution. maybe it's good to use it at least for a temporary solution until someone comes up with a fix #12

### DIFF
--- a/app/components/views/Stake/index.tsx
+++ b/app/components/views/Stake/index.tsx
@@ -1,357 +1,357 @@
-import React, {FC, useState} from "react";
-import {useSelector} from "react-redux";
-import {ethers, providers} from "ethers";
+import React, { FC, useState } from "react";
+import { useSelector } from "react-redux";
+import { ethers, providers } from "ethers";
 
 import {
-    changeApprovalTransaction,
-    changeStakeTransaction,
+  changeApprovalTransaction,
+  changeStakeTransaction,
 } from "actions/stake";
-import {useAppDispatch} from "state";
-import {incrementStake, decrementStake, setStakeAllowance} from "state/user";
+import { useAppDispatch } from "state";
+import { incrementStake, decrementStake, setStakeAllowance } from "state/user";
 import {
-    selectAppState,
-    selectBalances,
-    selectStakeAllowance,
+  selectAppState,
+  selectBalances,
+  selectStakeAllowance,
 } from "state/selectors";
-import {TxnStatus} from "actions/utils";
+import { TxnStatus } from "actions/utils";
 
-import {Spinner, TextInfoTooltip} from "@klimadao/lib/components";
+import { Spinner, TextInfoTooltip } from "@klimadao/lib/components";
 import {
-    secondsUntilBlock,
-    prettifySeconds,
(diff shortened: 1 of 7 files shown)


Atmosfearful:
@Vienio99 Are you using Prettier + VSCode? I'm confused by the formatting/spacing going on in the diff. Please rectify if you can.

Also surprised that this adds 3k+ lines of code to the package-lock. Seems like an excessive number of sub-dependencies. I'm going to compare this to tippy and see which is preferable in terms of bundle-size.

Atmosfearful:
A few visual notes:
Let's add some opacity to the icon, so it doesn't distract from the actual label. `opacity: 0.4` looks good.
To the parent, add `align-items: center` and `gap: 0.4rem`, so it centers nicely. Here is screenshot of before/after for reference:

image

And for the tooltip text, I wonder if `caption` typography  would look better than `body2` typography. Either way the line-height is a bit tall which hurts readability.

Vienio99:
@Atmosfearful I've fixed mentioned issues, looks way better now. When it comes to dependencies, indeed react-tooltip has it more unfortunately

Atmosfearful:
@Vienio99 Thanks it looks perfect on my end!!
I decided I am going to copy paste your code to my branch and install the deps myself (just to avoid the security risk as I can't visually audit all those package-lock changes).

Feel free to DM me on discord and we can get you set up on another ticket, if you want :) ❤️

Issue #29, opened 2021-12-06
Add internationalization framework


0xAurelius:
In order to facilitate the community team adding localizations for as many languages as possible, we need to add a framework for internationalization such as `lingui.js`

This blog post gives a good overview of how to set this type of thing up: https://blog.logrocket.com/complete-guide-internationalization-nextjs/

We should also have a PR template specifically for translators to make sure they have a step-by-step include all necessary changes

Atmosfearful:
We need to wait until the new site and brand refresh is completed done before doing any i18n. I do think we need to get there at some point, but my experience has been that it creates a ton of overhead and can significantly slow down pace of development. I don't think it will add much value until we also have marketing and advertising teams creating content and reaching an audience in those languages.

- Adds QA overhead: testing the feature with different word sizes, and RTL layouts.
- A more rigid translation and release pipeline needs to be set up (feature to staging -> translation to staging -> QA-> release).
- I need to add static generation for locale paths (not that hard).

I only have experience with i18n in two apps: one was a megacorp using a bespoke in-house i18n pipeline. Strings were kept in a set of JSON files and translations were committed by the translation team directly to staging before releases. The other was a smaller mobile app with a 4 person team. There we used lokalise and fetched translations via API before release. That app didn't have much text.

A bare minimum-viable product here will be to translate the docs or user guides first. With a proper reference guide + browser translation, the vast majority of users should be able to interact with the app

Atmosfearful:
Work is ongoing in the `i18n` branch. Hoping to merge ASAP

PR #43, opened 2021-12-10
Translation framework

biwano:
Integrates a translation framework (linguijs)
Notes:
- The framework is only integrated in the app, not the site.
- This is a very minimal effort (It does not use nextjs internationalization standards, It does not autodetect the user language, It does not save the user preference, The menu to choose the language is ugly)


0xAurelius:
Resolves #29

app/pages/_app.tsx
```diff
@@ -2,9 +2,25 @@ import "@klimadao/lib/theme/colors.css";
...
+import { I18nProvider } from "@lingui/react";
+import { i18n } from "@lingui/core";
+
+import { init } from "../lib/i18n";
 
 function MyApp({ Component, pageProps }: AppProps) {
-  return <Component {...pageProps} />;
+  useEffect(() => {
+    // Activate the default locale on page load
+    init();
```
Atmosfearful, line 15: Does the fact that this only runs after mount cause a flash of un-translated text?

app/package.json
```diff
@@ -2,13 +2,21 @@
...
+    "lingui:extract": "NODE_ENV=test lingui extract",
+    "lingui:compile": "NODE_ENV=test lingui compile",
+    "postinstall" : "npm run lingui:compile"
   },
   "dependencies": {
+    "@babel/core": "^7.16.0",
+    "@lingui/cli": "3.0.3",
+    "@lingui/macro": "3.0.3",
+    "@lingui/react": "3.0.3",
+    "typescript": "4.0.5",
```
Atmosfearful, line 19: `typescript` and `@lingui/cli` should be dev-dependencies

I believe `@lingui/macro` should be as well

app/package.json
```diff
@@ -2,13 +2,21 @@
...
   "private": true,
   "scripts": {
-    "dev": "next dev -p 3001",
+    "dev": "npm run lingui:compile && next dev -p 3001",
     "build": "next build",
     "start": "next start",
     "export": "next export",
-    "lint": "next lint"
+    "lint": "next lint",
+    "lingui:extract": "NODE_ENV=test lingui extract",
```
Atmosfearful, line 10: why `test`?

app/locale/en/messages.po
```diff
@@ -0,0 +1,70 @@
...
+msgid "footer.docs"
+msgstr "docs"
+
+#: components/views/Home/index.tsx:323
+msgid "footer.home"
+msgstr "home"
+
+#: components/views/Home/index.tsx:235
+msgid "header.connectedto"
+msgstr "You are connected to"
```
Atmosfearful, line 34: I could probably answer this question by reading the docs but,
does this mean in order to change a string, we need to change it in the component, but also here in the messages file?

Or is this auto-generated?

app/components/views/Home/index.tsx
```diff
@@ -268,7 +274,7 @@ export const Home: FC = () => {
     <nav className={styles.nav}>
       {chainId === 80001 && (
         <p className={styles.testnet_warning}>
-          ⚠️You are connected to <strong>testnet</strong>
+          ⚠️<Trans id="header.connectedto">You are connected to </Trans><strong>testnet</strong>
```
Atmosfearful, line 277: This is probably fine for now but it raised a question for me:
Generally speaking, is it better to use interpolation for strings like these?
What does that look like in lingui? in other libs I used its something like:
```json
"header.connectedto": "You are connected to {networkName}"
```

Atmosfearful: the rationale being, in another language the noun `testnet` might not appear at the end

app/package.json
```diff
@@ -2,13 +2,21 @@
...
+    "lingui:compile": "NODE_ENV=test lingui compile",
+    "postinstall" : "npm run lingui:compile"
   },
   "dependencies": {
+    "@babel/core": "^7.16.0",
```
Atmosfearful, line 15: are you sure we need babel, doesn't lingui and nextjs have it as a subdependency already?

Atmosfearful (changes requested):
I'm super stoked for this, insanely helpful. I added a few comments and questions, and there are two more issues before I can merge to the i18n branch:

1. (read my comments below before doing this) It looks like you generated some package-locks. There should only be one, at the root, which is a feature of `npm workspaces`. Try deleting the extras, removing the new dependencies from package.json, and reinstalling using the workspace command, from the root klimadao folder:
`npm install --save-dev -w app -w site @lingui/macro @lingui/cli`
and
`npm install -w app -w site @lingui/react`

That should install them, symlink them to the workspaces (so each is only installed once for all packages 🔥 💯 ), and add them to the root package-lock.json.

2. Please rebase on the latest staging and reinstall deps, then run `pretter --write` to format your code. With the new setup, builds will fail if not formatted.

Thank you!!

app/components/views/Home/index.tsx
kargakis: If we can't use interpolation in translations I would say this should be a deal-breaker for lingui.

app/pages/_app.tsx
biwano: I think so. I am a react/nextjs noob so I do not really know the lifecycles etc... Let me know if there is a better place to put init stuff.

app/package.json
biwano: There seem to be a bug with `lingui extract` when running it in production mode in particular setups. This is a workaround
https://github.com/lingui/js-lingui/issues/433

app/locale/en/messages.po
biwano: You need to change it only in the `messages.po` files. The strings in the component are only default values.

app/components/views/Home/index.tsx
biwano: Interpolation is definetly better and possible with lingui. I was just being lazy.

app/package.json
biwano: I like to compile translations as part of the build process. But i am okay with commiting js files too.

app/package.json
biwano: Good catch. I was following installation instructions blindly.

biwano:
I found a problem with the t component when translating the stake page. When trying this like the documentation says:
`t({
   id: "msg.refresh",
   message: "Refresh inbox"
})`
The compiler complains about the signature of t. 

I had to use a mix of:
- `defineMessage` (so lingui:extract picks up the translation)
- `i18n` (to do the actual translation)

app/pages/_app.tsx
Atmosfearful: OK we can fix this in a follow-up PR :)

Atmosfearful:
As mentioned in my DM please upgrade the package-lock and then this can merge to the i18n branch.

Atmosfearful (approved):
Nice work, let's merge to the i18n branch, work out the remaining issues/questions, and hopefully get this to production ASAP! 🚢 🚢

Atmosfearful:
Reminder: We will need to do regular rebases of staging to the i18n branch to keep things up to date

PR #47, opened 2021-12-16
links on landing page according to specs

Vienio99:
Issue #14


site/components/pages/Home/index.module.css
```diff
@@ -107,6 +107,17 @@
   gap: 1.6rem;
 }
 
+.lowerStack {
+  display: flex;
+  gap: 1rem;
```
Atmosfearful, line 112: Designers use an 8 pixel grid, so I try to stick to that, or sometimes use increments of 4px (0.4rem)
so lets make this 1.6rem
combined with my other suggestion below it lines up nicely!
```suggestion
  gap: 1.6rem;
```

site/components/pages/Home/index.module.css
```diff
@@ -107,6 +107,17 @@
...
+}
+
+.lowerIconButton {
+  composes: iconButton;
+  gap: 0;
+  padding: 1rem;
```
Atmosfearful, line 118: Looks like this makes the button 44px tall, but the others are 40px.
If we stick to the 8px grid and use 0.8rem padding here, then the button height is 40px 👍 
```suggestion
  padding: 0.8rem;
```

lib/components/Icons/GithubIcon.tsx
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+
+interface Props {
+  className?: string;
+}
+export const GithubIcon = ({ className }: Props) => {
```
Atmosfearful, line 6: Nice, good call passing down the className 💯

Atmosfearful (changes requested):

Atmosfearful:
ah, we also need the links to open in a new tab. I'll fix that and commit it directly. Thanks for taking this ticket! 🎉

Issue #48, opened 2021-12-17
Add banner advising users to check URL


0xAurelius:
We should have a banner that reminds people to check the URL and prints our official URL

Atmosfearful:
Great idea! I'll get this on the designers pipeline too, should be a really quick task.
Do you have any examples of this warning on other dapps besides olympusdao?

I'd suggest:
"⚠️Verify the URL and bookmark this page! **klimadao.finance** is the only official domain"

As a full-width top-of-page banner, that appears once every session, and is dismissible.

Weakness here is, the banner doesn't appear on the phishing site, but it could be good for getting people in the habit I guess.

ladytrekker:
I can take look into that 👋 

As far as I can see there are no session cookie handler so far.
Will have to read up on that first with Next.js

PR #50, opened 2021-12-19
Add Check URL banner

ladytrekker:
## Description

This PR adds a full-width top-page banner to DApp.

The banner is rendered by default on initial load.
The area below the banner is not clickable.
The button "Got it" closes the banner.

Closing the banner is not stored in a browser session cookie as mentioned [here](https://github.com/KlimaDAO/klimadao/issues/48#issuecomment-997192921).
Instead, the value is stored in the local React Component's state.
Therefore the banner is shown again as soon as the page is reloaded/revisited again.
I can imagine that this can be annoying for some users? WDYT?

As for the design, I went for a pretty catchy tone to grab the user's attention before they try to continue reading on the page. Let me know if there are any design guidelines I haven't seen yet.

See here some screenshots:

<p align="center">
banner_desktop
banner_mobile
</p>

### Tickets
closes https://github.com/KlimaDAO/klimadao/issues/48

### DIFF
--- a/app/components/CheckURLBanner/index.module.css
+++ b/app/components/CheckURLBanner/index.module.css
@@ -1,0 +1,39 @@
+@value textButton from "@klimadao/lib/theme/common.module.css";
+
+.bg {
+  position: fixed;
+  background-color: rgba(0, 0, 0, 0.5);
+  top: 0;
+  right: 0;
+  left: 0;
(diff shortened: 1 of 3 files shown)


Atmosfearful (approved):
Perfect!!

PR #51, opened 2021-12-19
Add format-all with prettier script

ladytrekker:
## Description

I saw that some files are not formatted yet with the bare VSCode formatting settings mentioned here: https://github.com/KlimaDAO/klimadao#requirements

Additionally, for people not using VSCode their code would probably be not formatted out-of-the-box when saving a file.

As it is recommended to run `prettier --write` on all files (see [this PR comment here](https://github.com/KlimaDAO/klimadao/pull/43#pullrequestreview-835702684)) I added a script on the root level which does exactly that.
~I also ran the script once and added all updated files.~
~Please take a closer look and let me know if this implementation can stay like that and I did not miss anything.~

If this script can stay like that it would make sense to add a default PR-template in which future devs are reminded to run this script once before opening a PR.

## Update

I rebased with staging again and saw that you added exactly that [2 days ago](https://github.com/KlimaDAO/klimadao/commit/d49b7e55a2487548df762cdd00d4b9386e35e851). 🦐 
=> running the script currently has no effect anymore, because now all files are already formatted correctly

Still, I think that adding a lint and format script on the root level for future PRs and mentioned this in a PR template would make sense. I left the script in the `package.json` and updated the README.

**Feel free to close this PR.**

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,5 +20,8 @@
     "dev-lib": "npm run dev --workspace=lib",
     "dev-site": "npm run build-lib && npm run dev --workspace=site",
     "dev-app": "npm run build-lib && npm run dev --workspace=app"
+  },
+  "dependencies": {
+    "prettier": "^2.5.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -21,8 +21,5 @@
     "dev-site": "npm run build-lib && npm run dev --workspace=site",
     "dev-app": "npm run build-lib && npm run dev --workspace=app",
     "format-all": "prettier --write ./"
-  },
-  "dependencies": {
-    "prettier": "^2.5.1"
   }
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -19,6 +19,10 @@
     "export-app": "cd ./lib && npm run build && cd ../app && npm run build && npm run export",
     "dev-lib": "npm run dev --workspace=lib",
     "dev-site": "npm run build-lib && npm run dev --workspace=site",
-    "dev-app": "npm run build-lib && npm run dev --workspace=app"
+    "dev-app": "npm run build-lib && npm run dev --workspace=app",
+    "format-all": "prettier --write ./"
+  },
+  "dependencies": {
```
kargakis, line 25:
```suggestion
  "devDependencies": {
```

ladytrekker: I removed prettier from the dependencies now. 
See the update in the description above and the mentioned commit [here](https://github.com/KlimaDAO/klimadao/pull/43#pullrequestreview-835702684)
Prettier is available already.

Atmosfearful (commented):
Why is the package-lock changed? Did prettier do that?

ladytrekker:
> Why is the package-lock changed? Did prettier do that?

I think with installing - deinstalling it happened.

I rebased again => no update on package-lock no more

PR #53, opened 2021-12-19
Increase rounding precision of next rebase value

### LINKED ISSUE #52: Increase precision when rounding "next rebase value"
Currently the next rebase value rounds to 2 decimals places, which means anyone with less than like 10 KLIMA will see "0"

We should update this to round with more precision, maybe 4-6 places?

https://github.com/KlimaDAO/klimadao/blob/c267873f99448bfc1d81c96165ab8a520092ae69/app/components/views/Stake/index.tsx#L322

0xAurelius:
Resolves #52


kargakis (approved):

Atmosfearful (approved):
thanks ❤️

PR #55, opened 2021-12-27
Show connected address in connect button

kargakis:
This is a UX improvement that catters to power users.
Personally I have been frustrated by having to look in
Metamask in order to figure out which address was connected
in the dapp. I haven't found the disconnect action to be of
much use elsewhere, though I understand it may be something
that newcomers expect to find in this whole Connect flow.


kargakis:
Nevermind, just realized the address is already in the dapp, just a different place

PR #56, opened 2021-12-29
mobile menu

borbesaur:


app/components/views/Home/index.tsx
```diff
@@ -121,17 +122,37 @@ const useProvider = (): [
   ];
 };
 
+const useIsMobile = (): { isMobile: boolean } => {
```
Atmosfearful, line 125: Can we just use CSS breakpoints to show/hide the `.mobileMenu` class?

app/components/views/Home/MobileMenu/index.tsx
```diff
@@ -0,0 +1,73 @@
+import { FC, useState } from "react";
+import { Link } from "react-router-dom";
+import { IconButton, Menu, MenuItem, Zoom } from "@mui/material";
```
Atmosfearful, line 3: Please use this import syntax, which iirc has better tree shaking.

```ts
import IconButton from '@mui/material/IconButton';
```

Atmosfearful (changes requested):
Awesome work!
Just a couple changes.

- [ ] To align with the typography system, can we use `typography.button` styles on the mobile menu item buttons?
- [ ] I think its worth tweaking the icon so it appears (visually) aligned with the card horizontally, and centered on the logo vertically

Untitled

app/components/views/Home/index.tsx
borbesaur: no prob

app/components/views/Home/MobileMenu/index.tsx
borbesaur: done

Atmosfearful (approved):

PR #59, opened 2022-01-02
DRAFT: I18n

biwano:
- Used time-delta for prettifyseconds localization


app/state/app/index.ts
```diff
@@ -18,6 +19,7 @@ const initialState: AppState = {
   stakingRebase: undefined,
   treasuryBalance: undefined,
   rebaseBlock: undefined,
+  showCheckURLBanner: true,
```
Atmosfearful, line 22: @biwano hmm If changing the language causes the banner to reappear, this means that the `Home` component is fully unmounting and remounting.

A full remount miiight cause problems with other states and `useEffects` too (fetching app data etc, etc,). We should be able to just switch the language without remounting everything...

Plus it would be nice to keep single-use states like this out of redux.

Can you investigate:
1. confirm _app.tsx component is _not_ remounting, and confirm that `Home` and other child components are remounting? This means it's the i18n provider causing the re-mount.
2. see if we can prevent it somehow?

Atmosfearful (commented):
I updated i18n to latest staging, please rebase to resolve conflicts then force-push to this branch, thanks!

app/state/app/index.ts
biwano: 1. I confirmed that the i18n provider is causing the re-mount
2. Great advice! I could prevent it this way: 
```    <I18nProvider i18n={i18n} forceRenderOnLocaleChange={false}>
```

Atmosfearful (changes requested):
As discussed on our call:
- add messages.js files to gitignore
- remove the new dependency and fix the UI to not rely on time deltas
- hide language switcher on prod

Then we can merge to production and see if it works on Fleek! And wrap up translations. Awesome work!!

site/pages/index.tsx
```diff
@@ -5,18 +5,24 @@ import {
   getTreasuryBalance,
 } from "@klimadao/lib/utils";
 import { Home, Props } from "components/pages/Home";
+import { loadTranslation } from "@klimadao/site/lib/i18n";
```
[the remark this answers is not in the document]
biwano: This is done.
I had to add the locale to redux to make sure components could hook to locale changes change. Otherwise Localizing the date was not updating at the right time.

Atmosfearful (approved):

Issue #62, opened 2022-01-04
Landing page redesign


Atmosfearful:
# Rough work plan
- [ ] Get figma designs for desktop and mobile (confidential)
- [ ] Implement new theme color variables
- [ ] Create a branch called `site-redesign`
- [ ] Create a new component `site/components/pages/Home` colocated with a new stylesheet `styles.ts`
- [ ] Optimize images using `next/image`
- [ ] Get final text from creative team

## Stretch goals
- [ ] Upload all landing page images to the CMS and implement a custom `loader` for `next/image`. When @rstycedar is finished with the CMS. This will reduce our vercel bandwidth costs.
- [ ] "Buy Klima" section can link to a tutorial. we can integrate sushi later.

Atmosfearful:
@ladytrekker will start work on this! We can work together to flesh out this ticket a bit more if needed.

ladytrekker:
@Atmosfearful I checked some more items above.
Let's talk if this ticket can be split up into smaller single tickets for each missing task.

PR #69, opened 2022-01-08
Redesign: Add Header, Footer and Navigation for Desktop & Mobile

ladytrekker:
## Description

These are the first steps for the new site redesign.

This PR adds:
- Header for Desktop
- Header for Mobile
- Navigation for Desktop
- Navigation and Toggle for Mobile
- basic Footer
- some theme styles

It is not pixel perfect, some styles are still missing but I want to gather some feedback for the first component set-up and folder structure before I continue.

Currently the set-up is as such:
- add all components to /lib
- use emotion/css only
- have in mind that all components have to be re-usable for site and app which are different websites

So far I am repeating styles a lot.
I am thinking of using emotion/react and emotion/styled to avoid that.

**What is your thinking?**

## Related Ticket

Related to https://github.com/KlimaDAO/klimadao/issues/62

## Checklist


lib/components/Buttons/styles.ts
```diff
@@ -0,0 +1,24 @@
...
+export const button_primary = css`
+  box-sizing: border-box;
+  margin: 0px;
+  min-width: 0px;
+  appearance: none;
+  display: inline-block;
+  text-align: center;
+  line-height: inherit;
+  text-decoration: none;
+  padding: 10px 18px;
```
Atmosfearful, line 12: padding, margin, borders, widths and heights should all use `rem` so that they scale up equally as the user increases their device font size

lib/components/index.ts
```diff
@@ -3,3 +3,13 @@ export { WebFonts } from "./WebFonts";
 export { TextInfoTooltip, useTooltipSingleton } from "./TextInfoTooltip";
 export { GithubIcon } from "./Icons/GithubIcon";
 export { TwitterIcon } from "./Icons/TwitterIcon";
+export { PageWrap } from "./Layout/PageWrap";
+export { HeaderDesktop } from "./Header/HeaderDesktop";
+export { HeaderMobile } from "./Header/HeaderMobile";
+export { Footer } from "./Footer/Footer";
```
Atmosfearful, line 9: What's your opinion on `index.tsx` files? I know they are controversial. I don't really care either way but I guess we should pick a pattern to use everywhere

lib/constants/index.ts
```diff
@@ -59,6 +59,9 @@ export const urls = {
   discordInvite: "https://discord.gg/kx4pahaFw8",
   gitbook: "https://klima-dao.gitbook.io/klima-dao/",
   app: "https://dapp.klimadao.finance",
+  stake: "https://dapp.klimadao.finance/#/stake",
+  wrap: "https://dapp.klimadao.finance/#/wrap",
+  bond: "https://dapp.klimadao.finance/#/bonds",
```
Atmosfearful, line 64: nice ty

site/components/pages/Redesign/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+        metaDescription="Drive climate action and earn rewards with a carbon-backed digital currency."
+        mediaImageSrc="/og-media.jpg"
+      />
+      <PageWrap>
+        <HeaderDesktop
+          buttons={[
+            <ButtonPrimary
+              key="Connect Wallet"
+              label="Connect Wallet"
+              onClick={() => console.log("Do we want this here?")}
```
Atmosfearful, line 40: no, no wallet connection anywhere in `site`. I brought this up a few weeks ago but designs never got updated, sorry!

All wallet interactions should be contained in the `app` because `ethers.js` is beefy and because the `app` can be exported as a standalone plain html app. But the rest of the website can not.

site/components/pages/Redesign/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+          <NavItemDesktop url={urls.stake} name="Stake" />
+          <NavItemDesktop url={urls.wrap} name="Wrap" />
+          <NavItemDesktop url={urls.bond} name="Bond" />
+        </HeaderDesktop>
+        <HeaderMobile>
+          <NavItemMobile url={urls.home} name="Home" />
+          <NavItemMobile url={urls.tutorial} name="Buy Klima" />
+          <NavItemMobile url={urls.stake} name="Stake" />
+          <NavItemMobile url={urls.stake} name="Wrap" />
+          <NavItemMobile url={urls.bond} name="Bond" />
```
Atmosfearful, line 55: nice composition 🔥

lib/components/Header/HeaderMobile.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import {
+  LogoWithClaim,
+  ToggleNavButton,
+  NavMobile,
+} from "@klimadao/lib/components";
+import { urls } from "@klimadao/lib/constants";
+
+interface Props {
+  children: JSX.Element[];
```
Atmosfearful, line 12: i dont think you need to type children if you use `FC`

lib/components/Navigation/styles.ts
```diff
@@ -0,0 +1,222 @@
...
+  &:hover {
+    color: var(--surface-04);
+  }
+`;
+
+export const navMain_MobileItemInnerNumber = css`
+  font-size: 2rem;
+  font-weight: 600;
+  color: var(--surface-05);
+  margin-right: 20px;
```
Atmosfearful, line 90: since we are making progress on translation system we have to be ready for RTL languages.

`margin-inline-start` will work on both RTL and LTR

But my personal fav is, where possible, to have parent div space children using grid or flex `gap`

lib/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <footer className={styles.footer}>
+      <div className={styles.footer_content}>
+        <nav className={styles.footer_nav}>
+          <a href={urls.home}>Klima Dao</a>
+          <a href={urls.tutorial}>Buy Klima</a>
+          <a href={urls.stake}>Stake</a>
+          <a href={urls.bond}>Bond</a>
+          <a href={urls.officialDocs}>Docs</a>
+          <a href={urls.discordInvite}>Join Klima</a>
```
Atmosfearful, line 18: docs, and discord should open in new tabs like the github and twitter pls

Atmosfearful (changes requested):
Awesome progress!
People are hyped :)

image

lib/components/Header/HeaderMobile.tsx
ladytrekker: true, removed

site/components/pages/Redesign/index.tsx
ladytrekker: 🔥

lib/components/Footer/Footer.tsx
ladytrekker: done, did the same for "Buy Klima" as this points to the tutorial for now

site/components/pages/Redesign/index.tsx
ladytrekker: Thanks for the update. Changed to "Enter App" as stated in the updated designs

lib/components/index.ts
ladytrekker: Not a super fan of index.tsx but not a big thing either. 
Will stick to that index pattern as this is already started like this
Changed.

lib/components/Buttons/styles.ts
ladytrekker: done

lib/components/Navigation/styles.ts
ladytrekker: in this special case I think it is ok as it targets the number next to the main nav link on the menu item
Will keep in mind flex grid for the future

brendanmc6 (approved):

Atmosfearful (approved):
AWESOME work!

PR #70, opened 2022-01-11
Block rate fix

### LINKED ISSUE #31: Rebase & Bond times are inaccurate
Due to the fact that the DAPP uses a constant for the block rate (see https://github.com/KlimaDAO/klimadao/blob/main/lib/constants/index.ts#L95) which is pinned at `2.5` seconds, and because this is not really how often the block rate actually works, the 'time until fully vested' value on the bond dapp can be fairly inaccurate.

I've seen this value be off by as much as 20 hours. The correct way to do this would be to poll polygonscan's API to check what the actual value should be, instead of hard-coding a constant.

Vienio99:
## Related Ticket

Resolves #31 

## Checklist

### DIFF
--- a/lib/utils/getEstimatedDailyRebases/index.ts
+++ b/lib/utils/getEstimatedDailyRebases/index.ts
@@ -1,0 +1,9 @@
+import { EPOCH_INTERVAL } from "../../constants/index";
+
+// Calculates estimated daily rebases based on current block rate
+export async function getEstimatedDailyRebases(blockRate: number) {
+  const rebaseRate = blockRate * EPOCH_INTERVAL;
+  const secondsInDay = 86400;
+  const rebasesPerDay = secondsInDay / rebaseRate;
+  return rebasesPerDay;
+}
--- a/app/actions/app.ts
+++ b/app/actions/app.ts
@@ -3,7 +3,7 @@ import { ethers, providers } from "ethers";
 import { Thunk } from "state";
 import { setAppState } from "state/app";
 
-import { getTreasuryBalance } from "@klimadao/lib/utils";
+import { getBlockRate, getTreasuryBalance } from "@klimadao/lib/utils";
 import { addresses, ESTIMATED_DAILY_REBASES } from "@klimadao/lib/constants";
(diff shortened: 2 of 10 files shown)


kaxline:
@Atmosfearful @Vienio99 Is there another use for Block Rate besides estimating time to rebase? If not, can't we just get the `endBlock` from the `epoch` and then query the `EstimateTimeInSec` of the `endBlock`?

Atmosfearful:
> @Atmosfearful @Vienio99 Is there another use for Block Rate besides estimating time to rebase? If not, can't we just get the `endBlock` from the `epoch` and then query the `EstimateTimeInSec` of the `endBlock`?

Good point, we can use this to replace the constant values used in `ESTIMATED_DAILY_REBASES` and perhaps in the bonds as well. Me and @Vienio99 spoke about this on a call and he is gonna look into it.

Recap:
```
// number of blocks per rebase
EPOCH_INTERVAL = 11520;
rebaseRate = EPOCH_INTERVAL * blockRate;
rebasesPerDay = secondsPerDay / rebaseRate;
```

lib/utils/getEstimatedDailyRebases/index.ts
```diff
@@ -0,0 +1,9 @@
+import { EPOCH_INTERVAL } from "../../constants/index";
+
+// Calculates estimated daily rebases based on current block rate
+export async function getEstimatedDailyRebases(blockRate: number) {
```
Atmosfearful, line 4: this has no `await` so it doesn't need to be an `async` function. You can remove it out of the promises array here 
https://github.com/KlimaDAO/klimadao/pull/70/files#diff-b96f585d57b9419336b96f03d5d29b05d4b7d5855d5162ab0d73d2cca3d8afbdR59

And remove the `awaits` where used

Atmosfearful (changes requested):
NICE, good call putting it in redux.
Just one quick change

Atmosfearful:
If this has a significant impact on APY I will need to check with coms and policy team before merging

Atmosfearful:
Closing this PR in favor of #90 I guess I didn't need to, oh well 🤷

PR #87, opened 2022-01-16
feat: add ens name and avatar support

bigint:
## Description

add ens name and avatar support

Screenshot 2022-01-16 at 8 04 42 PM

## Checklist

### DIFF
--- a/app/components/views/Home/WalletAction/index.tsx
+++ b/app/components/views/Home/WalletAction/index.tsx
@@ -2,7 +2,7 @@ import { concatAddress } from "@klimadao/lib/utils";
 import useENS from "components/hooks/useENS";
 import { FC } from "react";
 import { WalletProps } from "../constants";
-import { TextInfoTooltip, useTooltipSingleton } from "@klimadao/lib/components";
+import { TextInfoTooltip } from "@klimadao/lib/components";
 
 import styles from "../index.module.css";
 
@@ -12,7 +12,6 @@ const WalletAction: FC<WalletProps> = ({
   loadWeb3Modal,
   disconnect,
 }) => {
-  const [singletonSource, singleton] = useTooltipSingleton();
   const { ensName, ensAvatar } = useENS(address);
   return !isConnected ? (
     <button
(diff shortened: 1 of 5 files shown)


bigint:
cc @Atmosfearful

Atmosfearful (changes requested):
This is awesome. Didn't realize resolving was so easy!

Only downside is now it's not clear in the ui how to 'disconnect'. Can you wrap the disconnect button in a `<TextInfoTooltip>` that says 'disconnect'? should be a quick fix.🙏

Atmosfearful:
I'll make sure this makes it to the new site as well!

Atmosfearful:
When you push up a change the failing `cms` build should (hopefully) fix itself

bigint:
@Atmosfearful Can you review now 🙇🏻 

image

app/components/views/Home/WalletAction/index.tsx
```diff
@@ -17,13 +23,17 @@ const WalletAction: FC<WalletProps> = ({
...
     </button>
   ) : (
-    <button
-      type="button"
-      className={styles.disconnectWalletButton}
-      onClick={disconnect}
-    >
-      DISCONNECT WALLET
-    </button>
+    <TextInfoTooltip content="DISCONNECT WALLET" singleton={singleton}>
```
Atmosfearful, line 26: ah no need for the singleton, thats only when you want to animate/sync up multiple tooltips

Atmosfearful (changes requested):
oops meant to request changes. Just remove that singleton and we good to go! Thanks again!

app/components/views/Home/WalletAction/index.tsx
bigint: @Atmosfearful removed it 👍🏻

Atmosfearful (approved):

bigint:
@Atmosfearful any ideas when it will go to prod? If available there I'll add the side in ens.domains homepage

PR #99, opened 2022-01-20
refactor: specify directories to lint

0xAeterno:
## Description

Follow up on the previous PR.
[Next lint](https://nextjs.org/docs/basic-features/eslint#linting-custom-directories-and-files) will only run the linter against `pages/`, `components/` and `lib/` directories. 
Needed to specify `actions/` and `state/` for proper linting coverage. Something to bear in mind if we add new directories to the repos.

Also added the base eslint rulesets recommended by Next.js

Picked up on this when attempting to share common eslint config across all the repos as we discussed @Atmosfearful . Unfortunately, didn't have much luck with it. Shared config convention seems to be publishing a separate shared config module to npm, which seems overkill at this point in time 🤪 

## Checklist


Atmosfearful:
waiting to merge this after staging goes out later today (new bond launching)

Issue #124, opened 2022-02-01
Cleanup /blog and /[post] styles


Atmosfearful:
Fonts are too big. See the message and link to figma that masasmune posted in frontend channel in discord

PR #126, opened 2022-02-03
Cleanup /blog and /[post] styles

rstycedar:
## Description

*  update typography and colors of /blog and /post page to match designs
*  update join discord card styles to it doesn't get overlapped by the resources menu
*  remove h3 and h4 from sanity schema
*  add body3 to typography and --shadow-07, --surface-04, --surface-05 theme variables
*  add block quote styles to the blog

## Related Ticket

#124 

## Checklist

### DIFF
--- a/lib/theme/variables.css
+++ b/lib/theme/variables.css
@@ -3,8 +3,6 @@
   --klima-blue: #0ba1ff;
   --gray: rgb(160, 160, 160);
   --white: #ffffff;
-  --surface-04: #767676;
-  --surface-05: #a1a1a1;
   --font-family: Inter, sans-serif;
   --border-radius: 0.4rem;
   --header-height: 4.8rem;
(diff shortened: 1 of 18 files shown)


lib/theme/variables.css
```diff
@@ -3,6 +3,8 @@
   --klima-blue: #0ba1ff;
   --gray: rgb(160, 160, 160);
   --white: #ffffff;
+  --surface-04: #767676;
+  --surface-05: #a1a1a1;
```
Atmosfearful, line 7: these need to have dark mode variants and be a part of .light and .dark class.

Since they are only used for text lets call them
```
--font-04
--font-05
```

If they are very very similar to --font-01, 02 or 03 then just use those colors instead and remove them altogether.

Atmosfearful (changes requested):
NICE cleanup! Thanks for the atomic commits and good PR description too.

Just one change

Atmosfearful (approved):

Issue #54, opened 2021-12-27
Dark Mode / Light Mode toggle


Atmosfearful:
The codebase needs a convention for color variables that allows us to easily switch dark & light mode.

Currently I use css variables for all colors, so this shouldn't be too difficult to enhance.
I think we only need to use js to toggle a css class somewhere.

We don't have completed light theme designs yet, but we can improvise until then. We can comment out a UI switch until it's ready.

```css
:root.theme-default {
 --surface: #121212;
 --surface-01: #1e1e1e;
 --primary: #33972d;
 --primary-variant: #33732f;
 --secondary: #14c3b9;
 --secondary-variant: #1ea39b;
}
:root.theme-light {
 --surface: #??????;
 --surface-01: #??????;
 --primary: #??????;
 --primary-variant: #??????;
 --secondary: #??????;
 --secondary-variant: #??????;
}
```

Hopefully don't need to mess with existing CSS too much, but there could be some missing color pairs.

A stretch-goal might be to extract this to a drop-in component, and export it from `lib/components` so it can be used in both `app` and `site`.

```tsx
// app/pages/_app.tsx
<InjectThemeVariables />
```

Toggle button can probably be pretty simple?

```
const ColorModeSwitch = () => {
  const handleClick = () => {
    // set className on root
  }

  return <button onClick={handleClick} className={styles.colorModeSwitch} />
}

```

Atmosfearful:
Let's put this on hold until after site redesign

PR #130, opened 2022-02-04
Add Theme Toggle to Site

ladytrekker:
## Description

This PR adds a Theme Toggle to the Site.

To prevent the "color flash" on server rendering I followed this tutorial:
https://electricanimals.com/articles/next-js-dark-mode-toggle
Let me know if you have a better idea.

I focused only on the styles for the **LandingPage** as I know that people are working on the other pages.

**OPEN TODOS for DARK-THEME:**

- add another Icon for Dark Theme (designs missing, only "Moon" exists)
- fix color values for Resources SubNavigation (designs are missing for DarkTheme, also current colors in Figma do not match the Theme Colors)
- "Our Partner" icons do not work well in dark mode on "Community" page
- fix color values for Blog Page
- fix color values for Blog Single Page

## Related Ticket
Resolves https://github.com/KlimaDAO/klimadao/issues/54

## Checklist


Atmosfearful (approved):
@ladytrekker  Wow this works great!

Pulled locally just to play around and understand how the solution works. It's hacky but neat, I like it and can't think of anything better.

I made 4 small changes while I was in there:
- extract to a component in Lib so we can reuse it for dapp and to keep _document clean.
- add a comment to explain why we do it this way
- refactor the `getUserPreference()` fn to be extra super safe and return the string, rather than returning whatever the browser had saved in `localStorage`. Was probably fine but `dangerouslySetInnerHTML` scares me.
- fix a color problem I spotted in footer

Issue #131, opened 2022-02-06
[site-redesign] Remaining cleanup items


Atmosfearful:
If you are starting work on one of these, leave a comment below please.

* [x] The hover state for the twitter and github logos in the footer make them disappear
* [x] white bg for Partners section so logos look better on dark theme @kaxline 
* [x] fix remaining broken color variables to support dark theme on blog, blogpost, community, and contact pages. @ladytrekker 
* [x] make the learn more + arrow text into an anchor link that scrolls the page to the next section @ladytrekker 
* [x] logo height should be the same as the enter app button its currently too small @ladytrekker 
* [x] higher resolution image for community discord screenshot https://github.com/KlimaDAO/klimadao/pull/147
* [x] larger resolution image for dummyswap.jpg https://github.com/KlimaDAO/klimadao/pull/147
* [x] fix resource nav on mobile small devices   @ladytrekker 
![image](https://user-images.githubusercontent.com/88635679/152921157-834ce3da-b340-434b-aa93-818aa60a89da.png)

PR #147, opened 2022-02-09
Replace images for Buy Klima and Join Discord

ladytrekker:
## Description

Both images are blurry. 
I exchanged both images with the ones provided by design.

The images are more crispy now, but still not perfect.

I also aligned the Discord section on the Community page so that the sticky Resources Sub-navigation is not overlapping this section no more.

## Related Ticket

Part of https://github.com/KlimaDAO/klimadao/issues/131

## Checklist


site/components/pages/Resources/Community/index.tsx
```diff
@@ -182,34 +182,34 @@ export const Community: NextPage<Props> = ({}) => {
...
+                  smart contracts to create
+                </Trans>
+              </Text>
+              <DiscordButton />
+            </div>
+            <div className="joinDiscord_row2">
+              <div className="joinDiscord_dummy">
+                <Image
+                  src={screenShotDesktop}
+                  width={804}
```
Atmosfearful, line 207: hmm on my viewport it looks like the image is never larger than 560 px wide. We can save some bandwidth $ by reducing this

ladytrekker: I reduces the image now https://github.com/KlimaDAO/klimadao/pull/151
The image is now 64KB which is fine IMO
Changing the width and height to 560 x 415.82 made it look even blurrier again in my opinion. 😭  Why NextJS?

PR #168, opened 2022-02-15
feat: add discord icon to site footer

ladytrekker:
## Description

Requested that more social links should be added to the footer.
Right now we do only have the discord icon. Which is the most important IMO.
Let´s add more in the future.

## Checklist


Atmosfearful (approved):
nice, this might overlap with the disclaimer text now tho ill merge and see

Issue #160, opened 2022-02-10
Add podcast, youtube, reddit icons & links to website


Atmosfearful:
And maybe to the dApp as well?

PR #183, opened 2022-02-16
feat/add social links to site

0xAeterno:
Hehe was hoping to get this done before site launch 😅

## Description

Add more social links to site: 
- Youtube
- Reddit
- Telegram
- Twitch
- POTK RSS feed

Include `<title>` tag in svg files to be more accessible.
Updated the footer styling on smaller viewports to accommodate the longer block of social links. 

## Changes
| Screen | Before  | After  |
| ------ |---------|--------|
| Mobile |image|image|
| Tablet | image|image|
| Dapp |image|image|

### TODO: 
- [x] Add links to dapp
- [x] Look into why Disclaimer nav is missing from footer

## Related Ticket

Resolves https://github.com/KlimaDAO/klimadao/issues/160

## Checklist


site/components/Footer/index.tsx
```diff
@@ -2,68 +2,86 @@ import React, { FC } from "react";
...
+        <a href={urls.twitter} target="_blank" rel="noreferrer noopener">
+          <TwitterIcon />
+        </a>
+        <a href={urls.youtube} target="_blank" rel="noreferrer noopener">
+          <YoutubeIcon />
+        </a>
+        <a href={urls.discordInvite} target="_blank" rel="noreferrer noopener">
+          <DiscordIcon />
+        </a>
+        <a href={urls.reddit} target="_blank" rel="noreferrer noopener">
```
Atmosfearful, line 69: I keep saying to myself I should create a `<A>` component that does `target` and `rel` by default :) lets do that in a follow up at some point

lib/components/Icons/MetamaskFox.tsx
```diff
@@ -1,106 +1,105 @@
...
-      x="0"
-      y="0"
-      enableBackground="new 0 0 318.6 318.6"
-      version="1.1"
-      viewBox="0 0 318.6 318.6"
-      xmlSpace="preserve"
-    >
-      <path d="M274.1 35.5L174.6 109.4 193 65.8z" className="st0"></path>
-      <path d="M44.4 35.5L143.1 110.1 125.6 65.8z" className="st1"></path>
+export const MetaMaskFox = () => (
```
Atmosfearful, line 3: thanks for cleaning these up!
Would you mind, since you are in here, aligning all the SVGs to use this pattern:

```tsx
interface Props {
  className: string;
}

const MySVG = (props: Props) => {
  return <svg className={props.className} />
}
```

Namely:
- include a classname prop for each
- [dont destructure props](https://github.com/KlimaDAO/klimadao/wiki#do-not-destructure-props)

Atmosfearful (changes requested):
As far as design, looks good to me! No need to wait for design feedback. More important that we get these links up IMO.

I'm a little concerned that small tap-targets will ding us on Lighthouse. But we can check and improve later.

Just one chore for ya:

Atmosfearful:
@0xAeterno See recent commits to staging to address your merge conflicts-- we updated some of those URLs this morning.

lib/components/Icons/MetamaskFox.tsx
0xAeterno: Easy, I'll spend a bit of time polishing these SVG files to follow our conventions.
Do we have a preference to be explicit with the `return` keyboard?

site/components/Footer/index.tsx
0xAeterno: Thought I might as well include it into this PR since I'm in the space, turned out to be quite the rabbit hole retrofitting the new component 😅

lib/components/Icons/MetamaskFox.tsx
Atmosfearful: neat thanks! I have no preference for `return` keyword

Atmosfearful (approved):
💯

Issue #179, opened 2022-02-15
Links between dapp/site should not open in new tabs


Atmosfearful:

Liftedangifted:
hey I commented on this in Discord! Sup all this is Real

Atmosfearful:
Should be fixed! in 46edf98d1699de42b2075786add23b8363e86005

PR #184, opened 2022-02-16
Fixing 2 issues

mosnicholas:
## Related Ticket

https://github.com/KlimaDAO/klimadao/issues/179
https://github.com/KlimaDAO/klimadao/issues/180

## Checklist


mosnicholas:
Screen Shot 2022-02-15 at 9 13 51 PM

Atmosfearful:
thanks for jumping on these! 💯 
Notes:
- package.lock shouldn't have any changes
- your link fix works but we need a better solution to allow differentiating external vs non-external links in the nav there. I have an idea....
- i think i'll also enhance the balances card to only show wsklima if the user has a wsklima balance

It will be faster if I just port these fixes in myself, so I'll close this PR. Thanks again

PR #189, opened 2022-02-17
Enable Coinbase WalletLink provider

0xAurelius:
## Description

Enables Coinbase WalletLink provider in Web3Modal (just upgrading to the new version doesn't automatically enable it)

## Related Ticket

Resolves #<issue-number>

## Checklist


app/package.json
```diff
@@ -30,6 +30,7 @@
     "react-redux": "^7.2.6",
     "react-router": "^6.1.0",
     "react-router-dom": "^6.1.0",
+    "walletlink": "^2.4.7",
```
Atmosfearful, line 33: looks like package.lock needs to be updated too,
make sure you are on a recent version of npm and install with the workspace flag

```
npm install -w app walletlink
```

Atmosfearful (changes requested):
Nice,

app/package.json
0xAurelius: Ran that and added the lockfile, not sure if it looks reasonable though

0xAurelius:
Updating the lockfile got this working, was able to connect a test wallet to the Dapp and switch to the proper chain

Atmosfearful (approved):

PR #192, opened 2022-02-17
Add canonical tag

rstycedar:
add canonical tag to site Pagehead and make sure it doesn't contain query params or hash params or locale and also works for dynamic blog routes


Atmosfearful (approved):
nice solution!

Issue #190, opened 2022-02-17
Blog post images should fill max width


Atmosfearful:
Bonus points if you can let authors decide how the image looks...

![image](https://user-images.githubusercontent.com/88635679/154523433-98af08c8-000e-41a6-859d-c5d74ef8f7b2.png)

PR #195, opened 2022-02-18
blog image layout to take up fullwidth

rstycedar:
#190


Atmosfearful (changes requested):
Here are some more clear requirements for how images should work:
- constrain width to content area (76rem on desktop). Maybe the css grid can do this for you?
- no constraint on height. if the author uploads a 760x9000 image, it should appear at full resolution

Next layout="responsive" should work, it takes the container width and fits the image inside. https://nextjs.org/docs/api-reference/next/image

PR #200, opened 2022-02-22
feat: link to latest blog post on homepage

0xAeterno:
## Description

Dynamically set the landing page news banner to link to the latest blog post.

## Changes

image

## Checklist

### DIFF
--- a/site/components/pages/Home/index.tsx
+++ b/site/components/pages/Home/index.tsx
@@ -1,15 +1,17 @@
 import React, { useRef } from "react";
 import { NextPage } from "next";
 import Image from "next/image";
+import Link from "next/link";
 import { Trans, t } from "@lingui/macro";
 import ArrowDownwardIcon from "@mui/icons-material/ArrowDownward";
+import { urls } from "@klimadao/lib/constants";
 import { Section, ButtonPrimary, Text } from "@klimadao/lib/components";
 
 import { Footer } from "components/Footer";
 import { Navigation } from "components/Navigation";
 import { PageHead } from "components/PageHead";
 import { IS_PRODUCTION } from "lib/constants";
-import { urls } from "@klimadao/lib/constants";
+import { LatestPost } from "lib/queries";
 
 import forest from "public/forest.jpg";
 import cars from "public/cars.jpg";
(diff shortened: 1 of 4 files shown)


site/components/pages/Home/index.tsx
```diff
@@ -65,16 +68,19 @@ export const Home: NextPage<Props> = (props) => {
...
-            <Text t="button" align="end">
-              <Trans>📰 Recent Post: </Trans>
-            </Text>
-            <Link href="/blog/a-new-face-for-klimadao">
-              <a>
-                <Trans>A New Face For KlimaDAO</Trans>
-              </a>
-            </Link>
-          </div>
+          {props.latestPost && (
```
0xAeterno, line 71: If we fail to fetch data from sanity, instead of raising an error as we do in other places, allow the query to fail silently and don't render the news banner. Any thoughts on this?

site/lib/queries.ts
```diff
@@ -20,6 +20,8 @@ export const queries = {
     '*[_type == "post"] | order(publishedAt desc) {summary, "slug": slug.current, title, publishedAt, author->, "imageUrl": mainImage.asset->url}',
   /** fetch all the blog post slugs, for getStaticPaths */
   allPostSlugs: '*[_type == "post"] {"slug": slug.current}',
+  latestPost:
+    '*[_type == "post"] | order(_createdAt desc){"slug": slug.current, title}[0]',
```
Atmosfearful, line 24: We should sort by the `publishedAt` field which exists on `Post` documents.
They might not be published in the order that they are created!

site/components/pages/Home/index.tsx
```diff
@@ -65,16 +68,19 @@ export const Home: NextPage<Props> = (props) => {
...
+            <div className="hero_newsBanner">
+              <Text t="button" align="end">
+                <Trans>📰 Latest News: </Trans>
+              </Text>
+              <Link href={`/blog/${props.latestPost.slug}`}>
+                <a>
+                  <Trans>{props.latestPost.title}</Trans>
```
Atmosfearful, line 78: you can remove the translate tags

Atmosfearful (changes requested):

site/lib/queries.ts
0xAeterno: That's a good point, nice catch!

site/components/pages/Home/index.tsx
0xAeterno: Is this because sanity will return content in the requested locale?

Atmosfearful: Eventually, yes. Right now it doesn't. But Linguee uses babel macros at compile time-- since the CMs content is at runtime I don't think we can translate it with linguee

Atmosfearful (approved):
pinging @rstycedar , looks good to me

rstycedar (approved):
nice work! great attention to detail

PR #205, opened 2022-02-24
Add KLIMA supply API

### LINKED ISSUE #198: Add current total supply API
In order for CMC to properly display our market cap on their site, we need to provide them with a specific API endpoint that just returns the total supply as a number (and nothing else)

Tried to direct them to use our Subgraph but they don't seem to be that flexible and I can't figure out a way to get a single value out of a GraphQL URL-based query

So we can just create a simple endpoint like `/supply` that fetches the current total supply from the KLIMA contract ABI and returns that single value.

Original request received from CMC:
image

0xAurelius:
## Description
Adds API endpoint for current KLIMA supply

## Related Ticket

Resolves #198 

## Checklist

### DIFF
--- a/lib/utils/getKlimaSupply/index.ts
+++ b/lib/utils/getKlimaSupply/index.ts
@@ -1,0 +1,16 @@
+import { ethers } from "ethers";
+import { getJsonRpcProvider } from "../getJsonRpcProvider";
+import { addresses } from "../../constants";
+import IERC20 from "../../abi/IERC20.json";
+
+export const getKlimaSupply = async (): Promise<number> => {
+  const provider = getJsonRpcProvider();
+  const klimaContract = new ethers.Contract(
+    addresses.mainnet.klima,
+    IERC20.abi,
+    provider
+  );
+  const totalSupply = await klimaContract.totalSupply();
+
+  return totalSupply;
+};
--- a/lib/utils/index.ts
+++ b/lib/utils/index.ts
@@ -1,4 +1,5 @@
 export { getInteger } from "./getInteger";
+export { getKlimaSupply } from "./getKlimaSupply";
 export { getKlimaUsdcPrice } from "./getKlimaUsdcPrice";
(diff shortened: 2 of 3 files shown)


GreenGauss:
nice will take a closer look soon, but lets put all /api/ endpoints under our main domain not the dapp

in other words if you put it in `site` instead of `dapp` repo then the endpoint will be:
```
klimadao.finance/api/klima-supply/
```

Atmosfearful (changes requested):
Nice! Let's move this to `site` instead of `dapp` so all api endpoints are at our root domain `klimadao.finance/api/klima-supply`

app/pages/api/klima-supply/index.ts
```diff
@@ -0,0 +1,19 @@
+import { NextApiRequest, NextApiResponse } from "next/types";
+import { getKlimaSupply } from "@klimadao/lib/utils";
+
+type ResponseData = {
+    currentKlimaSupply?: string;
+    error?: string;
+};
+
+export default async function handler(
```
Atmosfearful, line 9: We can improve the typings a bit:
```suggestion
interface Response {
  currentKlimaSupply: string;
};

const handler: NextApiHandler<Response> = async (req, res) => {
```

Atmosfearful (changes requested):

app/pages/api/klima-supply/index.ts
0xAurelius: I realized this API actually needs to be simplified to meet CMC's requirement of just returning the supply and nothing else

0xAurelius:
OK moved it to `site` and simplified things to only return the supply itself and no JSON wrapping

0xAurelius:
Finally got the build working, API endpoint seems to work in preview deployment: https://klimadao-site-git-fork-0xaurelius-supply-api-klimadao.vercel.app/api/klima-supply

Atmosfearful (approved):

PR #208, opened 2022-02-25
DRAFT: 18n: First pass on Navbar

biwano:
## Description

Added translation tags for navbar.

## Related Ticket

NA

## Changes

NA

## Checklist


biwano:
@Atmosfearful 
Some tags have ids some not. What should we do?
IMHO having ids is superior.

Issue #157, opened 2022-02-10
Release French & German translations.


Atmosfearful:
* [x] Trans tags on everything except blog posts, and /buy (/buy will likely be moved / refactored soon)
  - from now on, new tags should get IDs as suggested by @biwano
* [x] extract strings
* [x] translators add FR and DE strings to .po files
* [x] make sure `toLocaleString()` and `Intl.format()` and similar methods reflect the chosen locale.
  - pass locale to `Intl.RelativeTimeFormat`
  - on site, the locale is reflected in the url  `klimadao.finance/fr/blog`
  - on app the locale is at runtime
* [ ] show the language picker on production URL (currently hidden). 
* [x] Make sure Pseudo is only visible on dev/staging/preview urls. https://github.com/KlimaDAO/klimadao/pull/320
* [x] Make sure it looks nice
* [x] Make sure it is tab-navigable.
* [x] #293 
* [x] pass locale to prettifySeconds https://github.com/KlimaDAO/klimadao/blob/staging/lib/utils/prettifySeconds/index.ts#L7
* [x] Update locale on HTML root for `site`'s server-side page generation - might have to read the docs for the correct way to do this using the built-in nextjs localization tools https://github.com/KlimaDAO/klimadao/blob/staging/site/pages/_document.tsx#L10 https://github.com/KlimaDAO/klimadao/pull/320
* [ ] Enable Chinese and Russian on Staging only => https://github.com/KlimaDAO/klimadao/issues/329

biwano:
* [ ] App/NavMenu
* [ ] App/Home & Loading
* [ ] App/Stake 
* [ ] App/ChooseBond
* [ ] App/Bond
* [ ] App/PKlima
* [ ] App/Info
* [ ] Site/Footer & Header & Navigation & PageHead
* [ ] Site/Home
* [ ] Site/Blog
* [ ] Site/Disclaimer
* [ ] Site/Resources

PR #211, opened 2022-02-25
i18n: a2. home, loading, checkurlbanner, basic notifications

biwano:
## Description

Added translation tags for home, loading, checkurlbanner, basic notifications components.

## Related Ticket

## Checklist

### DIFF
--- a/app/components/CheckURLBanner/index.tsx
+++ b/app/components/CheckURLBanner/index.tsx
@@ -2,6 +2,7 @@ import React, { FC } from "react";
 
 import styles from "./index.module.css";
 import { Text } from "@klimadao/lib/components";
+import { Trans } from "@lingui/macro";
 
 interface Props {
   onHide: () => void;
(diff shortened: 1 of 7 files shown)


biwano:
The loading component never shows on my computer, is it still needed?

Atmosfearful (approved):
@biwano yup we need `<Loading />` as a hacky fix to make react-router's client-side `HashRouter` compatible with server-side next.js. Next really isn't a good choice for SPAs like this but I am stubborn and want both apps to have as much overlap as possible :)

app/components/CheckURLBanner/index.tsx
```diff
@@ -22,18 +23,23 @@ export const CheckURLBanner: FC<Props> = ({ onHide }) => {
...
-            ⚠️ Verify the URL and bookmark this page!
+            <Trans id="checkurlbanner.verify_url_and_bookmark">
+              ⚠️ Verify the URL and bookmark this page!
+            </Trans>
           </Text>
           <Text t="caption" color="lighter" align="center">
-            <strong>dapp.klimadao.finance</strong> is the only official domain.
+            <strong>dapp.klimadao.finance</strong>{" "}
+            <Trans id="checkurlbanner.is_the_only_official_domain">
+              is the only official domain.
```
Atmosfearful, line 33: I don't think all translations will follow the format of `<noun> is`
so we should wrap the whole thing in a tag right?

Atmosfearful (changes requested):

app/components/CheckURLBanner/index.tsx
biwano: You are right.
I also put a comment to help translators

Atmosfearful (approved):

PR #220, opened 2022-02-28
Add Nomspace integration

Zhethan:
## Related Ticket

Resolves #<issue-number>

## Checklist


Atmosfearful (commented):
pinging @bigint on this--

- I'd prefer to avoid using a third party lib for this, it seems ethers.js supports ENS https://docs.ethers.io/v5/api/providers/provider/#Provider--ens-methods
- Why `nom` lol is that a common thing in the ENS world?

Zhethan:
@Atmosfearful, We're using ensdomains/ens because it's officially maintained by the ENS team and handles things like namehashing and reverse resolutions under the hood. Doing this without a third party-lib would introduce a significant amount of new code.

".nom" is the TLD for the name service we run (compared to .eth for ENS)

bigint:
Why not .eth? What is .nom?

IMO .eth is very famous 😎

Atmosfearful:
@Zhethan Ah I see now. Going to pass on this for the time being, for a few reasons:
- Nobody in the community or contributor discords has mentioned or asked for this.
- As written in our [styleguide](https://github.com/KlimaDAO/klimadao/wiki#avoid-dependencies), we are averse to extra dependencies & increasing bundle size
- The twitter linked from your homepage has been [banned](https://twitter.com/Nomspace)
- I don't want to rely on or send our user's addresses to another third-party RPC if I don't have to.

Happy to revisit when some Klimates want it, and when you guys have spoken with our partnerships team. Thanks and best of luck with nom.space!

PR #224, opened 2022-03-01
Rebased dynamic block rate

### LINKED ISSUE #31: Rebase & Bond times are inaccurate

0xAurelius:
## Description

I took the `dynamic-block-rate` branch, rebased in the latest changes, moved the API itself to `site` instead of `app` and added caching

## Related Ticket

Resolves #31 

## Checklist

### DIFF
--- a/app/actions/app.ts
+++ b/app/actions/app.ts
@@ -3,8 +3,12 @@ import { ethers, providers } from "ethers";
 import { Thunk } from "state";
 import { setAppState } from "state/app";
 
-import { getTreasuryBalance } from "@klimadao/lib/utils";
-import { addresses, ESTIMATED_DAILY_REBASES } from "@klimadao/lib/constants";
+import {
+  getEstimatedDailyRebases,
+  getTreasuryBalance,
+  fetchBlockRate,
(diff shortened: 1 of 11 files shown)


Atmosfearful (changes requested):
2 things:
- [x] ~~move the `API_BASE_URL` and the `getBlockRate` fn to `@klimadao/lib` as they are used in both~~ shit nvm they depend on a vercel var
- [ ] inside getBlockRate do a try/catch and fallback to a default value. Whatever the API is giving us right now is fine :) This way our Fleek deployment will still work and show APY even if our vercel goes down.

Atmosfearful (approved):

PR #225, opened 2022-03-01
Add vienio's dynamic block rate API

0xAurelius:
## Description

Just adds the API without actually integrating into the app so we will have a staging deployment to hit for testing

## Checklist

### DIFF
--- a/site/lib/getBlockRate.ts
+++ b/site/lib/getBlockRate.ts
@@ -1,0 +1,35 @@
+import { POLYGONSCAN_API_KEY, MONTH_IN_SECONDS } from "lib/constants";
+import { getJsonRpcProvider } from "@klimadao/lib/utils";
+
+/** Calculates block rate by dividing one month in seconds by number of blocks that got created during that month */
+export async function getBlockRate(): Promise<number> {
+  const fallbackBlockRate = 2.5;
+
+  // Unix timestamp in seconds
+  const currentTimestamp = Math.floor(Date.now() / 1000);
+  const timestamp30DaysAgo = currentTimestamp - MONTH_IN_SECONDS;
+  try {
+    const provider = getJsonRpcProvider();
+    const latestBlock = await provider.getBlock("latest");
+
+    let url = `https://api.polygonscan.com/api?module=block&action=getblocknobytime&timestamp=${timestamp30DaysAgo}&closest=before`;
+
+    if (POLYGONSCAN_API_KEY !== undefined) {
+      url = url + `&apikey=${POLYGONSCAN_API_KEY}`;
+    }
+
+    const res = await fetch(url);
+    const block30DaysAgo = await res.json();
+
+    const blockRateSeconds =
+      MONTH_IN_SECONDS / (latestBlock.number - block30DaysAgo.result);
+
+    if (!Number.isNaN(blockRateSeconds)) {
+      return blockRateSeconds;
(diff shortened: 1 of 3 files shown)


site/lib/getBlockRate.ts
```diff
@@ -0,0 +1,36 @@
+import { POLYGONSCAN_API_KEY, MONTH_IN_SECONDS } from "lib/constants";
+import { getJsonRpcProvider } from "@klimadao/lib/utils";
+
+// Calculates block rate by dividing one month in seconds by number of blocks that got created during that month
+
+export async function getBlockRate() {
```
Atmosfearful, line 6:
```suggestion
export async function getBlockRate(): Promise<number> {
```

Atmosfearful, line 6: protip for vscode, if you use JSDoc style comments, then they appear in the hover-preview when you hover the variable elsewhere in the codebase

```suggestion

/** Calculates block rate by dividing one month in seconds by number of blocks that got created during that month */
export async function getBlockRate() {
```

site/lib/getBlockRate.ts
```diff
@@ -0,0 +1,36 @@
...
+  const currentTimestamp = Math.floor(Date.now() / 1000);
+  const timestamp30DaysAgo = currentTimestamp - MONTH_IN_SECONDS;
+  try {
+    const provider = getJsonRpcProvider();
+    const latestBlock = await provider.getBlock("latest");
+
+    let url = `https://api.polygonscan.com/api?module=block&action=getblocknobytime&timestamp=${timestamp30DaysAgo}`;
+
+    if (POLYGONSCAN_API_KEY !== undefined) {
+      url = url + `&closest=before&apikey=${POLYGONSCAN_API_KEY}`;
```
Atmosfearful, line 19: wouldn't we want &closest=before for unathenticated requests as well?

site/pages/api/block-rate/index.ts
```diff
@@ -0,0 +1,28 @@
...
+    // >24 hours -> stale, recalculate & return fresh value
+    res.setHeader(
+      "Cache-Control",
+      "public, s-maxage=43200, stale-while-revalidate=43200"
+    );
+
+    const blockRate = await getBlockRate();
+    res.status(200).json({ blockRate30Day: blockRate.toFixed(4) });
+  } catch (err) {
+    res.status(404).json({ error: "failed to fetch data" });
```
Atmosfearful, line 26: I believe the convention is:
status 500 - internal server error
json({ message })

if typescript doesnt like that, you can try typing the res on line 11 like this
```ts
interface ResponseData {
  blockRate30Day: string;
};
interface ErrorObj {
}


res: NextApiResponse<ResponseData | ErrorObject>
```

Atmosfearful: this is also more handy because then we can export and use the `ResponseData` type in our apps when we fetch from this endpoint, and it wont complain about being an optional property

Atmosfearful (changes requested):
just nitpicks, i want to start enforcing my style guide better :)

site/pages/api/block-rate/index.ts
0xAurelius: awesome TypeScript is so cool!

Atmosfearful (approved):

PR #228, opened 2022-03-02
Add ability to upload inline PDFs in blog

rstycedar:
feature request that enables authors to upload inline PDFs in Sanity IO

### DIFF
--- a/cms/schemas/blockContent.js
+++ b/cms/schemas/blockContent.js
@@ -55,19 +55,20 @@ export default {
             },
           },
           {
-            title: "Internal link",
+            title: "Link to file",
             name: "internalLink",
             type: "object",
             fields: [
               {
-                name: "item",
+                title: "Upload a file and link to it below",
+                name: "uploadedFile",
                 type: "reference",
                 to: [{ type: "fileUpload" }],
                 validation: (Rule) => Rule.required(),
               },
             ],
             blockEditor: {
-              icon: () => <div>🗂</div>,
+              icon: () => <div>📂</div>,
             },
           },
         ],
--- a/cms/schemas/blockContent.js
+++ b/cms/schemas/blockContent.js
@@ -59,5 +59,11 @@ export default {
       type: "image",
       options: { hotspot: true },
     },
+    {
+      name: "pdf",
+      title: "PDF",
+      type: "reference",
+      to: { type: "pdf" },
+    },
   ],
 };
--- a/cms/schemas/blockContent.js
+++ b/cms/schemas/blockContent.js
@@ -1,3 +1,5 @@
+import React from "react";
+
 /**
  * This is the schema definition for the rich text fields used for
  * for this blog studio. When you import it in schemas.js it can be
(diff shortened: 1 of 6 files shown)


site/components/BlockContentRenderer/index.tsx
```diff
@@ -101,6 +101,13 @@ const serializers: BlockContentProps["serializers"] = {
         </div>
       );
     },
+    pdf: (params: { node: { name: string; url: string } }) => {
+      return (
+        <Text t="body2" className={styles.paragraph}>
+          Download PDF <A href={`${params.node.url}?dl=`}>{params.node.name}</A>
```
0xAeterno, line 107: What do you think about allowing content writers to control the "Download PDF" CTA copy?
Could we see a use case where we might not want have this hardcoded?

0xAeterno (commented):
This is really cool @rstycedar, nice work 🔥🔥🔥
I learnt a lot going through this PR of yours!

Couple things I thought about while reviewing this:
- How might we apply the pdf download link to an image or button? I think this was flagged as potential use cases for this functionality too.

Atmosfearful (changes requested):
Indeed me and @rstycedar spoke already but I'm posting here for transparency:
ideally this can be a generic "file" of known file types, and can just be inline text rather than a text block.
I believe sanity refers to this as 'annotation' https://www.sanity.io/docs/configuration#057f677d86d9

rstycedar:
> Indeed me and @rstycedar spoke already but I'm posting here for transparency: ideally this can be a generic "file" of known file types, and can just be inline text rather than a text block. I believe sanity refers to this as 'annotation' https://www.sanity.io/docs/configuration#057f677d86d9

@Atmosfearful @0xAeterno we can also add a displayAsButton option to the schema if we wanted to let authors choose to display the annotations as a button

cms/schemas/blockContent.js
```diff
@@ -48,6 +50,25 @@ export default {
...
+            fields: [
+              {
+                name: "item",
+                type: "reference",
+                to: [{ type: "fileUpload" }],
+                validation: (Rule) => Rule.required(),
+              },
+            ],
+            blockEditor: {
+              icon: () => <div>🗂</div>,
```
Atmosfearful, line 70: looks like a weird black blob on my windows  screen, can we try this emoji instead:
```suggestion
              icon: () => <div>📂</div>,
```

Atmosfearful (changes requested):

cms/schemas/blockContent.js
Atmosfearful: image

site/components/BlockContentRenderer/index.tsx
```diff
@@ -82,6 +82,14 @@ const serializers: BlockContentProps["serializers"] = {
         </A>
       );
     },
+    internalLink: ({ children, mark }) => {
+      const { href } = mark;
+      return (
+        <A className={styles.link} href={`${href}?dl=`}>
```
Atmosfearful, line 88: lets stick with browser default i dont think most people want to save it

```suggestion
        <A className={styles.link} href={`${href}`}>
```

Atmosfearful (changes requested):

site/components/BlockContentRenderer/index.tsx
rstycedar: good call that would come off as fishy

Atmosfearful (approved):

PR #245, opened 2022-03-08
Preload english translations in app

biwano:
## Description

Fixes the translations message id flashing by preloading the English language file into the HTML file produced by nextjs

## Related Ticket

## Checklist

### DIFF
--- a/app/pages/index.tsx
+++ b/app/pages/index.tsx
@@ -4,7 +4,16 @@ import { WithRedux } from "components/WithRedux";
 import { NextPage } from "next";
 import { PageHead } from "components/PageHead";
 import { IS_PRODUCTION } from "lib/constants";
+import { messages as default_messages } from "../locale/en/messages";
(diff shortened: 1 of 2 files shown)


app/pages/index.tsx
```diff
@@ -4,7 +4,16 @@ import { WithRedux } from "components/WithRedux";
 import { NextPage } from "next";
 import { PageHead } from "components/PageHead";
 import { IS_PRODUCTION } from "lib/constants";
+import { messages as default_messages } from "../locale/en/messages";
+import { i18n } from "@lingui/core";
 
+export async function getStaticProps() {
+  i18n.load("en", default_messages);
+  i18n.activate("en");
```
Atmosfearful, line 12: So this means non-english pages will see a flash of english text. Better than the string IDs I guess! We don't have any other option since we are treating this like a static Single Page App.

Atmosfearful (approved):

PR #253, opened 2022-03-14
Includes hints into po files during extraction phase

biwano:
## Description

The problem: When messages are extracted by lingui, the original text is not in the po files. Translating can be difficult if the text in not easily visible on the translator browser.
This could be solved by adding comments with the translation tags but this would give developpers more work.
This PR tries to solve this by adding automatically the original text as a comment into the po files during extraction phase.

## Related Ticket

## Checklist


Atmosfearful:
As discussed on our call, we can close this in favor of translation.io -- we can reopen if translation.io turns out to not work for us.

PR #260, opened 2022-03-19
Integration with translation.io

biwano:
## Description

Integrates KlimaDAO with translation.io

Notes: 
- The README is updated with notes for translators
- `npm run sync` and `npm run sync_and_purge` use a local config file named `.linguirc_local`. this is to not commit the translation.io api key. I could make a script that builds this file using the .linguirc file and then .env file.
- From my tests: Running `npm run sync` does in that order:
  - Pull translations (content of msgstr) from translation.io
  - Extract translations from source files
  - Send new tags (content of msgid) to translation.io 
  - Compile translations into js files
-  Running `npm run sync_and_purge` does the same, plus it deletes obsolete translation tags
- Considerations: 
  - The synching should be done on only one branch or there will be constant changes. 
  - If the master branch is chosen, translations will come at a later stage leaving some part of the application untranslated.
  - If the develop branches are chosen syncing with purge may delete translation no longer used in develop but still existing on the master branch. I think this is low risk though.

I recommend using `npm run sync` (sync without purge) on the staging or development branch in continuous integration scripts. And doing `npm run sync_and_purge` (sync and purge) manually once in a while when development activity is low.

- Caveat: The texts are displayed in alphabetical order and not in the po file order. Consequently, it is difficult for the translator to stay in context.

## Related Ticket

## Checklist


site/package.json
```diff
@@ -8,6 +8,8 @@
     "lint": "next lint",
     "lingui:extract": "NODE_ENV=test lingui extract --clean",
     "lingui:compile": "NODE_ENV=test lingui compile",
+    "sync": "NODE_ENV=test LINGUI_CONFIG=.linguirc_local lingui extract --overwrite && lingui compile",
```
Atmosfearful, line 11: lingui.config.ts should be supported out of the box according to the lingui docs. Which would let us use `process.env`

At some point we only need to keep the api key in Github so the Github action can run the sync for us right?
Do you think devs would want to be running this locally @biwano ?

I believe github should let us add secrets to process.env is this correct @ladytrekker ?

README.md
```diff
@@ -44,3 +44,22 @@ A set of NPM Workspace commands are provided, and can be run from the root folde
...
+## Translatiing
+
+### For translators
+
+#### Setup
+
+- Create an account on [translation.io](https://translation.io)
+- Request access to the project atmosfearful/klimadao in the #content-translation KlimaDaO discord server
+
+#### Translating
```
Atmosfearful, line 57: So when we run `sync` does it create changes in Git? For each locale except english right?

Or would you rather we only commit English and then just pull everything from the translation.io cloud at build time?

Tradeoffs for both approaches...

site/package.json
biwano: Yeah, we should do that. But given the order in which lingui looks at the configuration file, the file commited should be lingui.config.js and the override .linguirc, I will make the change.

Atmosfearful:
> The synching should be done on only one branch or there will be constant changes.

On this point, do you forsee any problem with syncing on staging but having a bot automate it? I.e. after every PR run a github action that extracts strings and commit them?

site/package.json
ladytrekker: Yes, that is the idea with github actions. 
It may make sense then to upload new translations of the master translation (pseudo-en ?)to translation.io on every merge to staging atomatically. Let's discuss what would be the best workflow with the new integration.

biwano: > lingui.config.ts should be supported out of the box according to the lingui docs. Which would let us use `process.env`
Definitely, I changed that.
> 
> At some point we only need to keep the api key in Github so the Github action can run the sync for us right? Do you think devs would want to be running this locally @biwano ?
They should not have to.
> 
> I believe github should let us add secrets to process.env is this correct @ladytrekker ?

I made some changes that makes lingui sync if the KLIMA_TRANSLATION_IO_KEY variable is setup.
I also factorized the linguijs config between app and site. I can make it configurable if need be later.

README.md
biwano: So sync:
- pushes only message ids to translation.io
- pulls all translations in local po files
- 
It has nothing to do with git.
We should add a command after sync that does something like
```git add site/locale
git add app/locale
git commit -m "fetched changes from translation.io`
git push```

biwano:
> > The synching should be done on only one branch or there will be constant changes.
> 
> On this point, do you forsee any problem with syncing on staging but having a bot automate it? I.e. after every PR run a github action that extracts strings and commit them?

No there should be none because it will only add *new* keys in translation.io.

Issue #206, opened 2022-02-25
Use GitHub actions for frictionless automated releases


Atmosfearful:
# Goal
- A robot merges `staging` to `main` automatically on Mondays , Tuesdays, Wednesdays and Thursdays

Stretch Goals
- Merge & rebase the translators branch. We can do this manually to start to test out the workflow.
- Publish some form of changelog to the public discord server so the community can pat us on the back with 🎉 emojis @0xAurelius 

### But what about QA and translations?
- A PR should never be merged to staging unless we are confident it's production-ready.
- PRs with big or critical changes should get extra testing from QA team before being merged to staging.
- It's OK to release untranslated content. With a frequent release schedule, translators can follow-up quickly.
  - We should never have devs get blocked by translation

# Requirements

- [ ] Extract Strings and commit to staging
- [ ] Fast-forward merge of staging to main `git merge --ff staging`. Vercel does the rest.
  - needs to be a local `--ff` merge so the commit hashes match. Otherwise we end up having to constantly rebase all the downstream branches.
- [ ] Runs on a [schedule](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule), I prefer early morning PST

PR #268, opened 2022-03-25
Schedule release staging to main with github actions

ladytrekker:
## Description

This PR does the following:

On **every Monday and Thursday at 15h** (=> 3pm UTC => 8am PCT) trigger a **merge** from `staging` into `main`.

The **merge** is only possible if:
- translations are compiled and pushed with a commit to `staging`
- the last commit on `staging` was successfully deployed (to Vercel)
- check if `staging` can be merged with `main` with `fast-forward` only

To accomplish this whole workflow several files are added:
- workflow file **update_translations** 
- workflow file **wait_for_status_check**
- TypeScript action **check-status**
- workflow file **merge_staging**
- bash script **merge_staging** for merge staging to main with `--ff-only`
- workflow file **schedule_auto_merge**

--- 

### Workflow "update_translations"
Steps:
- checkout `staging` with a personal access token of a user which can push to `staging`
- install node and npm with the correct versions
- install all npm packages
- run `npm run compile strings`
- check if new translation have been created
- if yes => push these new locale files to `staging`

--- 

### Workflow "wait_for_status_check"
Steps:
- checkout the repo with the default github token (always available on action "checkout")
- build the local Typescript action **check-status**
- run this action to get the status checks of the last commit on `staging`
- return the status result
 
--- 

### TypeScript action "check-status"
A TypeScript action in this repository which:
- takes a reference as an argument (can be sha, tag or branch name, here: `staging`)
- create a github octokit with the default GITHUB_TOKEN
- GET the status of the last commit on branch `staging`
- if status is still `pending` => retry after 10 seconds until timeout of 240 (values can be changed)
- return the status of either `failure` or `success`

--- 

### Workflow "merge_staging"
Steps:
- checkout `staging` with a personal access token of a user which can push to `main`
- run the bash script `merge_staging.sh` which does the final merge

--- 

### bash script **merge_staging**
Steps:
- check if `main` and `staging` can be merged with fast forward only
- exit if fast forward is not possible
- otherwise do the merge and push to `main`

--- 

### Workflow "schedule_auto_merge"
The main workflow file which calls each workflow mentioned above in a specific order.
In this file the branch names are defined and passed as inputs to each workflow.

---

## TESTING

See here the final successful run https://github.com/KlimaDAO/klimadao/actions/runs/2040194399
**Note:** you can open each job and see the output of each step

See here the result of the fast forward merge of both test branches including the commit with the extracted translations:
- https://github.com/KlimaDAO/klimadao/commits/tmp-staging-test-branch
- https://github.com/KlimaDAO/klimadao/commits/tmp-main-test-branch

**=> all commit hashes are the same!**

---

### Why is this PR a "draft" ? 🔥  

Before merging this PR the following steps have to be done 🔥 :
- edit the protection rules for `staging` and `main` according to the two test branches used for this PR

--- 

## Related Ticket

Resolves https://github.com/KlimaDAO/klimadao/issues/206


.github/actions/check-status/__tests__/poll.test.ts
```diff
@@ -0,0 +1,138 @@
...
+    log: () => {},
+    statusURL: "repos/owner/repo-name/commits/ref/status",
+    owner: "owner",
+    repo: "repo-name",
+    ref: "ref",
+    timeoutSeconds: 3,
+    intervalSeconds: 1,
+  });
+
+describe("poll", () => {
```
Atmosfearful, line 19: neat, when do these actually run? I didn't see them in the vercel logs. Once our tests get more extensive i'd be down to create a separate testrunner (separate from builds, to keep those fast)

Atmosfearful (approved):
looks good to me, I'll let @0xAurelius give some feedback if he has any.
So we just need to add the ENV vars as I described in discord, and then the sync should work automatically.

0xAurelius (approved):
Overall looks good to me!

PR #278, opened 2022-03-29
refactor: offset ui redesign

0xAeterno:
## Description

Implemented design changes suggested by our designer, key callouts:
- Updated offset form ux
- Centering token logos
- Update copy and grouped 'retirement beneficiary' form fields

## Changes

| Before  | After  |
|---------|--------|
|image|image|
-->

## Checklist


app/components/views/Offset/index.tsx
```diff
@@ -464,54 +464,50 @@ export const Offset = (props: Props) => {
...
-              placeholder={t({
-                id: "offset.which_address_retiring",
-                message: "Which address are you retiring on behalf of?",
-              })}
-            />
+            </div>
           </div>
+
           <div className={styles.input}>
             <label>
```
Atmosfearful, line 502: I know this isn't your doing but since we in here:
Do you think screenreaders or other a11y tools benefit from adding an `id` here along with a `for` on the TextArea? I'm getting rusty on a11y already

app/components/views/Offset/index.tsx
```diff
@@ -520,13 +516,14 @@ export const Offset = (props: Props) => {
...
+
           <div className="disclaimer">
             <GppMaybeOutlined />
             <Text t="caption">
               <Trans id="offset_disclaimer">
-                Be careful not to expose any sensitive personal information. The
-                information you provide will be made a permanent, immutable
-                record on a public blockchain.
+                Be careful not to expose any sensitive personal information.
+                Your message can not be edited and will permanently exist on the
```
Atmosfearful, line 525: IMO the whole 'it's public' piece is key to mention:

```suggestion
                Your message can not be edited and will permanently exist on a public
```

app/components/views/Offset/index.tsx
```diff
@@ -464,54 +464,50 @@ export const Offset = (props: Props) => {
...
+                <Trans id="offset.retirement_message">Retirement message</Trans>
               </Text>
             </label>
             <textarea
               value={retirementMessage}
               onChange={(e) => {
-                if (e.target.value.length >= 280) {
-                  return;
-                }
+                if (e.target.value.length >= 280) return;
```
Atmosfearful, line 510: This would be a change request and +1 day delay for re-review on my last team: "always use brackets for the return block"

....I'm just commenting to celebrate that I dgaf and it doesn't matter either way 🎉 🎉🎉 😛

app/components/views/Offset/index.tsx
```diff
@@ -464,54 +464,50 @@ export const Offset = (props: Props) => {
...
-            />
-          </div>
-          <div className={styles.input}>
-            <label>
-              <Text t="caption" color="lightest">
-                <Trans id="offset.beneficiary_address">
-                  BENEFICIARY ADDRESS (optional: defaults to connected address)
+          <div className={styles.beneficiary}>
+            <Text t="caption" color="lighter">
+              <Trans id="offset.retirement_message">
```
Atmosfearful, line 469: Duplicate id
Indeed I open the preview and I can see you need to run `extract-strings` as well, and that this is ignoring the wrapped text and using the pre-defined value in the `.po` file

It's annoying that this problem only surfaces once you are in preview mode, but everything seems to work fine in local dev mode. Probably because the .PO file is missing an id so it uses the fallback?
I'm going to open an issue and see if biwano can find a way to smooth this out so we either:
A. catch these mistakes earlier  (warn on duplicates?)
B. don't introduce them at all (no ids?)

app/components/views/Offset/index.tsx
```diff
@@ -374,36 +374,13 @@ export const Offset = (props: Props) => {
...
-            }}
-            onItemSelect={handleSelectRetirementToken}
-          />
           <div className={styles.input}>
             <label>
-              <Text t="caption" color="lightest">
+              <Text t="caption" color="lighter">
                 <Trans id="offset.amount_in_tonnes">
-                  AMOUNT IN CARBON TONNES
+                  How many carbon tonnes would you like to retire?
```
Atmosfearful, line 383:
```suggestion
                  How many tonnes of carbon would you like to offset?
```

Wording tweak to illustrate the relationship between offset -> retirement better
1. enter quantity to "offset"
2. select which token to "retire"

Atmosfearful (changes requested):

app/components/views/Offset/index.tsx
0xAeterno: Absolutely, we have placeholder text that will help with a11y in lieu of having explicit labels.
Thinking of addressing this in the next piece of pledge dashboard work, where I'll create some reusable form input components. We can retrospectively use those form components in that this form.

What do you think? :)

https://www.a11ywithlindsey.com/blog/introduction-accessible-labeling

app/components/views/Offset/index.tsx
0xAeterno: Nice catch 👀
Yeah I didn't catch that in my local dev environment.

Let's document that somewhere as part of the dev workflow. I'll add a it to the PR template as a prompt.

app/components/views/Offset/index.tsx
0xAeterno: Done! :)

Atmosfearful (approved):

PR #285, opened 2022-03-30
Makes the `lingui:extract` command return an error if a file cannot be processed

biwano:
## Description

Makes the `lingui:extract` command return an error if a file cannot be processed

## Related Ticket

## Checklist


Atmosfearful (approved):
Awesome! Simple way to catch common mistakes. Cheers.

Issue #250, opened 2022-03-10
[CMS] Create a reuseable Disclaimer for blog posts


Atmosfearful:
Right now authors are copy-pasting the same disclaimer into every blog post.

We should render this in smaller text at the bottom, automatically.

Atmosfearful:
A good one for you @rstycedar but not very high priority

PR #288, opened 2022-03-30
add disclaimer to sanity schema

rstycedar:
#250 
* add disclaimer boolean to Sanity Post schema that will default to true for all new blog posts
*  this will make it so the authors wont have to keep copy and pasting it into the post

Screen Shot 2022-03-31 at 10 11 04

### DIFF
--- a/site/components/pages/Blog/Post/index.tsx
+++ b/site/components/pages/Blog/Post/index.tsx
@@ -88,10 +88,10 @@ export const PostPage = (props: PostProps) => {
             <BlockContentRenderer blocks={props.post.body} />
             {props.post.showDisclaimer && (
               <div className={styles.disclaimer}>
-                <Text t="caption">
+                <Text t="caption" color="lighter">
                   <Trans id="blog.disclaimer.title">Disclaimer:</Trans>
                 </Text>
-                <Text t="caption">
+                <Text t="caption" color="lighter">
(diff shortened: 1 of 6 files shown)


site/components/pages/Blog/Post/index.tsx
```diff
@@ -78,6 +85,56 @@ export const PostPage = (props: PostProps) => {
...
+                <Text t="caption">
+                  <Trans id="blog.disclaimer.title">Disclaimer:</Trans>
+                </Text>
+                <Text t="caption">
+                  <Trans
+                    id="blog.disclaimer.description"
+                    comment="Long sentence"
+                  >
+                    The information provided in this blog post pertaining to
+                    Klima DAO (“Klima DAO”), its crypto-assets, business assets,
```
0xAeterno, line 99: Just need to update copy to "KlimaDAO" in a few places :)

0xAeterno (changes requested):
Nice one @rstycedar! 🔥

Will this change maintain the italic script on the disclaimer?

site/components/pages/Blog/Post/index.tsx
Atmosfearful: good catch!

Atmosfearful (approved):

Issue #255, opened 2022-03-15
[EPIC] Klima Infinity Pledge dashboard


Atmosfearful:
#### Feature branch for epic
https://github.com/KlimaDAO/klimadao/tree/klima-infinity-pledge-dashboard

A shareable, search-indexable dashboard.
- shows your carbon assets and retirements
- allows you to connect your wallet, edit the pledge name/description/methodology and footprint.
- charts show your total balances over time

Designs [in figma.](https://www.figma.com/file/krpWuauhMXxgNBcvkPqP3q/Klima-Infinity-DASHBOARD?node-id=125%3A1500)

TODO:
- [x] Get subgraph endpoint and/or relevant contract addresses from and integrate data using `getStaticProps` @0xAeterno 
  -  We don't need subgraphs for v0.5, just when we add the charts/graphs
- [x] Create basic layout without any graph or charting libraries (just pledge data read/write, and contract/graph data)
- [x] Mobile + desktop views

0xAeterno:
> * [x]  The 0x address of the pledge should always be visible somewhere. It acts as the fallback when there is no name. But when there is a name, we need to show it somewhere (DESIGN NEEDED).

A concatted address is always visible on a pledge regardless of whether or not there is an pledge tied to address :)

PR #311, opened 2022-04-11
[pledge-dashboard] feat:  modal form, query retirements and carbon assets

0xAeterno:
#### Related Epic https://github.com/KlimaDAO/klimadao/issues/255

-  reusable `Input`, `Textarea` and `Modal` components
    - colocated under `/site/components` for the time being
- add pledge search with eth address validation on `/pledge`
    - users will redirect to `/pledge` if `/pledge/:address` resolves to an invalid address
- add `PledgeForm`, to be wired up in next PR
- fetch retirements and balances by pledge page address
- refactored and restructured `/pledge` folders a tad
    - types and queries are a bit all over the place, will refactor when I get a better sense of how to best structure it within the app
    - open to feedback ofc :)
## Changes

| Before  | After  |
|---------|--------|
|img here |img here|

## Checklist


lib/components/Buttons/styles.ts
```diff
@@ -16,13 +16,10 @@ export const button_primary = css`
...
-  &:hover:not(:disabled),
-  &:focus:not(:disabled) {
+  &:active:not(:disabled) {
+    transform: scale(0.95);
     opacity: 0.6;
   }
-  &:focus:not(:disabled) {
-    transform: scale(0.9);
-  }
   &,
```
0xAeterno, line 23: This was causing buttons to stay scaled down on user click.

site/components/Modal/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export const Modal: FC<Props> = (props) => {
+  useEffect(() => {
+    if (props.showModal) {
+      document.body.style.overflow = "hidden";
+    } else {
+      document.body.style.overflow = "";
+    }
+  }, [props.showModal]);
```
0xAeterno, line 31: Disables body scroll when modal is open.

site/components/pages/Pledge/PledgeDashboard/PledgeForm/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+import * as styles from "./styles";
+
+// const InputField: FC = (props) => (
+//   <>
+//     <label>{props.label}</label>
+//     <props.Component id={props.name} name={props.name} />
+//     {props.errorMessage && errorMessage}
+//   </>
+// );
```
0xAeterno, line 15: Will revisit on integration with moralis.

site/components/Input/index.tsx
```diff
@@ -0,0 +1,64 @@
+import React, { InputHTMLAttributes } from "react";
+import styled from "@emotion/styled";
+import * as typography from "@klimadao/lib/theme/typography";
+import breakpoints from "@klimadao/lib/theme/breakpoints";
+
+const StyledInput = styled.input<InputProps>`
+  ${typography.caption}; // TODO: not working, syntax not compatible with styled?
```
0xAeterno, line 7: Any ideas on why this css interpolation doesn't work? :/

Atmosfearful: Also. I could be wrong but `styled` miight need special babel transpilation magic, unlike `css` which we can safely use anywhere including `@klimadao/lib`

site/components/Modal/index.tsx
Atmosfearful: Good opportunity for a custom hook 🎉 probably need this for our other modals
```ts
useScrollLock(props.showModal);

// or alternative api for manual control
const setScrollLock = useScrollLock();
if (props.showModal) {
  setScrollLock(true);
}
```

site/components/Modal/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { Text } from "@klimadao/lib/components";
+
+import * as styles from "./styles";
+
+type ModalWrapperProps = {
+  showModal: boolean;
+};
+
+export const ModalWrapper = styled.div`
+  display: ${(props: ModalWrapperProps) =>
```
Atmosfearful, line 13: oh gross, another reason I left styled-components 😛 😛 😛 

keeping logic in the react/js scope, and keeping classnames & styles in the stylesheets, is more readable and predictable IMO
```
// js
className={cx({ hidden: !props.showModal })}

// css
.hidden {
  display: none;
}
```

site/components/pages/Pledge/PledgeDashboard/Cards/RetirementsCard/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+type Props = {
+  pageAddress: string;
+};
+
+export const RetirementsCard: FC<Props> = (props) => {
+  const [retirements, setBalances] = useState<Retirements | null>(null);
+
+  useEffect(() => {
+    (async () => {
+      const retirements = await getRetirements({ address: props.pageAddress });
```
Atmosfearful, line 20: So fetching this client-side on mount means we can't properly server-render the balance, correct?
this should probably move to getStaticProps and be passed down

site/components/pages/Pledge/PledgeDashboard/styles.ts
```diff
@@ -0,0 +1,46 @@
...
+  grid-column: span 2;
+  display: grid;
+  gap: 1.6rem;
+  justify-items: center;
+  padding: 2rem 0;
+
+  .profileImage {
+    display: grid;
+    align-items: center;
+    width: 80px;
```
Atmosfearful, line 21: always `rem`s plz 🙏

site/pages/_app.tsx
```diff
@@ -45,11 +48,16 @@ function MyApp({ Component, pageProps, router }: AppProps) {
   }, [locale]);
 
   return (
-    <I18nProvider i18n={i18n}>
-      <GridContainer>
-        <Component {...pageProps} />
-      </GridContainer>
-    </I18nProvider>
+    <MoralisProvider
+      appId={process.env.NEXT_PUBLIC_MORALIS_APP_ID as string}
```
Atmosfearful, line 52: I need your input:
i've gotten burned in the past by apps with process.env scattered everywhere. It led to bugs when the env vars changed, because it's untyped or type-asserted and TS didn't throw errors.

For this reason I try to export all process env vars from a typed `constants.ts` file. But I get paranoid that this might lead to secrets being exposed in the client? That's more my lack of understanding at how treeshaking/bundling works.

So we can either:
export all process.env vars from `site/lib/constants`
or create a `secrets.ts` and `constants.ts` file and export from both
or keep using process.env

preference?

site/components/pages/Pledge/PledgeDashboard/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { PledgeForm } from "./PledgeForm";
+import { PledgeLayout } from "../PledgeLayout";
+import * as styles from "./styles";
+
+type Props = {
+  pageAddress: string;
+  pledge: Pledge;
+};
+
+const defaultValues = (pledge: Pledge): Pledge =>
```
Atmosfearful, line 28: interesting pattern. If we just have strong types across the api endpoint and fetch code, do we need to do this?

Atmosfearful (changes requested):

site/components/pages/Pledge/PledgeDashboard/Cards/RetirementsCard/index.tsx
0xAeterno: As discussed, contract calls impact the intial page load significantly so we'll keep fetching client-side for now

site/components/Modal/index.tsx
0xAeterno: Funnily enough, all the other modals don't run into the body scroll issue. It's mind boggling!
Let's card this up as a future refactor that anyone can work on :)

lib/components/Buttons/styles.ts
0xAeterno: Nice one, yeah I noticed that too. What do you think utilising `:active` so the button can return to its original size after a user click?

site/components/Input/index.tsx
0xAeterno: No strong opinions here. I've collaborated on design systems using the following pattern above.
I'd love to learn and work out how to similar thing using css composition instead!

site/components/pages/Pledge/PledgeDashboard/styles.ts
0xAeterno: 🦅 👁️

site/components/pages/Pledge/PledgeDashboard/index.tsx
0xAeterno: Potentially. Setting default values for the form/page removes the need to handle `{ pledge: null }`.
I'll defs revisit this when I get a clearer picture on the patterns we'd like to establish

0xAeterno:
Going to merge this into the feature branch for now to avoid crazy rebases conflicts again 😅
I'm keeping tabs on all the comments to polish the code once all the bits are in place!

PR #320, opened 2022-04-19
Add German Translation

ladytrekker:
## Description

This PR adds german to both site and app.

It also includes minor refactorings:
- re-enable translations on Site again for NextJS (FOR STAGING ONLY)
- autodetect locale for Site (FOR STAGING ONLY)
- load PO files if on development in App (Site does the same, better developer experience for local testing)
- ensure Pseudo is only shown when not on production
- Make Menu translatable
- Highlight current active locale in Menu
- Better Styles for Menu in App
- Let NextJS set the HTML language tag by itself

## Related Ticket

Part of https://github.com/KlimaDAO/klimadao/issues/157

## Checklist


Atmosfearful (approved):
sweeeet

Issue #117, opened 2022-01-31
Enable automated dependency updates


ladytrekker:
## Problem

To maintain a Node project one need to maintain a lot of dependencies too.
Github comes with dependabot by default. However, it is still not possible to group dependencies which might be quite annoying when hosting a monorepo.
- [issue](https://github.com/dependabot/dependabot-core/issues/1190)
- [issue](https://github.com/dependabot/dependabot-core/issues/2137)

**Requirements for an alternative tool:**

- Monorepo support
- Ability to group upgrades of the same dependency in multiple sub-projects into one PR

### Solution

- [ ] Look into [renovate](https://github.com/renovatebot/renovate/) which supports [monorepos](https://github.com/apps/renovate)
- [ ] discuss the research
- [ ] integrate

Atmosfearful:
You still think we should do this @ladytrekker ?

PR #323, opened 2022-04-20
Remove cms from NPM workspaces, upgrade CMS dependencies

Atmosfearful:
## Description

 This PR removes `cms` from the npm workspaces and goes back to relying on the sanity CLI, which uses yarn instead of npm. The cli makes it easier to keep sanity up to date, and should improve install speeds by reducing the number of dependencies in our root package-lock.json.
 
 Only devs who work with the cms need to install the cms deps.

After this is updated I will open a PR to upgrade all the core dependencies for `app` and `site`


ladytrekker (approved):
Thanks. Looks good.

Speaking of core dependencies:
We can have dependabot to open PR for each dependency update.
See the ticket here: https://github.com/KlimaDAO/klimadao/issues/117
Want that?

Issue #331, opened 2022-04-21
Add Error Pages for Site and App


ladytrekker:
Both site and app have no fallback on HTTP errors.

Add pages for:

- 404
- 500
- https://nextjs.org/docs/advanced-features/custom-error-page

PR #337, opened 2022-04-25
Site: Add 404 and 500 Error pages

ladytrekker:
## Description

This PR does:
- add simple 404 page to Site => [Test URL](https://klimadao-site-git-add-error-pages-klimadao.vercel.app/asdasdasdasd)
- add simple 500 page to Site
- see https://nextjs.org/docs/advanced-features/custom-error-page
- **Small fix: let Footer icons break on small devices**

## Related Ticket

Resolves https://github.com/KlimaDAO/klimadao/issues/331

## Changes

| Before  | After  |
|---------|--------|
| footer_before | footer_after |

## Checklist

### DIFF
--- a/site/components/pages/errors/styles.ts
+++ b/site/components/pages/errors/styles.ts
@@ -1,13 +1,15 @@
 import { css } from "@emotion/css";
 import breakpoints from "@klimadao/lib/theme/breakpoints";
 
-export const fullheight = css`
-  min-height: calc(
-    100vh - (var(--footer-height) / 2 + var(--header-height) / 2)
-  );
+export const errorPageWrapper = css`
+  display: grid;
+  grid-column: full;
+  grid-template-columns: inherit;
+  min-height: 100vh;
+  grid-template-rows: 1fr auto;
 
   ${breakpoints.large} {
-    min-height: calc(100vh - (var(--footer-height) + var(--header-height)));
+    grid-template-rows: auto 1fr auto;
   }
 `;
 
--- a/site/components/pages/errors/styles.ts
+++ b/site/components/pages/errors/styles.ts
@@ -1,0 +1,20 @@
+import { css } from "@emotion/css";
+import breakpoints from "@klimadao/lib/theme/breakpoints";
+
+export const fullheight = css`
+  min-height: calc(
+    100vh - (var(--footer-height) / 2 + var(--header-height) / 2)
+  );
+
+  ${breakpoints.large} {
+    min-height: calc(100vh - (var(--footer-height) + var(--header-height)));
+  }
+`;
+
+export const textGroup = css`
+  grid-column: main;
+  display: flex;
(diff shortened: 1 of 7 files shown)


site/pages/404.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { Custom404 } from "components/pages/errors/Custom404";
+import { loadTranslation } from "lib/i18n";
+
+export const getStaticProps: GetStaticProps = async (ctx) => {
+  const translation = await loadTranslation(ctx.locale);
+  return {
+    props: {
+      translation,
+    },
+    revalidate: 60,
```
Atmosfearful, line 11: we don't need to revalidate since this page will never change between builds

site/pages/500.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { Custom500 } from "components/pages/errors/Custom500";
+import { loadTranslation } from "lib/i18n";
+
+export const getStaticProps: GetStaticProps = async (ctx) => {
+  const translation = await loadTranslation(ctx.locale);
+  return {
+    props: {
+      translation,
+    },
+    revalidate: 60,
```
Atmosfearful, line 11: same here, no need to revalidate

site/components/pages/errors/styles.ts
```diff
@@ -0,0 +1,20 @@
+import { css } from "@emotion/css";
+import breakpoints from "@klimadao/lib/theme/breakpoints";
+
+export const fullheight = css`
+  min-height: calc(
+    100vh - (var(--footer-height) / 2 + var(--header-height) / 2)
```
Atmosfearful, line 6: The fixed footer height creates a scroll bar as the footer isn't always exactly 15rem :( It's also somewhat fragile, if we add more content/icons to the footer, or change logo size or something.

Grid can do this dynamically, via `GridContainer` CSS:

```css
/* mobile */
min-height: 100vh;
grid-template-rows: 1fr auto;
  
/* desktop */
grid-template-rows: auto 1fr auto;
```

Problem: GridContainer is rendered in `_app`, so this would break some pages that rely on the implicit row sizes.
So maybe you can try wrapping just this page in a div and applying the grid?

```css
.404PageWrapper {
  display: grid;
  grid-column: full;
  grid-template-columns: inherit;
  min-height: 100vh;
  grid-template-rows: 1fr auto;
}
```

Atmosfearful (changes requested):

site/components/pages/errors/styles.ts
ladytrekker: Thanks for the grid. I start liking it more and more. Is changed.

site/pages/404.tsx
ladytrekker: done

Atmosfearful (approved):

PR #338, opened 2022-04-25
Script: Extract source "en" translations only for local development

ladytrekker:
## Description

This PR does:
- add a new script `npm run extract-strings:dev` which extracts the source translations for "en" only
- ~remove the error catch for "can not process files" as this script runs on GH servers only!~
- => let´s keep this one for now to at least stop the github action workflow run if an error happens

This new script should help developers for local development and extract the source language only.
And should **prevent developers from pushing other language files** as these are handled by Github actions anyways.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,19 @@ A set of NPM Workspace commands are provided, and can be run from the root folde
 
 `npm run format-all`: Format all files with `prettier`.
 
+`npm run extract-strings:dev`: Extract translation files for the source language "en".
+
 ## Translations
+
+### For Developers
+
+This repo uses [lingui](https://lingui.js.org/tutorials/react.html) in combination with [translation.io](https://translation.io).
+
+We follow these rules:
+
+- always add an ID to each translation
+- extract the source translation with `npm run extract-strings:dev`
+- commit source language files "en" only
 
 ### For translators
 
--- a/app/package.json
+++ b/app/package.json
@@ -7,7 +7,7 @@
     "start": "next start",
     "export": "next export",
     "lint": "next lint",
-    "lingui:extract": "cmd=\"NODE_ENV=test npx lingui extract --clean --overwrite 2>&1 | grep 'Cannot process file'\"; if [ -z \"$(eval $cmd)\" ]; then exit 0; else exit 1; fi",
+    "lingui:extract": "NODE_ENV=test npx lingui extract --clean --overwrite",
     "lingui:compile": "NODE_ENV=test lingui compile",
     "postinstall": "npm run lingui:compile"
   },
--- a/app/package.json
+++ b/app/package.json
@@ -7,7 +7,7 @@
     "start": "next start",
     "export": "next export",
     "lint": "next lint",
-    "lingui:extract": "NODE_ENV=test npx lingui extract --clean --overwrite",
+    "lingui:extract": "cmd=\"NODE_ENV=test npx lingui extract --clean --overwrite 2>&1 | grep 'Cannot process file'\"; if [ -z \"$(eval $cmd)\" ]; then exit 0; else exit 1; fi",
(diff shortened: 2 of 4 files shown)


Atmosfearful (commented):
@ladytrekker Is it possible for lingui:extract:dev to catch duplicate string ids like we used to do for lingui:extract? It's a very easy mistake to make and by default it fails silently.

ladytrekker:
> @ladytrekker Is it possible for lingui:extract:dev to catch duplicate string ids like we used to do for lingui:extract? It's a very easy mistake to make and by default it fails silently.

This is already integrated with `--clean`
https://lingui.js.org/ref/cli.html#cmdoption-extract-clean

See here: https://github.com/KlimaDAO/klimadao/pull/338/files#diff-35f15d66f9b39b506a11ac9b5c6b9a80e11b41bfd56832ab130372902258ccf7R9

Atmosfearful:
> > @ladytrekker Is it possible for lingui:extract:dev to catch duplicate string ids like we used to do for lingui:extract? It's a very easy mistake to make and by default it fails silently.
> 
> This is already integrated with `--clean` https://lingui.js.org/ref/cli.html#cmdoption-extract-clean
> 
> See here: https://github.com/KlimaDAO/klimadao/pull/338/files#diff-35f15d66f9b39b506a11ac9b5c6b9a80e11b41bfd56832ab130372902258ccf7R9

The docs don't say anything about that 🤔 To be clear I mean, duplicate string ids should throw an error and stop the build:
```html
<Trans id="copypasteme">I am text</Trans>
<Trans id="copypasteme">I am problematic because my id is already used</Trans>
```
If you can confirm that this happens then feel free to merge :)

ladytrekker:
> > > @ladytrekker Is it possible for lingui:extract:dev to catch duplicate string ids like we used to do for lingui:extract? It's a very easy mistake to make and by default it fails silently.
> > 
> > 
> > This is already integrated with `--clean` https://lingui.js.org/ref/cli.html#cmdoption-extract-clean
> > See here: https://github.com/KlimaDAO/klimadao/pull/338/files#diff-35f15d66f9b39b506a11ac9b5c6b9a80e11b41bfd56832ab130372902258ccf7R9
> 
> @ladytrekker The docs don't say anything about that hm 🤔 To be clear I mean, duplicate string ids should throw an error and stop the build:
> 
> ```
> <Trans id="copypasteme">I am text</Trans>
> <Trans id="copypasteme">I am problematic because my id is already used</Trans>
> ```
> 
> If you can confirm that this happens then feel free to merge :)

Ah that´s what you mean.

Yeah, it does not throw when the same message ID exists.
It just logs "Cannot process file .... Different defaults for the same message ID." and continues.

I will postpone this PR and improve it to:
- extract en AND en-pseudo only
- throw if same message ID exists
- log the errors

ladytrekker:
From my perspective, the extract command is broken with or without the stout hack around:

- with `cmd=\"NODE_ENV=test npx lingui extract --clean --overwrite  2>&1 | grep 'Cannot process file'\"; if [ -z \"$(eval $cmd)\" ]; then exit 0; else exit 1; fi`
- without `NODE_ENV=test npx lingui extract --clean --overwrite`

In both cases the PO files gets updated even if doubled message IDs exist.
This means when the Github action runs the files gets pushed to staging in any case.
If doubled message IDs exists then these just gets removed. Thats it.

Try it out yourself

ladytrekker:
@Atmosfearful 
I re-added the error code catch on the script command.
This ensures that on GH actions the workflow does stop.

I added a follow-up ticket for the issues mentioned above: https://github.com/KlimaDAO/klimadao/issues/348

I would vote for merging this anyways to have a local script for extracting "en" at least

Atmosfearful (approved):

PR #347, opened 2022-04-27
Fix missing strings for input placeholders

ladytrekker:
## Description

Dynamically loading message IDs is not working because: https://github.com/lingui/js-lingui/issues/1054#issuecomment-830687747

Therefore the message IDs have to exist before rendering the Translation.
If this is not the case, lingui does not find these messages. 
And **deletes the former ones** which happened here:
https://github.com/KlimaDAO/klimadao/commit/1633550d6de4a185b08bc9d5da8e2f19f0475d5d

This PR contains a fix with defining the possible messages beforehand and picking the correct one dynamically depending on the state value.

**Please note that on translation.io the new translation values have to be added again.**

## Checklist

### DIFF
--- a/app/components/views/Stake/index.tsx
+++ b/app/components/views/Stake/index.tsx
@@ -25,7 +25,7 @@ import {
   TextInfoTooltip,
 } from "@klimadao/lib/components";
 import { trimWithPlaceholder, concatAddress } from "@klimadao/lib/utils";
-import { Trans } from "@lingui/macro";
+import { Trans, defineMessage } from "@lingui/macro";
 import { BalancesCard } from "components/BalancesCard";
 import { RebaseCard } from "components/RebaseCard";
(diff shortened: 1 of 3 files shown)


ladytrekker:
@Atmosfearful FYI ! I am merging this PR to sync with translation.io

ShimonD-KlimaDAO:
@ladytrekker this doesn't seem to be working.  Is this because the Preview was created before the strings were translated?

Looking at the Preview, I see for German:
image


I also see that these strings are translated:
image

ladytrekker:
@ShimonD-KlimaDAO Thanks for reporting.

The preview link of this PR can not contain the translation strings.

You can find the translations working on staging and - because the automated merge with production ran this morning - also on production:

https://app.klimadao.finance/#/stake

PR #359, opened 2022-04-28
added checkbox for autostake on bond page and added param to redeem c…

### LINKED ISSUE #64: [bonds] Add "Auto-stake for sKLIMA" checkbox to bond redemption view
# Requirements
* [ ] checkbox appears in the Redeem view for any bond
* [ ] when checked, pass `true` as the second argument to the `redeem()` contract method
* [ ] checkbox can be focused with Tab and toggled with Spacebar (or Enter, or both, whatever the browser default is?)

If you don't have any LP to spend @Atmosfearful can test for you :)

goatsOnDrugs:
…ontract call

## Description
I added a checkbox to to bond page that will autostake upon redeeming a bond. It passes in a required prop to `redeemTransaction` and then passes that param to the `redeem` contract call. The check box will show as long as the view is "redeem" and loading is false. We may want to disable it if the bond is not claimable?

## Related Ticket

Issue #64 

Resolves #64 

## Checklist

- [x ] Building the site with `npm run build-site` works without errors
- [x ] Building the app with `npm run build-app` works without errors
- [x ] I formatted JS and TS files with running `npm run format-all`

### DIFF
--- a/app/actions/bonds.ts
+++ b/app/actions/bonds.ts
@@ -344,9 +344,11 @@ export const redeemTransaction = async (params: {
   bond: Bond;
   provider: providers.JsonRpcProvider;
   onStatus: OnStatusHandler;
+  shouldAutostake: boolean;
 }) => {
   try {
-    const autostake = false;
+    // update this line if we change prop to required
+    const autostake = params.shouldAutostake;
     const signer = params.provider.getSigner();
     const contractAddress = getBondAddress({ bond: params.bond });
     const contract = new ethers.Contract(
--- a/app/actions/bonds.ts
+++ b/app/actions/bonds.ts
@@ -347,8 +347,6 @@ export const redeemTransaction = async (params: {
   shouldAutostake: boolean;
 }) => {
   try {
-    // update this line if we change prop to required
-    const autostake = params.shouldAutostake;
     const signer = params.provider.getSigner();
     const contractAddress = getBondAddress({ bond: params.bond });
     const contract = new ethers.Contract(
@@ -357,7 +355,7 @@ export const redeemTransaction = async (params: {
       signer
     );
     params.onStatus("userConfirmation", "");
-    const txn = await contract.redeem(params.address, autostake);
+    const txn = await contract.redeem(params.address, params.shouldAutostake);
     params.onStatus("networkConfirmation", "");
     await txn.wait(1);
     params.onStatus("done", "Bond redeemed successfully");
--- a/app/components/views/Bond/index.tsx
+++ b/app/components/views/Bond/index.tsx
@@ -39,6 +39,7 @@ import { RootState, useAppDispatch } from "state";
 import { setBondAllowance } from "state/user";
 import { redeemBond, setBond } from "state/bonds";
 import InfoOutlined from "@mui/icons-material/InfoOutlined";
+import { Box, Checkbox } from "@mui/material";
 
 import * as styles from "./styles";
 import { BondBalancesCard } from "components/BondBalancesCard";
@@ -125,6 +126,7 @@ export const Bond: FC<Props> = (props) => {
 
   const [view, setView] = useState("bond");
   const [quantity, setQuantity] = useState("");
+  const [shouldAutostake, setShouldAutostake] = useState(false);
   const debouncedQuantity = useDebounce(quantity, 500);
 
   const { currentBlock, blockRate } = useSelector(selectAppState);
@@ -222,6 +224,8 @@ export const Bond: FC<Props> = (props) => {
     }
   };
 
+  const handleAutostakeCheck = () => setShouldAutostake(!shouldAutostake)
+
   const handleBond = async () => {
     try {
       if (
@@ -276,6 +280,7 @@ export const Bond: FC<Props> = (props) => {
         bond: props.bond,
         provider: props.provider,
         onStatus: setStatus,
+        shouldAutostake,
       });
       dispatch(
         redeemBond({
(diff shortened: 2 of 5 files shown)


app/actions/bonds.ts
```diff
@@ -344,9 +344,11 @@ export const redeemTransaction = async (params: {
   bond: Bond;
   provider: providers.JsonRpcProvider;
   onStatus: OnStatusHandler;
+  shouldAutostake: boolean;
 }) => {
   try {
-    const autostake = false;
+    // update this line if we change prop to required
+    const autostake = params.shouldAutostake;
```
Atmosfearful, line 351: ha now that my ugly hardcoded var is gone we can just pass `params.shouldAutostake` directly to the contract function below

app/components/views/Bond/index.tsx
```diff
@@ -39,6 +39,7 @@ import { RootState, useAppDispatch } from "state";
 import { setBondAllowance } from "state/user";
 import { redeemBond, setBond } from "state/bonds";
 import InfoOutlined from "@mui/icons-material/InfoOutlined";
+import { Box, Checkbox } from "@mui/material";
```
Atmosfearful, line 42: Something to get used to, VSCode auto-import will do this, but for better tree-shaking we need to import individually:

```
import Box from "@mui/material/Box";
import Checkbox from "@mui/material/Checkbox";
```

app/components/views/Bond/index.tsx
```diff
@@ -749,10 +754,39 @@ export const Bond: FC<Props> = (props) => {
...
+                {view === "redeem" && !showSpinner && (
+                  <Box
+                    sx={{
+                      display: "flex",
+                      width: "100%",
+                      alignItems: "center",
+                      justifyContent: "center",
+                      paddingRight: "16px",
+                    }}
+                  >
```
Atmosfearful, line 773: Don't think we need `Box`, prefer plain divs and plain css in `styles.ts`.

I like to keep styles out of JSX/js files as best as I can :)

app/components/views/Bond/index.tsx
```diff
@@ -749,10 +754,39 @@ export const Bond: FC<Props> = (props) => {
...
+                  className={styles.submitButton}
+                />
+                {view === "redeem" && !showSpinner && (
+                  <Box
+                    sx={{
+                      display: "flex",
+                      width: "100%",
+                      alignItems: "center",
+                      justifyContent: "center",
+                      paddingRight: "16px",
```
Atmosfearful, line 771: We use `rem` everywhere because it scales with the user's font-size. I put an override in the root stylesheet so that 1rem is always 10px.
```suggestion
                      paddingRight: "1.6rem",
```

app/components/views/Bond/index.tsx
```diff
@@ -749,10 +754,39 @@ export const Bond: FC<Props> = (props) => {
...
+                    }}
+                  >
+                    <Checkbox
+                      checked={shouldAutostake}
+                      onChange={handleAutostakeCheck}
+                      sx={{ color: "#fff" }}
```
Atmosfearful, line 777: we will be adding lightmode/darkmode to the dapp soon, so always use color variables, in this case I think it's `--font-01`
https://github.com/KlimaDAO/klimadao/blob/c15c6fcf288d5bc6eafad69316ccda6f727da40a/lib/theme/variables.css#L34

app/components/views/Bond/index.tsx
```diff
@@ -749,10 +754,39 @@ export const Bond: FC<Props> = (props) => {
...
+                    />{" "}
+                    <Text t="caption" align="center">
+                      <Trans
+                        id="bond.should_autostake"
+                        comment="should autostake?"
+                      >
+                        Autostake
```
Atmosfearful, line 784: trying to think if something more translate-able.... Let's go with:
```suggestion
  Automatically stake for sKLIMA
```

Atmosfearful (changes requested):

Atmosfearful:
Aha we need to increment the sKLIMA balance instead of KLIMA balance.

First add an optional autostake boolean to the action type here:
https://github.com/KlimaDAO/klimadao/blob/2813b1785c23d97e61c7d2060d74b8c509d5f80c/app/state/bonds/index.ts#L41
```ts
redeemBond: (s, a: PayloadAction<{ bond: Bond; value: string; autostake?: boolean }>) => { 
```

And then we can use that here to decide which value gets incremented, sKLIMA or KLIMA
https://github.com/KlimaDAO/klimadao/blob/188b07f3c333090ca23dbb04a670418b681b94d6/app/state/user/index.ts#L223

app/actions/bonds.ts
goatsOnDrugs: oops, fixed.

app/components/views/Bond/index.tsx
goatsOnDrugs: updated, I will keep this in mind for the future.

app/components/views/Bond/index.tsx
goatsOnDrugs: updated, I will keep this in mind for the future.

app/components/views/Bond/index.tsx
goatsOnDrugs: updated, I will keep this in mind for the future.

app/components/views/Bond/index.tsx
goatsOnDrugs: updated, I will keep this in mind for the future.

app/components/views/Bond/index.tsx
goatsOnDrugs: sounds good

app/state/user/index.ts
```diff
@@ -220,7 +220,10 @@ export const userSlice = createSlice({
   extraReducers: (builder) => {
     builder.addCase(redeemBond, (s, a) => {
       if (!s.balance) return;
-      s.balance.klima = safeAdd(s.balance.klima, a.payload.value);
+      s.balance.klima = safeAdd(
```
Atmosfearful, line 223: When autostake is `true` we need to increment the `sklima` balance. Here you only increment `klima` balance

Atmosfearful (changes requested):

Atmosfearful (approved):

ShimonD-KlimaDAO:
It's not working for me. A few notes below.

1. The functionality of the checkbox not working. 
Small redeem amount with autostake:
image

Without autostake (after this I get a MM prompt to confirm and everything goes through and I get my Klima)
image

2. The checkbox text display is "bond.should_autostake" - shouldn't it be something more user friendly?

3. This is a minor note, and I promise not to be a pain in the butt going forward. I'd generally recommend using the convention of a present imperative tense in the title and to keep it shorter — for example, use “Fix dashboard bug” instead of “Fixed” or “Fixes,” and then include a detailed description as you already have.

PR #366, opened 2022-05-02
Pass Retirement from App to Site and generate dynamic Pages

ladytrekker:
## Description

This PR does:
- App: get the total retirement number of an address
- App: set the total retirement number to the offset state
- App: Show external Link to Site with the beneficiary address and total retirement number **(Staging only)**
- Site: Generate dynamic pages from URL `retirement/<ADDRESS>` **(Staging only)**
- Site: Generate dynamic pages from URL `retirement/<ADDRESS>/<INDEX>` **(Staging only)**

The implemented contract methods on Site are **placeholders** until we get the actual GraphQL queries working from subgraph. **See the list of further steps below.**

## HOW TO TEST

1. Do an Offset here:
https://klimadao-app-git-retirement-pages-klimadao.vercel.app/#/offset
2. Wait for the Success Modal
3. In the Success Modal at the bottom copy the URL from the link "Open retirement details on: LINK`
4. In the copied URL replace `https://klimadao.finance/<ADDRESS>/<INDEX>` with the URL from this PR preview link for SITE: `https://klimadao-site-git-retirement-pages-klimadao.vercel.app/<ADDRESS>/<INDEX>`
5. Paste this new URL into your browser and let NextJS generate the result for you

## Further steps
 
- [x] Remove the contract methods used in Site for generating the pages **=>** [see PR](https://github.com/KlimaDAO/klimadao/pull/367)
- [x] Instead query the subgraph with ADDRESS and INDEX from [this query](https://thegraph.com/hosted-service/subgraph/klimadao/polygon-bridged-carbon?query=Recent%20KlimaDAO%20Retirements) **=>**  [see PR](https://github.com/KlimaDAO/klimadao/pull/367)
- [x] Pass the data to the dynamic pages [ **=>**  see PR](https://github.com/KlimaDAO/klimadao/pull/367)
- [ ] Style and build the pages according to Figma Designs

## Related Ticket

Part of https://github.com/KlimaDAO/klimadao/issues/345

## Checklist

- [x] Building the site with `npm run build-site` works without errors
- [x] Building the app with `npm run build-app` works without errors
- [x] I formatted JS and TS files with running `npm run format-all`

Issue #348, opened 2022-04-27
Translations: Add script for extracting "en" and "pseudo" for developers


ladytrekker:
Follow-up on https://github.com/KlimaDAO/klimadao/pull/338

- `npm run extract-strings:dev` should extract locale "en" and "en-pseudo"
- => with the extract cli command not possible, need to write a custom script for passing more than one locale
- the script should fail if doubled message IDs have been found 
- => script does not stop when errors happen, opened a discussion question on their repo here: https://github.com/lingui/js-lingui/discussions/1235
- => otherwise wrap errors in own handler and exit

Further ideas:

- run this script before every git push on local dev machine?
- run this script on every opened PR to check if the extract-strings script would not fail on staging?

PR #368, opened 2022-05-04
Fix extract strings with same message ID

ladytrekker:
## Description

See here that the latest GH actions are failing: https://github.com/KlimaDAO/klimadao/actions

Reason is [this PR](https://github.com/KlimaDAO/klimadao/pull/366) which introduced same translation message IDs in different places. In this case lingui "Cannot process file" which [is caught in the actual script](https://github.com/KlimaDAO/klimadao/blob/staging/site/package.json#L9) which runs on the GH servers and exits the action run.

This PR contains the fix for the same message IDs.
I ran `npm extract-strings:dev` too which we should include in a check for developers in general for the future.
=> see this [ticket](https://github.com/KlimaDAO/klimadao/issues/348)

## Checklist

### DIFF
--- a/app/locale/en/messages.po
+++ b/app/locale/en/messages.po
@@ -13,19 +13,19 @@ msgstr ""
 "Language-Team: \n"
 "Plural-Forms: \n"
 
-#: components/views/Offset/index.tsx:636
+#: components/views/Offset/index.tsx:643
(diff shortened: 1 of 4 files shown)


ladytrekker:
@Atmosfearful **FYI**

I merge this fix to make the Github actions run again.

Issue #362, opened 2022-04-29
.klima domain and ENS domain name integration


Atmosfearful:
Designs needed!

Requirements:
In the dapp, when a user connects their wallet, we can show their .klima profile and domain.
If they don't have a .klima, show their ENS.
If they don't have ENS, show the public address (current implementation)

Other considerations:
- Should we still show the public address somewhere?
- What if they have an ENS or Klima domain, but no profile pic?

tempe-techie:
- `.klima` contract address on Polygon: `0xe8b97542A433e7eCc7bB791872af04DF02A1a6E4`
- in order to get user's `.klima` name (reverse-lookup), call the `defaultNames` method: `defaultNames(address)`
- if user holds multiple domain names and wants to change their default name, they can call this function: `editDefaultDomain(domainName)` (note that only name is added here without TLD, so if domain is `timmy.klima`, only enter `timmy` as the input)
- in order to check which address holds a certain domain name (lookup), call this method: `getDomainHolder(domainName)`

Let me know if you'll have any questions 🙂

Issue #361, opened 2022-04-28
[Website] Flagged as Phishing


ShimonD-KlimaDAO:
The website is getting flagged by multiple antivirus and phishing screeners as reported by multiple users.

This may be related to the URL redirect from dapp to app, since @Atmosfearful confirmed that the normal domain is fine.

**Examples:**

1. https://phishtank.org/phish_detail.php?phish_id=7494506
2. 
![image](https://user-images.githubusercontent.com/99001851/165855574-ef5409b1-f942-4137-9f71-030e357e06a3.png)

3. https://www.virustotal.com/gui/url/6d5b90966cc520b592edf33a7c1dc6d64d2f2d6cb608deb837cc1f4cc5eb5562

4. 
![image](https://user-images.githubusercontent.com/99001851/165855641-321ba1fc-1f93-4c86-b491-37c8792d0395.png)

ShimonD-KlimaDAO:
Atmos did the following:
1. Sent a support email to vercel and to Valkyrie.comodo.com.
2. Submitted reclassification request to Trend Micro.
3. Submitted a ticket to Fortiguard.

Atmosfearful:
![image](https://user-images.githubusercontent.com/88635679/165990406-f0a85b95-a176-49c5-a052-47ca4d0e322a.png)

Atmosfearful:
![image](https://user-images.githubusercontent.com/88635679/165990688-e1707f53-a94a-4ca4-a2c7-bc3bc4528af6.png)

Atmosfearful:
I filed a request on the forums here:
https://forums.comodo.com/comodo-valkyrie-fls/report-false-positives-for-valkyrie-service-here-2022-t127715.0.html;msg915116#msg915116

Atmosfearful:
There are still 4 vendors flagging us as malicious:
![image](https://user-images.githubusercontent.com/88635679/165992292-a72eae3b-85c9-4753-a302-45a6d0de159b.png)

goatsOnDrugs:
Reproduced on windows 11 (chrome and firefox) brave seems to be working fine

Atmosfearful:
I submitted a request to Sophos to recategorize, their documentation is very unclear. I am also reaching out to them on twitter.

![image](https://user-images.githubusercontent.com/88635679/165995915-d1ac8b0e-97b9-419e-9b21-80d79ed3926d.png)

Atmosfearful:
[...]

PR #372, opened 2022-05-05
#361 klima and ens domains 

goatsOnDrugs:
## Description
Okay, we are first checking if the user has a klima domain and using that if it exists. if not then we grab the ens domain. if no ens then we use address. I have not tested the ens bc I dont own one but Chaz is going to help with that.

Questions: We need to decide what to do if the domain forward lookup check fails
what is a graceful fallback if default image fails?
should I move functions in actions/user to the utils file?

## Related Ticket

#362 

Resolves #<issue-number>
#362 

## Checklist


app/actions/user.ts
```diff
@@ -81,6 +85,19 @@ export const loadAccountDetails = (params: {
         IERC20.abi,
         params.provider
       );
+      // remember to add this to addresses object
+      const klimaDomainContract = new ethers.Contract(
+        "0xe8b97542A433e7eCc7bB791872af04DF02A1a6E4",
```
Atmosfearful, line 90: pls export from the addresses object in @klimadao/lib/constants, leave empty string on testnet if none exists

app/actions/utils.ts
```diff
@@ -32,3 +33,90 @@ export const getStatusMessage = (status: AppNotificationStatus) => {
...
+  contract: Contract;
+}): Promise<Domain> => {
+  const domain: any = {};
+
+  try {
+    const domainName = await params.contract.defaultNames(params.address);
+    // what do we do if this is false?
+    const isNameVerified =
+      (await params.contract.getDomainHolder(domainName)) === params.address;
+    if (!isNameVerified) null;
```
Atmosfearful, line 47:
```suggestion
    if (!isNameVerified) return null;
```

app/actions/utils.ts
```diff
@@ -32,3 +33,90 @@ export const getStatusMessage = (status: AppNotificationStatus) => {
...
+    domain.name = `${domainName}.klima`;
+    const customImage = await params.contract.getDomainData(domainName);
+    if (customImage && JSON.parse(customImage).imgAddress) {
+      domain.image = JSON.parse(customImage).imgAddress;
```
Atmosfearful, line 51:
```suggestion
    if (customImage) {
      const data = JSON.parse(customImage);
      domain.image = data?.imgAddress;
```

This is probably not a costly parse but just for good practice, parse once and fallback to undefined or something?

app/actions/utils.ts
```diff
@@ -32,3 +33,90 @@ export const getStatusMessage = (status: AppNotificationStatus) => {
...
+    const avatar = ethProvider.getAvatar(ensDomain);
+    ens.avatar = avatar;
+  }
+  if (ensDomain) {
+    ens.name = `${ensDomain}.eth`;
+  }
+  return ens;
+};
+
+const parseURL = (url: string): any => {
```
Atmosfearful, line 86: can you leave a what/why comment, i can't really  understand what this is doing lol

also lets avoid `any` return type, looks like you can make an interface for `parsed`

app/components/NavMenu/index.tsx
```diff
@@ -96,12 +97,32 @@ interface Props {
...
+  const domains = useSelector(selectDomain);
+  const [domain, setDomain] = useState<Domain | undefined>(undefined);
+  const svg = useRef(null);
+  useEffect(() => {
+    if (domains?.knsDomain.name) {
+      setDomain(domains.knsDomain);
+    } else if (domains?.ensDomain.name) {
+      setDomain(domains.ensDomain);
+    }
+  }, [domains]);
```
Atmosfearful, line 109: no need to set to a secondary state `domain`, the useSelector hook should already trigger a rerender whenever `state.user.nameServiceDomains` changes

app/components/NavMenu/index.tsx
```diff
@@ -96,12 +97,32 @@ interface Props {
...
   const handleHide = () => {
     props.onHide?.();
   };
-
+  const htmlToElement = (htmlString: string): Element => {
+    const template = document.createElement("template");
+    htmlString = htmlString.trim(); // Never return a text node of whitespace as the result
+    template.innerHTML = htmlString;
+    return template.content.firstChild as Element;
+  };
```
Atmosfearful, line 125: did you try this:

```jsx
<div dangerouslySetInnerHTML={{ __html: domain.defaultImage }} />
```

It should be safe enough because
1. it only works if it's the user's own kns/ens domain via their connected wallet
2. its the default image, not user generated

Atmosfearful (changes requested):
Nice work so far, this looked tricky!

ShimonD-KlimaDAO:
Here is a wallet of mine that doesn't have a .klima nor a .ens domain:
image

Currently there is an empty ".klima".
I think it would be better to leave this area blank or have a link saying something like "Get Your KNS Domain" with a pretty Klima logo. The default logo is illegible (even when zooming in) and doesn't seem meaningful.

Here is a wallet with my ShimonD.klima domain (my test wallet). It looks good!
image

ShimonD-KlimaDAO:
I asked a friend to check this with an ens domain, and he sent me a screenshot that looks exactly the same as my test with neither a KNS nor an ENS domain - a ".klima" only, instead of the ENS domain.

app/actions/user.ts
goatsOnDrugs: fixed

app/actions/utils.ts
goatsOnDrugs: fixed

app/actions/utils.ts
goatsOnDrugs: updated. I left the check for imageUrl because this call can return a twitter url or something and the user can also change they imgAddress key.

app/components/NavMenu/index.tsx
goatsOnDrugs: cool, yea good point. updated.

app/actions/utils.ts
goatsOnDrugs: updated

app/actions/utils.ts
goatsOnDrugs: updated

app/components/NavMenu/index.tsx
goatsOnDrugs: Ok, I made the address section its own functional component to clean this up.

goatsOnDrugs:
The bug @ShimonD-KlimaDAO mentioned should be fixed in this next commit

Atmosfearful:
Can you confirm that it's fixed @ShimonD-KlimaDAO

ShimonD-KlimaDAO:
Update:

1. No KNS, no ENS: shows the address as before
2. KNS exists: shows the KNS domain without any icon (different than yesterday):
image

**Note** - Should the KNS domain name link to the KNS page? I think that would be nice. It's currently just plain text.

3. No KNS, yes ENS: I will post an update later when I reach a friend who has an ENS domain to help me test.

ShimonD-KlimaDAO:
Everything is consistent now!

I still see a little space in front of the domains (both KNS and ENS), but it's probably fine:
image

image


This small space does not happen when the wallet has neither KNS nor ENS:
image


Also - just as a note, I don't have access to a wallet that has **both** an ENS and a KNS domain, so I was not able to test the combination of the two (I would expect KNS to show)

goatsOnDrugs:
I fixed the padding on the left of the name when no image is available but the desired state is the default image from klima name service should show.

ShimonD-KlimaDAO:
> 

Looks like I'm still getting the same. This is on Chrome. And blankslate was also seeing this with ENS.
image

ShimonD-KlimaDAO:
Looks good now and can be merged!

image

goatsOnDrugs:
@Atmosfearful got this working!

Atmosfearful (approved):

PR #374, opened 2022-05-05
Infura rpc requests

### LINKED ISSUE #231: Use Infura for server-side / build-blocking RPC requests
# Requirements
- [x] Add `process.env.INFURA_ID` to env vars in vercel. ✅ 
- [ ] Export from `site/lib/constants` similar to other env vars.
- [ ] Use this ID to construct the infura endpoint `https://polygon-mainnet.infura.io/v3/${INFURA_ID}`
- [ ] Fallback to polygon-rpc instead of infura endpoint, if no INFURA_ID is available. This is the fallback for devs on localhost, as this Id only exists on vercel.
- [ ] use this endpoint to create the `jsonRPCProvider` with `ethers`, for all the functions that run on the serverside. Ensure a clear deliniation for functions that are called on the server and rely on infura-- we don't want these being used client side now or in the future.

Vienio99:
## Related Ticket

Resolves #231

## Checklist

- [ x ] Building the site with `npm run build-site` works without errors
- [ x ] Building the app with `npm run build-app` works without errors
- [ x ] I formatted JS and TS files with running `npm run format-all`


lib/utils/getJsonRpcProvider/index.ts
```diff
@@ -1,5 +1,12 @@
+import { INFURA_ID } from "@klimadao/site/lib/constants";
 import { ethers } from "ethers";
+import { urls } from "../../constants";
 
-export const getJsonRpcProvider = () => {
+export const getJsonRpcProvider = (provider: string) => {
```
Atmosfearful, line 5: looks like you pass in the provider string but don't actually use it. We can simplify this:
use the params syntax:
https://github.com/KlimaDAO/klimadao/wiki#prefer-params-objects-over-multiple-arguments

And call it `infuraId`

Then the code looks something like:
```ts
if (params.infuraId) {
  return new ethers.providers.JsonRpcProvider(
      `${urls.infuraRPC}${params.infuraId}`
    );
}
return new ethers.providers.JsonRpcProvider(urls.polygonMainnetRpc);
```

lib/utils/getStakingRewards/index.ts
```diff
@@ -9,7 +9,7 @@ export const getStakingRewards = async (
   days: number,
   blockRate: number
 ): Promise<number> => {
-  const provider = getJsonRpcProvider();
+  const provider = getJsonRpcProvider(urls.infuraRpc);
```
Atmosfearful, line 12: once you implement my comment above, you will need to be passing down `INFURA_ID` rather than the entire url.

```ts
const provider = getJsonRpcProvider({ infuraId: INFURA_ID });
```

lib/utils/getTreasuryBalance/index.ts
```diff
@@ -1,12 +1,14 @@
...
 import PairContract from "../../abi/PairContract.json";
 import { addresses } from "../../constants";
+import { getJsonRpcProvider } from "../getJsonRpcProvider";
 
-const getOwnedBCTFromSLP = async (adr: "bctUsdcLp" | "klimaBctLp") => {
-  const provider = getJsonRpcProvider();
+const getOwnedBCTFromSLP = async (
+  adr: "bctUsdcLp" | "klimaBctLp",
+  provider: providers.JsonRpcProvider
+) => {
```
Atmosfearful, line 11: Please refactor to params syntax:
https://github.com/KlimaDAO/klimadao/wiki#prefer-params-objects-over-multiple-arguments

Atmosfearful (changes requested):

Atmosfearful:
pinging @ladytrekker for review since this touches her recent work, but first would you mind addressing these merge conflicts?

ladytrekker:
Will review when the merge conflicts are resolved. 
You can ping me @Vienio99 if you need help resolving the conflicts

PR #377, opened 2022-05-07
feat: add web3context provider

0xAeterno:
### Introduces a Web3Context provider 🎉 

Wrapping the root level app component or a subset of pages with `<Web3ContextProvider />` will allow components further down the DOM tree to access key web3 objects and functions, namely the:
- `provider` object
- `address` connected
- `network` and
- wallet `connect()` and  `disconnect()` functions.

#### Why this pattern
Connecting and context of user's wallet is a core component to interacting with any dApp.
This is visibly evident in our app, every page requires the user's web3 context which is passed down as props to every page. Fortunately, our pages aren't too deeply nested so prop drilling it's not too much of an issue, however, it does prevent us from refactoring pages into smaller compositional components to simplify the app code without resorting to prop drilling.

Using the Context pattern will simplify how we access user's web3 context in our apps and will refer to a single source of truth. It will also enable code refactors to the app at some point to improve readability and workability (that bond page 👀)

### Example usage
In the root app component or a specific subset of pages
```ts
import { Web3ContextProvider } from "hooks/useWeb3/web3context";

function MyApp({ Component, pageProps, router }: AppProps) {
 // other code
return (
  <Web3ContextProvider>
    <Component {...pageProps} />
  </Web3ContextProvider>
)}

```

Web3 auth button
```ts
import React, { FC } from "react";
import { useWeb3 } from "hooks/useWeb3";

export const Web3Auth = () => {
  const { address, connect, disconnect } = useWeb3();
  const isConnected = Boolean(address);

  return (
    {isConnected ? (
      <ButtonPrimary label={address} onClick={disconnect} />
     ) : (
      <ButtonPrimary label="Connect" onClick={connect} />
    )}
  )
}

```

Full `useWeb3()` hook API
```ts
// will probably add isConnected boolean too instead of constructing it outside the hook.
const { provider, address, network, web3Provider, connect, disconnect } = useWeb3();
```

Credits to this [blog](https://betterprogramming.pub/adding-web3-to-our-nextjs-typescript-project-861e9ed5feaf) post. I tried to integrate what we had in `app/home` into a context provider but there was a bit of jank somewhere that made it difficult :) Super stoked to have come across that blog.

Will move this into `lib` at some point so we can share it across our monorepo. Keeping it local to `site` for now so I can actively develop on it whilst we build on the pledge dashboard.

Always open to feedback of course!

## Checklist

### DIFF
--- a/site/hooks/useWeb3/index.tsx
+++ b/site/hooks/useWeb3/index.tsx
@@ -1,0 +1,140 @@
+import { useEffect, useReducer, useCallback } from "react";
+import WalletConnectProvider from "@walletconnect/web3-provider";
+import { urls } from "@klimadao/lib/constants";
+import { ethers } from "ethers";
+import WalletLink from "walletlink";
+import Web3Modal from "web3modal";
+
+import {
+  Web3ProviderState,
+  Web3Action,
+  web3InitialState,
+  web3Reducer,
+} from "./reducers";
+
+const providerOptions = {
+  walletconnect: {
+    package: WalletConnectProvider,
+    options: {
+      rpc: { 137: urls.polygonMainnetRpc },
+    },
+  },
+  walletlink: {
+    package: WalletLink,
+    options: {
+      appName: "Official KlimaDAO App",
+      rpc: urls.polygonMainnetRpc,
+      chainId: 137,
+      appLogoUrl: null,
+      darkMode: false,
+    },
+  },
+};
+
+let web3Modal: Web3Modal | null;
+if (typeof window !== "undefined") {
+  web3Modal = new Web3Modal({
+    cacheProvider: true,
+    providerOptions,
+  });
+}
+
+export const useWeb3 = () => {
+  const [state, dispatch] = useReducer(web3Reducer, web3InitialState);
+  const { provider, web3Provider, address, network } = state;
+
+  const connect = useCallback(async () => {
+    if (web3Modal) {
+      try {
+        const provider = await web3Modal.connect();
+        const web3Provider = new ethers.providers.Web3Provider(provider);
+        const signer = web3Provider.getSigner();
+        const address = await signer.getAddress();
+        const network = await web3Provider.getNetwork();
+
+        dispatch({
+          type: "SET_WEB3_PROVIDER",
+          provider,
+          web3Provider,
+          address,
+          network,
+        } as Web3Action);
+      } catch (e) {
+        console.log("Connect error:", e);
+      }
+    } else {
+      console.error("No Web3Modal");
+    }
+  }, []);
+
+  const disconnect = useCallback(async () => {
+    if (web3Modal) {
+      web3Modal.clearCachedProvider();
+      if (provider?.disconnect && typeof provider.disconnect === "function") {
+        await provider.disconnect();
+      }
+
+      dispatch({ type: "RESET_WEB3_PROVIDER" } as Web3Action);
+    } else {
+      console.error("No Web3Modal");
+    }
+  }, [provider]);
+
+  // Auto connect to the cached provider
+  useEffect(() => {
+    if (web3Modal && web3Modal.cachedProvider) {
+      connect();
+    }
+  }, [connect]);
+
+  // EIP-1193 events
+  useEffect(() => {
+    if (provider?.on) {
+      const handleAccountsChanged = (accounts: string[]) => {
+        dispatch({
+          type: "SET_ADDRESS",
+          address: accounts[0],
+        } as Web3Action);
+      };
+
+      // https://docs.ethers.io/v5/concepts/best-practices/#best-practices--network-changes
+      const handleChainChanged = (_hexChainId: string) => {
+        if (typeof window !== "undefined") {
+          console.log("Switched to chain...", _hexChainId);
+
+          window.location.reload();
+        } else {
+          console.log("Window is undefined");
+        }
+      };
+
+      const handleDisconnect = (error: { code: number; message: string }) => {
+        // eslint-disable-next-line no-console
+        console.log("disconnect", error);
+        disconnect();
+      };
+
+      provider.on("accountsChanged", handleAccountsChanged);
+      provider.on("chainChanged", handleChainChanged);
+      provider.on("disconnect", handleDisconnect);
+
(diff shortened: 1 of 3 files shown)


site/hooks/useWeb3/reducers.tsx
```diff
@@ -0,0 +1,68 @@
+import { ethers } from "ethers";
+
+export type Web3ProviderState = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  provider: any;
+  web3Provider: ethers.providers.Web3Provider | null | undefined;
+  address: string | null | undefined;
+  network: ethers.providers.Network | null | undefined;
+  connect?: () => Promise<void>;
+  disconnect?: () => Promise<void>;
```
ladytrekker, line 10: Thanks for linking to the Blog Post.
Just stumbled over `?` and `undefined` and thought that these values can never be undefined as soon as the wallet is connected.

ladytrekker (approved):
Thanks for the good description and the linked BlogPost.
I compared the hook with the current implementation on the dApp and it looks easy to integrate as soon as the hook moves to Lib.
Only main difference is that the Web3Modal is stored in a React.ref on the dApp which should not be a problem. 
Over all looks good to me.

site/hooks/useWeb3/reducers.tsx
```diff
@@ -0,0 +1,68 @@
+import { ethers } from "ethers";
+
+export type Web3ProviderState = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  provider: any;
```
Atmosfearful, line 5: `any` is too dangerous here since so many components will rely on it. Try `JsonRPCPRovider` from ethers, i believe that should work.

Atmosfearful: and remove that eslint-disable once you find a valid type

site/hooks/useWeb3/web3context.tsx
```diff
@@ -0,0 +1,29 @@
...
+    <Web3Context.Provider value={web3ProviderState}>
+      {children}
+    </Web3Context.Provider>
+  );
+};
+
+export const useWeb3Context = () => {
+  const context = useContext(Web3Context);
+  if (context === undefined) {
+    throw new Error("useWeb3 must be used within a Web3ContextProvider");
```
Atmosfearful, line 25: nice handrails for the devs!

site/hooks/useWeb3/reducers.tsx
Atmosfearful: based on the initial state it can also be null ya?

site/hooks/useWeb3/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+          window.location.reload();
+        } else {
+          console.log("Window is undefined");
+        }
+      };
+
+      const handleDisconnect = (error: { code: number; message: string }) => {
+        // eslint-disable-next-line no-console
+        console.log("disconnect", error);
```
Atmosfearful, line 113: what is this error from? i dont get it

site/hooks/useWeb3/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+    } else {
+      console.error("No Web3Modal");
+    }
+  }, []);
+
+  const disconnect = useCallback(async () => {
+    if (web3Modal) {
+      web3Modal.clearCachedProvider();
+      if (provider?.disconnect && typeof provider.disconnect === "function") {
+        await provider.disconnect();
```
Atmosfearful, line 74: we need to reset the entire app state when they disconnect (remove all balances etc).
so that's why I was calling window.reload on the old one

site/hooks/useWeb3/web3context.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { useWeb3 } from ".";
+import { Web3ProviderState, web3InitialState } from "./reducers";
+
+const Web3Context = createContext<Web3ProviderState>(web3InitialState);
+
+interface Props {
+  children: ReactChild;
+}
+
+export const Web3ContextProvider = ({ children }: Props) => {
```
Atmosfearful, line 12: for future reference I think this is what the react FC type is for:
```ts
// props.children is typed automatically
export const Web3ContextProvider: FC = (props) => {
```

Atmosfearful (changes requested):
Super cool, a couple changes.

At some point we can move this to `@klimadao/lib` and share it right? Feel free to create a github ticket and assign yourself :)

0xAeterno:
@ladytrekker @Atmosfearful
Thanks for the feedback ☺️  I'll address the comments in my current working branch and close this PR :)

site/hooks/useWeb3/web3context.tsx
0xAeterno: Thanks, that was a bit of copy pasta 😛

site/hooks/useWeb3/index.tsx
0xAeterno: no idea, i'll remove it!

site/hooks/useWeb3/index.tsx
0xAeterno: Gotcha! What do you think of defining a new disconnect function in `app` that handles the page reload?

site/hooks/useWeb3/reducers.tsx
0xAeterno: `providers.JsonRPCProvider` doesn't satisfy the types here :(

this should be typed as the return value of `web3Modal.connect()` but even web3modal package types it as `any` hahah

site/hooks/useWeb3/reducers.tsx
0xAeterno: Good pickup, I've simplified the type declarations a bit :)

site/hooks/useWeb3/reducers.tsx
Atmosfearful: 🤮 

> `providers.JsonRPCProvider` doesn't satisfy the types here :(
> 
> this should be typed as the return value of `web3Modal.connect()` but even web3modal package types it as `any` hahah

Issue #357, opened 2022-04-28
[app ux] add light mode to dapp


Atmosfearful:
* [ ] Add the button
* [ ] Ensure all colors are looking good on all views!

The selected mode will not transfer from site -> dapp. We can follow up with a query param that passes the state between websites similar to how we manage this with locales.

jabby09:
@Atmosfearful this one is 7 months old...still relevant and something worth doing?

PR #382, opened 2022-05-09
added themeToggle to app

goatsOnDrugs:
## Description
I added the theme toggle button to app. I smoke tested and didnt see anything that looked broken.

## Related Ticket

 #357 

Resolves #<issue-number>
#357 

## Checklist

### DIFF
--- a/app/components/views/Home/ThemeToggle.tsx
+++ b/app/components/views/Home/ThemeToggle.tsx
@@ -1,0 +1,16 @@
+import { FC } from "react";
+import { ThemeToggle } from "@klimadao/lib/components";
+
+// To prevent loading errors need to wrap external ESM again
+// as it is loaded dynamically later in Navigation
+// https://github.com/vercel/next.js/issues/25454#issuecomment-862571514
+
+interface Props {
+  className?: string;
+}
+
+const ThemeToggleExport: FC<Props> = (props) => (
+  <ThemeToggle className={props.className} />
+);
+
+export default ThemeToggleExport;
--- a/app/components/views/Home/index.tsx
+++ b/app/components/views/Home/index.tsx
@@ -25,6 +25,7 @@ import { InvalidNetworkModal } from "components/InvalidNetworkModal";
 import { InvalidRPCModal } from "components/InvalidRPCModal";
 import { CheckURLBanner, skipCheckURLBanner } from "components/CheckURLBanner";
 import { NotificationModal } from "components/NotificationModal";
+import dynamic from "next/dynamic";
 
 import { initLocale } from "lib/i18n";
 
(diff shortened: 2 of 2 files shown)


app/components/views/Home/ThemeToggle.tsx
```diff
@@ -0,0 +1,16 @@
...
+// as it is loaded dynamically later in Navigation
+// https://github.com/vercel/next.js/issues/25454#issuecomment-862571514
+
+interface Props {
+  className?: string;
+}
+
+const ThemeToggleExport: FC<Props> = (props) => (
+  <ThemeToggle className={props.className} />
+);
```
0xAeterno, line 14: What are your thoughts on moving `<ThemeToggleExport />` to the shared lib directory?

app/components/views/Home/index.tsx
```diff
@@ -302,6 +308,7 @@ export const Home: FC = () => {
                 onHide={() => setShowMobileMenu(false)}
               />
             </div>
+            <ThemeToggle />
```
0xAeterno, line 311: Let's override the base surface color to have the button color contrast with the background :)

image


https://github.com/KlimaDAO/klimadao/blob/8a593c81489705b6b6625cefe5805272b78ba21d/site/components/pages/Pledge/PledgeLayout/styles.ts#L30-L32

0xAeterno (changes requested):
I think we should hold off and pass this by our designers before proceeding.
IMO, the app looks really sloppy in light mode :(

We should be able to improve it by fixing up the form styling and changing the social icons colour to.. maybe a `var(--font-3)`? 🤞 

image

app/components/views/Home/ThemeToggle.tsx
goatsOnDrugs: Theme toggle is in shared lib directory but imported and re-exported in site bc of this bug ```// To prevent loading errors need to wrap external ESM again
// as it is loaded dynamically later in Navigation
// https://github.com/vercel/next.js/issues/25454#issuecomment-862571514```

goatsOnDrugs: so I did the same thing for app but if there is a better way then I'm happy to implement it.

Atmosfearful: But if its reexported isn't it the same problem as the first one? Namely that you cant reference the path in the dynamic import? `@klimadao/lib/components`

If you do need to re-export it, perhaps this saves on some boilerplate, rather than defining a new component?
`export { ThemeToggle } from "@klimadao/lib/components"`

Atmosfearful: wait actually @0xAeterno are you sure we need to wrap in dynamic import at all? can't we just do a good old check for `typeof document === 'undefined'` and fallback to light mode? then we can ditch the reexports altogether

Atmosfearful:
I created a thread on discord to discuss this with a designer, thanks for the call-out @0xAeterno

Atmosfearful:
@Go-Mori I would but the ball in your court here. If you see opportunities to choose better color pairs for certain elements (borders, text, surfaces buttons etc.) in light mode please let me know. Otherwise let me know if it's OK and I can merge!

Atmosfearful:
Hey @goatsOnDrugs closing this for now. @Go-Mori is busy with more important tasks and will likely do a bigger revamp of colors/layout in the future. So we will table this. Sorry for the go-around, I assumed our existing color pairs would look better in the dapp! 😭

Issue #10, opened 2021-11-13
(low priority) Migrate from CSS Modules to Emotion.


Atmosfearful:
I built the site initially with css-modules for development speed, as it was supported out-of-the-box with nextjs, and because I love the css-modules pattern overall.

But this was a mistake. CSS Modules is not under active maintenance, not well documented, and relies on transpilers, which means it can't be easily used in our `@klimadao/lib` if we want to publish a components library. I'd like to convert everything to Emotion, using a very similar pattern and structure (separating component files and styles files).

I have built production react apps with styled-components, css-modules, styled-jsx, and vanilla css. Emotion beats them all.

Note: we already have Emotion as a sub-dependency because we use @mui/material-icons
 
We probably want to go with the framework-agnostic `@emotion/css` https://emotion.sh/docs/introduction

dodecaKlima:
Good shout. If you are committed to CSS-in-JS, have you considered [Stitches](https://stitches.dev)?

Atmosfearful:
> Good shout. If you are committed to CSS-in-JS, have you considered [Stitches](https://stitches.dev)?

Ooh that's a new one for me, gonna take a closer look thank you for sharing!

rplusq:
Hey, have you considered other options like TailwindCSS / Chakra-UI? They have been getting a lot of traction lately, can make the onboarding for users / development times quicker too

Atmosfearful:
@goatsOnDrugs feel free to close this out if everything is done

PR #390, opened 2022-05-10
info components moved from css modules to emotion css

goatsOnDrugs:
## Description
I updated the info components to use emotion. there was a global rule but i believe it was being over-riden in home page anyways and when I got it working it made all the icons on the app small so I kept the rule local

## Related Ticket
#10 

Resolves #<issue-number>

## Checklist


app/components/views/Info/CopyAddressButton/index.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { useState, useEffect } from "react";
 import ContentCopyIcon from "@mui/icons-material/ContentCopy";
 import styles from "./index.module.css";
+// import * as styles from './styles'
```
Atmosfearful, line 4: what's this? wrong import?

app/components/views/Info/CopyAddressButton/styles.ts
```diff
@@ -0,0 +1,18 @@
+import { css } from "@emotion/css";
+
+// used !important here bc home component global css is more specific than this one
+// idk what this global was doing .copyAddressButtonIcon:global(.MuiSvgIcon-root) {
```
Atmosfearful, line 4: This was targeting a classname which is created by Mui
the :global() tells modules not to apply a random scoped classname to the selector, as random scoped classnames are the default behavior (to prevent style clashes when reusing the component).

So in emotion land I think the equivalent is this:

```ts
export const copyAddressButtonIcon = css`
  .MuiSvgIcon-root {
    height: 1.6rem;
    width: 1.6rem;
    cursor: pointer;
  }
`;
```

Atmosfearful (changes requested):

app/components/views/Info/CopyAddressButton/styles.ts
goatsOnDrugs: The `Home_container svg` class is more specific than the .MuiSvgIcon-root class. I think when we update home component to use emotion css and ```import { injectGlobal } from '@emotion/css'

injectGlobal`
  * {
    box-sizing: border-box;
  }``` this injectGlobal it might fix the specificity.

app/components/views/Info/CopyAddressButton/index.tsx
goatsOnDrugs: doh, fixed.

Atmosfearful (approved):

PR #413, opened 2022-05-16
Draft: Added translation tags to web3modal strings

### LINKED ISSUE #397: Translate web3modal strings
See how @kaxline added a string override to web3modal in this PR. We should be able to just use `t` or `trans` macros
#385

biwano:
## Description

Added translation tags to web3modal strings

## Related Ticket

Resolves #397

## Changes

| Before  | After  |
|---------|--------|
|image |image|

## Checklist

## Notes

This does not work properly!

We have to make `useEffect` depend on `locale` so the modal can be translated when the locales are loaded.
But if we change the language after the modal is loaded once (not necessarily shown), somehow it gets bugged (impossible to open and close the modal).
I suspect it has something to do with the fact that it creates multiple instances of Web3Modal. I am digging into that but I think I need another pair of eyes.

Edit: If I add this test : `if (locale !== undefined && ref.current === undefined)` I can obtain a behavior where the language is correctly loaded at application startup but the Modal text won't change if the user changes the language until they reload the application.

### DIFF
--- a/app/components/views/Home/index.tsx
+++ b/app/components/views/Home/index.tsx
@@ -36,6 +36,7 @@ import { NavMenu } from "components/NavMenu";
 import Menu from "@mui/icons-material/Menu";
 import { IsomorphicRoutes } from "components/IsomorphicRoutes";
 import { Buy } from "../Buy";
+import { t } from "@lingui/macro";
 
 type EIP1139Provider = ethers.providers.ExternalProvider & {
   on: (e: "accountsChanged" | "chainChanged", cb: () => void) => void;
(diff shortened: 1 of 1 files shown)


ShimonD-KlimaDAO:
As mentioned in the Notes, this is not working properly. French:

image

biwano:
> As mentioned in the Notes, this is not working properly. French:
> 
> image

Hi Shimon!

To see it work you would have to manually: 
- extract string
- change the french pofile
- compile string

I did not want to push changes to the po file in this commit to not mess up with the translation.io process when/if we merge.

Atmosfearful:
Update: been working on this on a local branch but ran into some webpack issues. Hope to have something soon...

Atmosfearful:
Made great progress including a refactor for how we handle the web3modal provider (so we can reinstantiate it when the locale changes). Will put up a PR soon......

biwano:
Same goal accomplished in https://github.com/KlimaDAO/klimadao/pull/480

PR #415, opened 2022-05-17
Optimized loadAccountDetails to use promise.all

naiknareshh:
## Description

As per changes suggested by @Atmosfearful at Issue #396 to optimize network calls

## Related Ticket

Issue #396 

Resolves #396

## Checklist

- [X] Building the site with `npm run build-site` works without errors
- [X] Building the app with `npm run build-app` works without errors
- [X] I formatted JS and TS files with running `npm run format-all`

PR #407, opened 2022-05-13
added nbo and ubo links to offset page

goatsOnDrugs:
## Description
Added two links on the offset page

## Related Ticket
#378

Resolves #378

## Checklist

- [x] Building the site with `npm run build-site` works without errors
- [x] Building the app with `npm run build-app` works without errors
- [x] I formatted JS and TS files with running `npm run format-all`

Issue #402, opened 2022-05-12
Enhance /retirements/ index page


Atmosfearful:
## Totals Retirement Page 

- URL: `https://www.klimadao.finance/retirements/<WALLET-ADDRESS>`

**Header**
- Title /h1: Carbon Retirements
- Subtitle: for beneficiary 0x1235 (address from URL)

**Content:**
- "They've retired XX tonnes of carbon"
- "They've got 21 Total retirements"
- "They've got XX Total tonnes claimed for NFTs"
- Breakdown: 0.0075 BCT, 1.1 NCT, 0.0003 MCO2 .....
- List all retirements for the given beneficiary
  - Ability to click one and navigate to the retirement: 
  URL: `https://www.klimadao.finance/retirements/<WALLET-ADDRESS>/<NUMBER-OF-RETIREMENT>`
  - Maybe some summary information about each retirement (tonnes, message, date?)
 
**Share Buttons**
- Copy URL
- Share on Social Media (what happens here when clicked?)
 
**Footer**
- Link to the index page from the app/#/offset view (perhaps from the card that shows retirement stats on the right hand column?)
- Same Footer as for the Single Retirement Page which includes all relevant text and Buttons ?

ShimonD-KlimaDAO:
I would also suggest adding "share on social media" functionality. This could be extremely beneficial from a marketing perspective.

PR #416, opened 2022-05-17
Refactor Retirement contract lib helpers

ladytrekker:
## Description

We should use the same contract methods in all places for site and app so as not to loose data.

Especially [since this PR here](https://github.com/KlimaDAO/klimadao/pull/407) which expanded the possible retirement tokens, which is not reflected on the Pledge site as it was using a copied and older version of the same contract method.
Same can be done later for the balances after this refactoring is merged: https://github.com/KlimaDAO/klimadao/pull/415

This is also a preparation for the Totals Retirement page on Site ([#402](https://github.com/KlimaDAO/klimadao/issues/402))

## Checklist


Atmosfearful (approved):

lib/utils/getRetirement/index.ts
```diff
@@ -25,33 +25,54 @@ export const createRetirementStorageContract = (
...
+      params.address,
+      addresses["mainnet"].nct
+    );
+    const ubo = await retirementStorageContract.getRetirementPoolInfo(
+      params.address,
+      addresses["mainnet"].ubo
+    );
+    const nbo = await retirementStorageContract.getRetirementPoolInfo(
+      params.address,
+      addresses["mainnet"].nbo
```
Atmosfearful, line 60: Please refactor to promise.all and then it's good to merge!

Atmosfearful (approved):

lib/utils/getRetirement/index.ts
ladytrekker: done

PR #417, opened 2022-05-17
New infura rpc requests

### LINKED ISSUE #231: Use Infura for server-side / build-blocking RPC requests

Vienio99:
## Description

New PR for this issue as previous one's commit history was a bit messed up
Old one PR no. #374

## Related Ticket

Resolves #231 

- [ x ] Building the site with `npm run build-site` works without errors
- [ x ] Building the app with `npm run build-app` works without errors
- [ x ] I formatted JS and TS files with running `npm run format-all`

### DIFF
--- a/app/actions/app.ts
+++ b/app/actions/app.ts
@@ -8,7 +8,7 @@ import {
   getTreasuryBalance,
   fetchBlockRate,
 } from "@klimadao/lib/utils";
-import { addresses } from "@klimadao/lib/constants";
+import { addresses, urls } from "@klimadao/lib/constants";
 import DistributorContractv4 from "@klimadao/lib/abi/DistributorContractv4.json";
 import SKlima from "@klimadao/lib/abi/sKlima.json";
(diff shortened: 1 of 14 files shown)


Atmosfearful (approved):
Nice! Hopefully this makes our builds more stable and decreases the time it takes to generate new pages on the server!

PR #439, opened 2022-05-23
Fix #432 - add better form validation for selective retirements

### LINKED ISSUE #432: Better string validation for Selective retirement address inputs
In the offset tool, if there is a space in the selective retirement address or other mistake:
* [ ] indicate to the user with red outline
* [ ] prevent submission of the transaction

We can use the `ethers.isAddress` util that I just learned about.

goatsOnDrugs:
## Description
I used the ethers isAddress util to test the addresses and added an array of booleans to `specificAddresses` for conditional styles on the input. I left the :focused the same so it doesnt flash at the user while typing. One questio is for the button props what priority do we want thew address format error to be? before or after the "insufficient funds"and other errors?

## Related Ticket
#432 

Resolves #432

## Checklist

### DIFF
--- a/app/components/views/Offset/index.tsx
+++ b/app/components/views/Offset/index.tsx
@@ -120,13 +120,9 @@ export const Offset = (props: Props) => {
   const [cost, setCost] = useState("");
   const [beneficiary, setBeneficiary] = useState("");
   const [beneficiaryAddress, setBeneficiaryAddress] = useState("");
-  const [beneficiaryAddressError, setBeneficiaryAddressError] = useState(false);
   const [retirementMessage, setRetirementMessage] = useState("");
   // for selective retirement
-  const [specificAddresses, setSpecificAddresses] = useState({
-    addresses: [""],
-    errors: [false],
-  });
+  const [specificAddresses, setSpecificAddresses] = useState([""]);
   const [retirementTransactionHash, setRetirementTransactionHash] =
     useState("");
   const [retirementTotals, setRetirementTotals] = useState<number | null>(null);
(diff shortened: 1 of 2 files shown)


app/components/views/Offset/index.tsx
```diff
@@ -120,12 +120,13 @@ export const Offset = (props: Props) => {
...
+  const [beneficiaryAddressError, setBeneficiaryAddressError] = useState(false);
   const [retirementMessage, setRetirementMessage] = useState("");
   // for selective retirement
-  const [specificAddresses, setSpecificAddresses] = useState([""]);
-  const validSpecificAddresses = specificAddresses.filter(
-    (str) => str.length === 42 && str.startsWith("0x")
-  );
+  const [specificAddresses, setSpecificAddresses] = useState({
+    addresses: [""],
+    errors: [false],
```
Atmosfearful, line 128: No need to persist these errors in state. We can just derive the error state at render time (components already rerender with every `onChange` event)

Cleanest would be to create a helper that returns a boolean and invoke it like this:
```
// at the input component
<input
  data-error={isValidAddress(specificAddresses[i]) })} // conditional className is also an option
 />
 
// styles.ts
export const input = css`
  &[data-error="true"]:not(:focused) {
   border: 0.2rem solid red;
  }
`;
```

I'm kind of a fan of `data` attributes because they are concise but you can also do a conditional classname using the `cx` util from emotion

Atmosfearful (changes requested):

app/components/views/Offset/index.tsx
```diff
@@ -230,6 +231,16 @@ export const Offset = (props: Props) => {
...
   };
 
+  const handleBeneficiaryAddressChange = (address: string) => {
+    if (ethers.utils.isAddress(address)) {
+      setBeneficiaryAddressError(false);
+      setBeneficiaryAddress(address);
+    } else {
+      setBeneficiaryAddressError(true);
+      setBeneficiaryAddress(address);
+    }
```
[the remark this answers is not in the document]
goatsOnDrugs: fixed with ethers.utils.isAddress(address) || address === ""

app/components/views/Offset/index.tsx
goatsOnDrugs: updated to use data attr

Atmosfearful (approved):
works great!

PR #451, opened 2022-05-26
Makes the language swither redirect to the same page in the other language

### LINKED ISSUE #420: NextJS: Switching language should load same page
## Problem

When switching the language on a subpage NextJS redirect to the Home page with the locale attached.
Instead, the user should stay on the same page, only the language should change

**Example:**
- Open `https://www.klimadao.finance/fr/buy`, 
- Switch the language to german 
- => loads the Home page `https://www.klimadao.finance/de` in german 
- Instead should load `https://www.klimadao.finance/de/buy`

biwano:
## Description

Makes the language swither redirect to the same page in the other language

## Related Ticket

Resolves #420

## Changes

NA

## Checklist

### DIFF
--- a/site/components/ChangeLanguageButton/index.tsx
+++ b/site/components/ChangeLanguageButton/index.tsx
@@ -14,7 +14,7 @@ import { useRouter } from "next/router";
  */
 export const ChangeLanguageButton: FC = () => {
   const [showMenu, setShowMenu] = useState(false);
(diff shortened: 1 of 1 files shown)


ladytrekker (approved):
🎉  Thanks

Atmosfearful (approved):
e.z. thanks!

PR #468, opened 2022-06-04
Inverse bonds (KLIMA -> USDC)

Atmosfearful:
Ready to merge when @ArchimedesCrypto deploys the market.
I have it hardcoded to market index `4` if this changes you can edit it here:

https://github.com/KlimaDAO/klimadao/pull/468/files#diff-05bc13fbc6c4f91260672ab863c12c9adcdc7c95c2aa1c6c2daaeb7dcdf0834eR179


app/actions/bonds.ts
```diff
@@ -306,36 +419,88 @@ export const calculateUserBondDetails = (params: {
...
+        (1 / Number(formatUnits(bondPrice, 6))) * Number(params.value) -
+        (1 / Number(formatUnits(bondPrice, 6))) *
+          Number(params.value) *
+          acceptedSlippage;
+      const formattedMinAmountOut = ethers.utils.parseUnits(
+        Number(minAmountOut.toFixed(6)).toString(),
+        "mwei"
+      );
+      // TODO add "remaining cpacity" line in view
+      // TODO check "capacity" in markets call to see if there is enough usdc in bond. returns usdc
```
Atmosfearful, line 450:
```suggestion
```

app/actions/bonds.ts
```diff
@@ -306,36 +419,88 @@ export const calculateUserBondDetails = (params: {
...
+      params.onStatus("userConfirmation", "");
+      const notFuckedAddress = await signer.getAddress();
+      const formattedValue = ethers.utils.parseUnits(params.value, "gwei");
+      // contract.deposit(__id, [amountIn (inKLIMA), min Amount Out (inUSDC)], [userAddress, DAOMSigAddress(0x65A5076C0BA74e5f3e069995dc3DAB9D197d995c)])
```
Atmosfearful, line 454:
```suggestion
```

app/actions/bonds.ts
```diff
@@ -306,36 +419,88 @@ export const calculateUserBondDetails = (params: {
...
+          Number(params.value) *
+          acceptedSlippage;
+      const formattedMinAmountOut = ethers.utils.parseUnits(
+        Number(minAmountOut.toFixed(6)).toString(),
+        "mwei"
+      );
+      // TODO add "remaining cpacity" line in view
+      // TODO check "capacity" in markets call to see if there is enough usdc in bond. returns usdc
+      params.onStatus("userConfirmation", "");
+      const notFuckedAddress = await signer.getAddress();
```
Atmosfearful, line 452: what was wrong with params.address? plz leave a comment if we need to know
otherwise rename this var `address`

app/components/views/Bond/index.tsx
```diff
@@ -593,44 +752,48 @@ export const Bond: FC<Props> = (props) => {
                   Number(bondState?.bondQuote) > Number(bondState?.maxBondPrice)
                 }
                 value={trimWithPlaceholder(bondState?.maxBondPrice, 2, locale)}
```
Atmosfearful, line 754: As mentioned in discord, this needs to fallback to capacity, if the capacity is lower than the maxbondprice.

Atmosfearful (commented):
good to go, feel free to address these now or in a follow up pr after we go live

Atmosfearful (commented):
NVM we need to fix the approval bug I mentioned in discord @goatsOnDrugs

ShimonD-KlimaDAO:
Should change the following strings:

1. "Percentage discount (or premium if negative) on KLIMA received for this bond, relative to the market value of KLIMA."
To (potentially):  "Percentage discount (or premium if negative) on USDC received for this bond, relative to the market value of USDC."

2. "Amount of bonded KLIMA you will get, at the provided input quantity"
To (potentially):  "Amount of USDC you will get, at the provided input quantity"

3. "Maximum amount of KLIMA you can acquire by bonding"
To (potentially): "Maximum amount of USDC you can acquire by bonding KLIMA"

-----------------------

The other bonds show "Balance" in orange color when it's attempted to be exceeded, but the inverse bonds do not.

Good:
image

Not good:
image


--------------------------

I think we should keep "Vesting Term" in the inverse bond, and just have it say "INSTANT". Most people might think that it's the same as other bonds (5 days), and this may prevent sell pressure from being absorbed as much.



-------------------------

BOND.MAX_EXCEEDED? 0.o

image

app/actions/bonds.ts
```diff
@@ -306,36 +419,87 @@ export const calculateUserBondDetails = (params: {
...
+      const contract = new ethers.Contract(
+        contractAddress,
+        Depository.abi,
+        signer
+      );
+      const calculatePremium = await contract.bondPrice();
+      const acceptedSlippage = params.slippage / 100 || 0.02; // 2%
+      const maxPremium = Math.round(calculatePremium * (1 + acceptedSlippage));
+      const valueInWei = ethers.utils.parseUnits(params.value, "ether");
+      const notFuckedAddress = await signer.getAddress();
```
ShimonD-KlimaDAO, line 485: still need to rename this in 4 places :)

app/actions/bonds.ts
goatsOnDrugs: oh, crap sorry I forgot to rename that. I have no idea what was wrong with the param. javascript was assigning it a random number instead of the address at runtime. I added a comment and change var name.

Issue #325, opened 2022-04-20
Migrate from iframe to Mobilum widget


Atmosfearful:
This will allow us to apply custom CSS, pre-fill fields like the user address, and hopefully respond to state changes too. DM me for the integration guide that the mobilum team gave me.

Issue #241, opened 2022-03-07
Integrate Transak SDK


Atmosfearful:
Looks like Transak is now supporting klima polygon purchases (GBP and other currencies it seems, but not USD yet) so we can get moving on this. I am awaiting information on an API key

https://global.transak.com/

* [ ] get clear designs (Add a /buy view to the dApp?)
* [ ] use constructed URL/widget https://transak.gitbook.io/transak-docs/quick-guides/setting-up-a-quick-demo-integration

PR #470, opened 2022-06-04
WIP: Transak and Mobilum on /buy page

### LINKED ISSUE #344: [EPIC] App /buy view refresh
See new designs in Figma.
https://www.figma.com/file/t2suR7Rz8f7lZF3KsCL7gu/Buy-Klima?node-id=0%3A1
* [ ] #325 
* [ ] #241

kaxline:
## Description

This PR will track adding Transak and Mobilum to the `/buy` page of the `app`. This involves creating a new modal and then pulling in the appropriate widget depending on which buy option the user selects. Previously we have been using the iframe version of the Mobilum integration, but as part of this task, we'll be moving to the widget version.

Currently this is a WIP and I'm creating the PR just to collect feedback on the direction and center the discussion in one place.

## Related Ticket

#241 #325 #344 

Resolves #344

## Checklist


ShimonD-KlimaDAO:
1. When the browser window is smaller than the widget, I can't scroll up or down to see the entire widget. The only way to see it is by expanding the window.
image

2. The default currency for Transak is USD, and yet USD is not supported. The default should be something else, like EUR. (see screenshot in 1)

3. "Swap on" should only have SushiSwap, not Uniswap nor Quickswap since there's no liquidity there.
image

4. Mobilum widget is cut on mobile (Brave on Android):
Screenshot_20220825-100227_Brave
Screenshot_20220825-100250_Brave

5. "Select" and the back arrow overlap on mobile:
Screenshot_20220825-100126_Brave

6. "You receive" and the KLIMA symbol overlap. Also, I think we should use a version of the symbol that looks nicer with this layout since it looks a bit unfitting.
Screenshot_20220825-100010_Brave

7. Should add Transak to this description:
image

kaxline:
@ShimonD-KlimaDAO Thanks for finding all those issues! I think I fixed them if you have time to check again.

app/pages/api/buy.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  const mobilumUrl = new URL("https://api.mobilum.com/widget/GetWidget");
+  mobilumUrl.searchParams.set("walletAddress", reqBody.address);
+
+  const body = {};
+
+  const widgetResponse = await fetch(mobilumUrl.href, {
+    method: "POST",
+    headers: {
+      ApiKey: process.env.MOBILUM_API_KEY as string,
```
Atmosfearful, line 29: I liked your suggestion to create this endpoint and protect this key. But I was poking around the response payload for this service and the key is right there 🙄 so I guess we can move all this logic to the client.
image
I guess these keys are not sensitive. But we will still keep them in `vercel` env vars as a best practice

app/pages/api/buy.ts
```diff
@@ -0,0 +1,51 @@
...
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify(body),
+  });
+
+  if (!widgetResponse) {
+    return res.status(500).json({ error: "Internal server error" });
```
Atmosfearful, line 36: if there is a network error, like Mobilum is down or our API key expired, `fetch` throws so it won't get to this point.
but if they give us a 500 or 404 or something instead, fetch does not throw

so we need to check `res.ok` and handle those cases

I also recommend using a "type guard" with `is` syntax to verify and apply the correct type to the response body

```ts
const isValidMobilumResponse = (body: unknown): body is MobilumResponse => {
  if (body?.result?.widgetBase64Html && body?.result?.widgetBase64ScriptUrl) return true;
  return false
}
```

app/components/BuyModal/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+        _getWidget();
+      }
+    }
+  }, [buyModalService]);
+
+  const resizeHandler = useCallback(() => {
+    if (window.innerHeight < 800) {
+      console.log(window.innerHeight);
+      setModalHeight(window.innerHeight - 200);
```
Atmosfearful, line 57: this is just viewport height, we can do it with plain css!
`calc(100vh - 20rem)`

This resize listener is buggy on windows, it doesn't always fire so I was able to create bugs where the modal stays smooshed. 
image

Take another stab, I'm confident we can scrap any javascript styling and accomplish this layout with pure css

Atmosfearful (changes requested):
Close button for transak is not visible on iphone SE
image

app/pages/api/buy.ts
kaxline: @Atmosfearful OK, I was able to find this finally. Not sure why I couldn't see it before. I have a question into Mobilum about it just so we have an answer.

Would you prefer I just remove the endpoint or should I strip out the `data-apikey` property and continue to send it through.

app/components/BuyModal/index.tsx
kaxline: Great! My CSS knowledge is stuck in 2008 or earlier ...

kaxline:
@Atmosfearful OK, back to you for review. I took out the API endpoint and, I hope, fixed the other stuff. Be sure to change the env variable in vercel from `MOBILUM_API_KEY` to `NEXT_PUBLIC_MOBILUM_API_KEY`

ShimonD-KlimaDAO:
https://github.com/KlimaDAO/klimadao/pull/470#issuecomment-1227545965

Issue 1 is still not fixed for Mobilum.
Issue 3 - the DEX options have been entirely removed, but I can still see "swap on" which should also be removed.
image

@Atmosfearful @goatsOnDrugs 

Moving back to In Progress column

Issue #364, opened 2022-04-29
[retirement pages] Resolve KNS and ENS domains in the url path


Atmosfearful:
After we finish the v1 implementation #345 
KNS (.klima) and ENS (.eth) domains are just unique names that resolve to public ethereum addresses.
```ladytrekker.eth -> 0x1234567890```

Let's enhance to resolve KNS (.klima) and ENS (.eth) domains so people can have cute URLs like `/retirements/ladytrekker.eth/1` or `/retirements/ladytrekker.klima/1`.

## KNS
You can reserve a kns domain for yourself here https://www.kns.earth/#/
From `TempeTechie (punk.domains)#1973`

> basically, you just need to call the .klima TLD contract (0xe8b97542A433e7eCc7bB791872af04DF02A1a6E4) and the function getDomainHolder which accepts domain name (just name, no extension) as a parameter
> pseudocode example:
> klimaContract.getDomainHolder("timmy");
> this returns the address that holds timmy.klima domain

## ENS
Reserve one for yourself here https://ens.domains/

Integration steps to be determined, probably a similar approach.

# Considerations
- [ ] should we enhance the design to show the .eth and .kns name on every retirement page? Or are we good with just showing the public address?
- If  `ladytrekker.eth` resolves to `0x12345`, and `ladytrekker.klima` also resolves to `0x12345`, that means there will be three duplicate pages:
```
/retirements/ladytrekker.eth/1
/retirements/ladytrekker.klima/1
/retirements/0x123456789123456789123456789123456789012/1   <- canonical
```
This is fine! We just need to tell search engines which one is canonical using a meta tag https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls
The public address `0x123...` is always the canonical.

- domains aren't permanent, they can be remapped. With `revalidate` in `getStaticProps`, we should be fine. If the user changes their domain to a different public address, everything should work fine once the page revalidates and regenerates itself. The canonical meta tag will just point to the new address's page.

PR #471, opened 2022-06-06
Retirements: Resolve KNS and ENS domains

ladytrekker:
## Description

This PR does:
- resolves KNS or ENS domains in the URL to the underlying address
- throw if no address could be resolved for the KNS or ENS domain
- use the resolved address for subgraph queries and retirementStorage contract methods
- show the domain instead of the beneficiary address on the pages
- ensure that the canonical URL is always the address (=> see [ticket](https://github.com/KlimaDAO/klimadao/issues/364) for explanation)

## Related Ticket

Resolves https://github.com/KlimaDAO/klimadao/issues/364

## Checklist

### DIFF
--- a/lib/utils/ens/index.ts
+++ b/lib/utils/ens/index.ts
@@ -1,0 +1,16 @@
+import { getDefaultProvider } from "@ethersproject/providers";
+
+export const isENSDomain = (domain: string) =>
+  !!domain && domain.includes(".eth");
+
+export const ETHProvider = getDefaultProvider();
+
+export const getAddressByENS = async (domain: string) => {
+  try {
+    const address = await ETHProvider.resolveName(domain);
+    return address;
+  } catch (e) {
+    console.log("Error in getAddressByENS", e);
+    return Promise.reject(e);
+  }
+};
--- a/lib/utils/kns/index.ts
+++ b/lib/utils/kns/index.ts
@@ -1,0 +1,36 @@
+import { ethers } from "ethers";
+import { addresses } from "../../constants";
+import PunkTLD from "../../abi/PunkTLD.json";
+import { getJsonRpcProvider } from "../getJsonRpcProvider";
+
+// https://www.kns.earth/#/
+export const isKNSDomain = (domain: string): boolean =>
+  !!domain && domain.toLowerCase().includes(".klima");
+
+// Use this import and overwrite your provider if needed with
+// import { KNSContract } from '...'
+// KNSContract.provider = myProvider;
+export const KNSContract = new ethers.Contract(
+  addresses["mainnet"].klimaNameService,
+  PunkTLD.abi,
+  getJsonRpcProvider()
+);
+
+// contract method getDomainHolder returns this address on not found
+const EMPTY_ADDRESS = "0x0000000000000000000000000000000000000000";
+
+const isValidAddress = (address: string) =>
+  !!address && address !== EMPTY_ADDRESS;
+
+export const getAddressByKNS = async (
+  domain: string
+): Promise<string | null> => {
+  try {
+    const strippedDomain = domain.replace(".klima", "");
+    const address = await KNSContract.getDomainHolder(strippedDomain);
+    return isValidAddress(address) ? address : null;
+  } catch (e) {
+    console.log("Error in getAddressByKNS", e);
+    return Promise.reject(e);
+  }
+};
--- a/site/lib/getAddressByDomain.ts
+++ b/site/lib/getAddressByDomain.ts
@@ -1,0 +1,25 @@
+import {
+  isKNSDomain,
+  getAddressByKNS,
+  isENSDomain,
+  getAddressByENS,
+} from "@klimadao/lib/utils";
+
+export const getAddressByDomain = async (
+  domain: string
+): Promise<string | null> => {
+  try {
+    let address;
+    if (isKNSDomain(domain)) {
+      address = await getAddressByKNS(domain);
+    } else if (isENSDomain(domain)) {
+      address = await getAddressByENS(domain);
+    } else {
+      address = null;
+    }
+    return address;
+  } catch (e) {
+    console.log("Error in getAddressByDomain", e);
+    return Promise.reject(e);
+  }
+};
--- a/site/pages/retirements/[beneficiary_address]/[retirement_index].tsx
+++ b/site/pages/retirements/[beneficiary_address]/[retirement_index].tsx
@@ -28,6 +28,7 @@ interface PageProps {
   retirementIndexInfo: RetirementIndexInfoResult;
   projectDetails: VerraProjectDetails | null;
   domain?: string;
+  canonicalUrl?: string;
 }
 
 // second param should always be a number
@@ -108,6 +109,9 @@ export const getStaticProps: GetStaticProps<PageProps, Params> = async (
         translation,
         projectDetails,
         domain: !!addressByDomain ? params.beneficiary_address : undefined,
+        canonicalUrl: !!addressByDomain
+          ? `https://www.klimadao.finance/${addressByDomain}/${params.retirement_index}`
+          : undefined,
       },
       revalidate: 240,
     };
--- a/site/pages/retirements/[beneficiary_address]/[retirement_index].tsx
+++ b/site/pages/retirements/[beneficiary_address]/[retirement_index].tsx
@@ -13,6 +13,8 @@ import { VerraProjectDetails } from "@klimadao/lib/types/verra";
 import { SingleRetirementPage } from "components/pages/Retirements/SingleRetirement";
 import { loadTranslation } from "lib/i18n";
 import { INFURA_ID } from "lib/constants";
+import { getIsDomainInURL } from "lib/getIsDomainInURL";
+import { getAddressByDomain } from "lib/getAddressByDomain";
 
 interface Params extends ParsedUrlQuery {
   beneficiary_address: string;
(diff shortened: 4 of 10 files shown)


lib/utils/ens/index.ts
```diff
@@ -0,0 +1,16 @@
...
+export const ETHProvider = getDefaultProvider();
+
+export const getAddressByENS = async (domain: string) => {
+  try {
+    const address = await ETHProvider.resolveName(domain);
+    return address;
+  } catch (e) {
+    console.log("Error in getAddressByENS", e);
+    return Promise.reject(e);
+  }
```
Atmosfearful, line 15:
```suggestion
    console.error("Error in getAddressByENS", e);
    return Promise.reject(e);
  }
```

Atmosfearful: I usually do `throw e` instead of `return Promise.reject(e)` what's the difference? an outer `await` in another try/catch will still catch right? I guess both achieve the same thing?

lib/utils/kns/index.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const getAddressByKNS = async (
+  domain: string
+): Promise<string | null> => {
+  try {
+    const strippedDomain = domain.replace(".klima", "");
+    const address = await KNSContract.getDomainHolder(strippedDomain);
+    return isValidAddress(address) ? address : null;
+  } catch (e) {
+    console.log("Error in getAddressByKNS", e);
```
Atmosfearful, line 33:
```suggestion
    console.error("Error in getAddressByKNS", e);
```

lib/utils/ens/index.ts
```diff
@@ -0,0 +1,16 @@
...
+  !!domain && domain.includes(".eth");
+
+export const ETHProvider = getDefaultProvider();
+
+export const getAddressByENS = async (domain: string) => {
+  try {
+    const address = await ETHProvider.resolveName(domain);
+    return address;
+  } catch (e) {
+    console.log("Error in getAddressByENS", e);
```
Atmosfearful, line 13:
```suggestion
    console.error("Error in getAddressByENS", e);
```

site/lib/getAddressByDomain.ts
```diff
@@ -0,0 +1,25 @@
...
+    if (isKNSDomain(domain)) {
+      address = await getAddressByKNS(domain);
+    } else if (isENSDomain(domain)) {
+      address = await getAddressByENS(domain);
+    } else {
+      address = null;
+    }
+    return address;
+  } catch (e) {
+    console.log("Error in getAddressByDomain", e);
```
Atmosfearful, line 22:
```suggestion
    console.error("Error in getAddressByDomain", e);
```

site/pages/retirements/[beneficiary_address]/index.tsx
```diff
@@ -31,12 +37,23 @@ export const getStaticProps: GetStaticProps<PageProps, Params> = async (
...
     }
 
+    let addressByDomain;
+    const isDomainInURL = getIsDomainInURL(params.beneficiary_address);
+    if (isDomainInURL) {
+      addressByDomain = await getAddressByDomain(params.beneficiary_address);
+    }
+
+    // Do not continue if KNS or ENS Domain can not be resolved
+    if (addressByDomain === null) {
```
Comment (Atmosfearful) [inline site/pages/retirements/[beneficiary_address]/index.tsx:47]: makes me uncomfortable that `undefined` is OK (not a domain) but `null` is an error.

What about something like this up on L40?
```ts
let resolvedAddress: string;
if (isDomainInURL(params.beneficiary_address) {
  resolvedAddress = await getAddressByDomain(params.beneficiary_address); // this fn should throw if it fails to resolve
} else if (ethers.utils.isAddress(address)) {
  resolvedAddress = params.beneficiary_address;
} else {
  throw new Error("Not a valid beneficiary address");
}
```

lib/utils/kns/index.ts
```diff
@@ -0,0 +1,36 @@
...
+const isValidAddress = (address: string) =>
+  !!address && address !== EMPTY_ADDRESS;
+
+export const getAddressByKNS = async (
+  domain: string
+): Promise<string | null> => {
+  try {
+    const strippedDomain = domain.replace(".klima", "");
+    const address = await KNSContract.getDomainHolder(strippedDomain);
+    return isValidAddress(address) ? address : null;
```
Atmosfearful, line 31: I would just throw here, and let the error bubble up on it's own, rather than trying to look for `null`
And so the return type is always `Promise<string>`
See my suggestion below about null vs undefined

site/components/pages/Retirements/List/index.tsx
```diff
@@ -11,10 +11,11 @@ import * as styles from "./styles";
 
 type Props = {
   klimaRetires: KlimaRetire[];
+  domain?: string;
 };
 
 export const AllRetirements: FC<Props> = (props) => {
-  const { klimaRetires } = props;
+  const { klimaRetires, domain } = props;
```
Atmosfearful, line 18:
```suggestion
  const { klimaRetires, nameserviceDomain } = props;
```

just to be very clear so noone gets confused and thinks this is the actual web domain :)

site/pages/retirements/[beneficiary_address]/index.tsx
```diff
@@ -52,6 +69,10 @@ export const getStaticProps: GetStaticProps<PageProps, Params> = async (
         totalsAndBalances,
         klimaRetires,
         beneficiaryAddress: params.beneficiary_address,
+        domain: !!addressByDomain ? params.beneficiary_address : undefined,
+        canonicalUrl: !!addressByDomain
+          ? `https://www.klimadao.finance/${addressByDomain}`
```
Comment (Atmosfearful) [inline site/pages/retirements/[beneficiary_address]/index.tsx:74]:
```suggestion
          ? `https://www.klimadao.finance/retirements/${addressByDomain}`
```

site/pages/retirements/[beneficiary_address]/[retirement_index].tsx
```diff
@@ -92,6 +108,10 @@ export const getStaticProps: GetStaticProps<PageProps, Params> = async (
         retirementTotals: params.retirement_index,
         translation,
         projectDetails,
+        domain: !!addressByDomain ? params.beneficiary_address : undefined,
+        canonicalUrl: !!addressByDomain
+          ? `https://www.klimadao.finance/${addressByDomain}/${params.retirement_index}`
```
Comment (Atmosfearful) [inline site/pages/retirements/[beneficiary_address]/[retirement_index].tsx:113]:
```suggestion
          ? `https://www.klimadao.finance/retirements/${addressByDomain}/${params.retirement_index}`
```

Atmosfearful (changes requested):

site/lib/getAddressByDomain.ts
ladytrekker: Done

lib/utils/ens/index.ts
ladytrekker: Done

lib/utils/kns/index.ts
ladytrekker: Done

site/components/pages/Retirements/List/index.tsx
ladytrekker: Done

site/pages/retirements/[beneficiary_address]/index.tsx
Comment (ladytrekker) [reply site/pages/retirements/[beneficiary_address]/index.tsx:74]: Ahh good catch, done

site/pages/retirements/[beneficiary_address]/[retirement_index].tsx
Comment (ladytrekker) [reply site/pages/retirements/[beneficiary_address]/[retirement_index].tsx:113]: Done

lib/utils/ens/index.ts
ladytrekker: Yes, both achieve the same thing in the end. But as we are awaiting a promise which could fail and not in a plain try/catch without promises, I decided to go for Promise.reject

site/pages/retirements/[beneficiary_address]/index.tsx
Comment (ladytrekker) [reply site/pages/retirements/[beneficiary_address]/index.tsx:47]: Thanks for the hint with the ether helper. Used in multiple places now. Done

lib/utils/kns/index.ts
ladytrekker: Is changed. Done

lib/utils/ens/index.ts
```diff
@@ -0,0 +1,16 @@
+import { getDefaultProvider } from "@ethersproject/providers";
+
+export const isENSDomain = (domain: string) =>
+  !!domain && domain.includes(".eth");
+
+export const ETHProvider = getDefaultProvider();
```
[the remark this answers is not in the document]
ladytrekker: Ah interesting stuff. Did not know this.

I refactored all places and pass now the Infura URL to the provider instead of the ID.
This way we can pass different Urls around.

There are no further occurrences of `getDefaultProvider` on the server.
Only in the App on the client.

Please check again

Atmosfearful (approved):

ladytrekker:
> When I input my KNS domain, it shows my KNS domain on the page. When I input the underlying address, it shows the same data, but my address is not resolved to my KNS domain. Compare the two strings under "for beneficiary" below. Same exact issue for ENS.
> 
> https://klimadao-site-git-kns-ens-retirement-klimadao.vercel.app/retirements/shimond.klima https://klimadao-site-git-kns-ens-retirement-klimadao.vercel.app/retirements/0x6F9F81eb4f54512Be8c91833783ee0074328E062

Yes, this is the intended behavior.
The data in the Url is the beneficiary name.
Both are valid and result in the same page.
Please read the ticket content again.

If you think it should always resolve in the KNS or ENS domain then this should be discussed in a broader channel.

Issue #448, opened 2022-05-25
Fix twitter cards for blog posts


Atmosfearful:
Perhaps we aren't passing the correct image URLs down...

biwano:
Hi @Atmosfearful !

Everything seems fine on the site side.
for instance, in page `https://www.klimadao.finance/fr/blog/response-to-verra-comments-on-crypto-instruments-and-tokens`
The meta for twitter:image is `<meta name="twitter:image" content="https://cdn.sanity.io/images/dk34t4vc/production/79d1b9b219d6feaa6eefa89b2e0a7c43d9ee4d9c-4312x2160.png"/>`
The image at address `https://cdn.sanity.io/images/dk34t4vc/production/79d1b9b219d6feaa6eefa89b2e0a7c43d9ee4d9c-4312x2160.png` seems okay.

Can you send me the link to the twitter post? Maybe there is a particular problem with (a) particular post(s).

biwano:
I think I found it after re-reading the docs: https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image

There are specifications for image size, notably: `maximum of 4096x4096 pixels. Images must be less than 5MB in size.`
The image in the post I checked is 4312x2160 and a wooping 7MB.

The problem must come from that.

biwano:
@Atmosfearful Do you think we can try to resize the image in one of the problematic posts to see if it really comes from that?

Atmosfearful:
@biwano I DM'd you on discord about this, I'll give you CMS access.

biwano:
@Atmosfearful Progress so far: I put the same image on this blog `https://staging-site.klimadao.finance/blog/klima-dao-fair-launch-pt-1-initial-discord-offering` that is on the home page. And it still does not work. Here are my thoughts:
- There may be a cache issue on the twitter side (which I doubt).
- Something really is wrong in the meta tags.
- Something is wrong with the way the CDN delivers the images.

I keep investigating

biwano:
@Atmosfearful 

# Root cause analysis

There was indeed a cache on the tweeter side. I had to change the blog slugs to progress.
You can check those tweets:
- Image respects twitter specs: https://twitter.com/biwa73847927/status/1536067000665001984?s=20&t=PYzzVDgpCK8qPckeMy98Mg => Tweet is beautiful
[...]

PR #472, opened 2022-06-07
Refactored CMS queries

biwano:
## Description

This change 
- refactors the `site/lib/queries.ts` file. It removes a bit of code duplication.
- Removes the use of the CDN on staging.

There is a collateral impact:
- Currently on the staging environnement, the hidden posts are not shown in the list but we can see them by entering the URL directly.
- This change shows the hidden posts in the blog list on the staging environnement.

## Related Ticket

Initialy I made these changes on my environnement so I could work on #448
I can keep the refactoring and remove the collateral impacts depending on what is really expected.

## Changes

NA

## Checklist

### DIFF
--- a/site/lib/queries.ts
+++ b/site/lib/queries.ts
@@ -6,10 +6,14 @@ wants to publish their posts but dont want it to appear in the production enviro
 NOTE: When constructing a new query for posts, please add the hideFromProduction filter to the prodQuery
 so the query wont reveal any hidden posts in the production environment */
 
-const prodQueries = {
+export const queryFilter = IS_PRODUCTION
+  ? "hideFromProduction != true"
+  : "true";
+
+export const queries = {
   /** fetch all blog posts, sorted by publishedAt */
   allPosts: /* groq */ `
-    *[_type == "post" && hideFromProduction != true] | order(publishedAt desc) {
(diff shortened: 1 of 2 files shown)


site/lib/queries.ts
```diff
@@ -6,60 +6,14 @@ wants to publish their posts but dont want it to appear in the production enviro
...
+export const queryFilter = IS_PRODUCTION
+  ? "hideFromProduction != true"
+  : "true";
 
-const stagingQueries: typeof prodQueries = {
+export const queries = {
   /** fetch all blog posts, sorted by publishedAt */
   allPosts: /* groq */ `
-    *[_type == "post" && hideFromProduction != true] | order(publishedAt desc) {
+    *[_type == "post" && ${queryFilter}] | order(publishedAt desc) {
```
Atmosfearful, line 16: We don't want to show articles on the main list on staging because I don't want important articles to get leaked and shared to the public while authors are still working on them. They should only be accessible on staging by those who know the exact URL

```suggestion
    *[_type == "post" && hideFromProduction != true] | order(publishedAt desc) {
```

Atmosfearful (changes requested):

Atmosfearful (approved):

Issue #474, opened 2022-06-07
Update lightmode/darkmode switch logic to support being disableed for a  specific page


Atmosfearful:
## Problem

Hiding the ThemeToggle is easy.
But the page should also render a fixed theme in the body tag `theme-dark` or `theme-light`.
Otherwise the `_document` would initilaze the theme by default with `theme-light`

**Ideas:**
- from specific page `getStaticProps` pass optional pageProp `fixedTheme` with the value `theme-dark` or `theme-light` to `_document`
- from `_document` get  `fixedTheme` from `this.props.__NEXT_DATA__.props.pageProps`
- in `_document` if `fixedTheme` is defined => do not render `<InitializeTheme />`
- in `_document` if `fixedTheme` is defined => add body data for this theme
- if `fixedTheme` is defined => do not render the `ThemeToggle` component

PR #476, opened 2022-06-09
Set optional fixed theme and hide ThemeToggle

ladytrekker:
## Description

Some pages require a fixed theme and the ThemeToggle should not be shown either.

## Test

1. Open the [preview site](https://klimadao-site-git-optional-theme-toggle-klimadao.vercel.app/) and change the theme to "light"
2. Click on "Buy" which has a fixed theme "dark"  => the site should be in dark and no ThemeToggle is shown 
3. Click on "Resources" or "Home" => theme should be "light" again (taken from the previous preferences in Step 1)
4. Change the theme to "dark"
5. Open the retirement page with [this link](https://klimadao-site-git-optional-theme-toggle-klimadao.vercel.app/retirements/0x087a7afb6975a2837453be685eb6272576c0bc06) which has a fixed theme "light" => theme should be light now and no ThemeToggle is shown 
6. Click on "Resources" or "Home" => theme should be "dark" again (taken from the previous preferences in Step 4)

## CAUTION

Remove test code from pages "buy" and "retirements" before merge !

## Related Ticket

Resolves https://github.com/KlimaDAO/klimadao/issues/474

## Checklist

### DIFF
--- a/app/pages/_document.tsx
+++ b/app/pages/_document.tsx
@@ -3,7 +3,6 @@ import { WebFonts } from "@klimadao/lib/components";
 import Document, { Html, Head, Main, NextScript } from "next/document";
 import createEmotionServer from "@emotion/server/create-instance";
 import { cache } from "@emotion/css";
-import { THEME_DARK } from "@klimadao/lib/theme/constants";
 
 class MyDocument extends Document {
   render() {
@@ -12,7 +11,7 @@ class MyDocument extends Document {
         <Head>
           <WebFonts />
         </Head>
-        <body data-theme={THEME_DARK}>
+        <body data-theme="theme-dark">
           <Main />
           <NextScript />
         </body>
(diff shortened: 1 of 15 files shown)


Atmosfearful (approved):
Nice, bit tricky. Thanks for the test examples!! The /buy page test does flash white at first which is unfortunate. Maybe you can take a second look at `InitializeTheme` and come up with a way to set the hardcoded theme there rather than in useEffect? We could pass `fixedtheme` boolean down to InitializeTheme via `document.body.dataset` maybe?

Just an idea i didnt think it through. Otherwise its good to go when you remove the text examples!

ladytrekker:
> Nice, bit tricky. Thanks for the test examples!! The /buy page test does flash white at first which is unfortunate. Maybe you can take a second look at `InitializeTheme` and come up with a way to set the hardcoded theme there rather than in useEffect? We could pass `fixedtheme` boolean down to InitializeTheme via `document.body.dataset` maybe?
> 
> Just an idea i didnt think it through. Otherwise its good to go when you remove the text examples!

_document is rendered only once on the initial creation of a page.
All the rest happens in _app
I added the `fixedThemeName` (it's a string, not a boolean) to the data of the body now in _document.
Feels smoother now after a lot of trial and error.

Test code is removed.
Merging to Staging now

PR #479, opened 2022-06-10
Fix 0x address on retirement index page

Atmosfearful:
## Description
Actually we were falling back to the full 0x address in in `pageprops.nameserviceDomain` which was a mistake. Now we correctly return `null` so the UI falls back to render `concatAddress()`

Plus a few other small cleanups:
- types, static props return null never `undefined` (only `null` is serializeable)
- differentiate `beneficiary` and `beneficiaryAddress`. The latter is always an 0x, the former might be a domain or an 0x.
- use nameservice in canonical urls

## Related Ticket
none

## Changes

| Before  | After  |
|---------|--------|
| image | image |

## Checklist

### DIFF
--- a/lib/utils/getRetirement/index.ts
+++ b/lib/utils/getRetirement/index.ts
@@ -27,7 +27,7 @@ export const createRetirementStorageContract = (
 };
 
 export const getRetirementIndexInfo = async (params: {
-  beneficiaryAdress: string;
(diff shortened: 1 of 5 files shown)


ladytrekker (approved):
Thanks! This mistake happened with all them if/else I guess. Thanks for the catch

PR #480, opened 2022-06-10
Translate Web3Modal & refactor app to web3 context provider

Atmosfearful:
# Big cascade of changes

Once upon a time, many moons ago, I set out to investigate the trouble @biwano was having in #413 

This led to a refactor of how we use web3modal (it needs to be re-instantiated when the locale changes).
Realizing @0xAeterno created a nice `useWeb3` hook, this needed to be adapted as well.
So instead of duplicating work, I moved `useWeb3` and associated libraries to `@klimadao/lib` and updated them to support locale changes.

During this time I realized that our `on("accountsChanged")` and `on("chainChanged")` events were not working with the new useWeb3 implementation, so I fixed those. This is probably a bug in ethers (see code comments)

@0xAeterno Because we need to reset redux state (user balances etc.) when the account changes or disconnects, I removed the useReducer pattern and went back to simply reloading the page.
# Try it
In the app preview, switch between EN and pseudo
image

# Tech Debts
- useWeb3 removes some props boilerplate in various places around the app, but I wanted to limit the amount of lines changed in this pr. We can go and clean those up now.
- If we do want to support clean `connect`/`disconnect`/`accountChanged`/`chainChanged` without reloading the whole app, we need a cleaner way to detect these events so the app can properly reset it's own redux state. Probably a callback pattern:
```ts
const { provider } = useWeb3({
  onChangeNetwork: handleChangeNetwork,
  onChangeAccount: handleChangeAccount
});
```


app/components/InvalidNetworkModal/index.tsx
```diff
@@ -6,7 +6,7 @@ import { providers } from "ethers";
 import { Trans } from "@lingui/macro";
 
 interface Props {
-  provider: providers.JsonRpcProvider;
+  provider?: providers.JsonRpcProvider;
```
biwano, line 9: Why do you allow for provider to be optional?

biwano (approved):
Very nice piece of code

0xAeterno (approved):
LGTM! Amazing refactor 🔥

app/components/InvalidNetworkModal/index.tsx
Atmosfearful: The old hook used to return a fallback jsonrpcprovider when a user is not connected, the new one returns undefined.

This is better, because we can instantiate a provider anywhere in the code, there is no need or the hook to return one.

Atmosfearful:
@ShimonD-KlimaDAO could use a quick sanity test on this as it's a critical path. It affects the connect, disconnect logic, both on `site` (the pledge) and on `app`.
Expected behavior is same as before-- "connect" should show the modal etc. Disconnect, change account, and change network should force a page refresh.

ShimonD-KlimaDAO:
> @ShimonD-KlimaDAO could use a quick sanity test on this as it's a critical path. It affects the connect, disconnect logic, both on `site` (the pledge) and on `app`. Expected behavior is same as before-- "connect" should show the modal etc. Disconnect, change account, and change network should force a page refresh.

Everything still works as it does in prod. However, I noticed that the Pledge behavior on prod (and here) is a bit strange.
I can switch networks and still edit my pledge. I never get prompted to switch to Polygon. I can also connect while being on Ethereum. This behavior is different than the main app which barks at me whenever I try to change networks or connect while on a different network.

Atmosfearful:
Good catch @ShimonD-KlimaDAO I created a ticket for that, just need to migrate over the Modal code (doesn't touch any of the changes here) https://github.com/KlimaDAO/klimadao/issues/507

PR #482, opened 2022-06-13
Request resized images from the CDN for og image meta tags

### LINKED ISSUE #448: Fix twitter cards for blog posts

biwano:
## Description

Sends an URL of a resized version of the CDN images for og image tags

I choose completly arbitrarily an height of 1000px 

## Related Ticket

Resolves #448

## Changes

When tweeting a post

| Before  | After  |
|---------|--------|
|image |image|

## Checklist

### DIFF
--- a/app/components/PageHead.tsx
+++ b/app/components/PageHead.tsx
@@ -1,7 +1,9 @@
 import Head from "next/head";
 import React from "react";
+
 import { useSelector } from "react-redux";
 import { selectLocale } from "state/selectors";
+import { getOgImageSrc } from "@klimadao/lib/utils";
 export interface PageHeadProps {
   production: boolean;
   /** <title> tag */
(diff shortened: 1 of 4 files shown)


app/components/PageHead.tsx
```diff
@@ -31,18 +35,16 @@ export const PageHead = (props: PageHeadProps) => {
       <meta name="description" content={props.metaDescription} />
       <meta property="og:description" content={props.metaDescription} />
       <meta property="og:title" content={props.mediaTitle} />
-      {props.mediaImageSrc && (
-        <meta property="og:image" content={props.mediaImageSrc} />
-      )}
+      <meta property="og:image" content={mediaImageSrc} />
+
```
Atmosfearful, line 39: do you think it causes problems if the page has no `mediaImageSrc`? like rendering an empty meta og:image tag, is that handled gracefully by most social media sites?

Atmosfearful (approved):

app/components/PageHead.tsx
biwano: Hi! 

I coded the getOgImageSrc function so that it returns `urls.mediaImage` if nothing is provided.
I though it was a nice addition, but maybe not. Let me know.

Issue #9, opened 2021-11-13
New approval and transaction modal experience


Atmosfearful:
We can solve bug reports related to insufficient approvals for a few users, and improve UX significantly.
Migrate from "max approvals" like 1000000000 to "as-needed" approvals based on the desired input quantity.

* [x] Get designs from Aaron. We agreed to break out into a multi-step modal that walks you through both the approve and submit process
https://www.figma.com/file/GtTkd4y6ksWyrDwg5g5YDd/%5B-Design-System-%5D
* [x] Replaces the existing txn status modals for all views
* [x] If the user has already approved that specific amount, auto-advance to the next step
  * [x] If the approval is larger, or smaller, than the desired quantity, re-do the approval txn. This way, everyone with an existing "max approval" will be migrated to the safer "as-needed" approval.
* [x] When approval is done, set the approval in state
* [x] When spend txn is complete, reset the state to "0"

# Stretch goals
- Optimization: only fetch the approval balances needed for a given view, when the view mounts. Don't fetch them all at app load anymore.

# Considerations
- The modal should support multiple approvals, like for pklima (spend bct + pklima)
- We may need to refactor state to handle the different approval pairs without so much copypasta. Every approval is a `token.(owner, spender)` pair. So perhaps this could be abstracted to something like this:
```ts
interface Approvals {
  [tokenName: string]: { // "bct" | "klima" | "sklima" etc.
    [spender: string]: string; // spender: value approved
  }
}
```
- Token decimals!!! Be careful. Properly format from `BigNumber` units before doing any comparisons. Properly format back to `BigNumber` with appropriate decimals before submitting the txn or manipulating state.
- Probably a good space for writing some tests (format/parse/compare)

## Implementation TODOS

- [x] refactor how allowances are stored in state
- [x] select allowances from new state shape
- [x] let user approve the amount which was entered (not 100000 gwei by default no more)
- [x] skip allowance only if the entered amount is the same as the allowance, otherwise always let the user approve first  => `if (allowance !== value) { showApprovalStep() }`
- [x] combine all approval transactions into one reusable action
- [x] use getTokenDecimal and formatUnit helpers everywhere
- [ ] load allowances when specific view was mounted
- [ ] implement all new changes for the BOND view too (getContract, getAllowances, getBalances etc...)

Dakavon:
> We need to catch this (rare) error where allowance is exceeded, and show the `allow` state request in the UI.

The dApp just throws an error without further explanation what to do.

[...]

PR #495, opened 2022-06-17
WIP: Refactor token allowances and approvals 🚧 

ladytrekker:
## Related Ticket

Resolves #<issue-number>

## Checklist


app/actions/user.ts
```diff
@@ -1,171 +1,161 @@
...
+
+const getBalance = async (params: {
+  token: Asset;
+  contract: ethers.Contract;
+  address: string;
+}): Promise<TokenValue> => {
+  try {
+    const balance = await params.contract.balanceOf(params.address);
+    return {
+      [params.token]: balance,
```
Atmosfearful, line 57: curious why you opt for an unformatted value in an object rather than just returning a formatted string?

Atmosfearful (changes requested):
im out of time for today but here are some thoughts so far

app/actions/pklima.ts
```diff
@@ -79,7 +77,7 @@ export const changeApprovalTransaction = async (params: {
     params.onStatus("networkConfirmation", "");
     await txn.wait(1);
     params.onStatus("done", "Approval was successful");
-    return value;
+    return formatUnits(value, 9);
```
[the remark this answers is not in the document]
ladytrekker: Sure that both should be 18?
Asking because of this comment here:
https://github.com/KlimaDAO/klimadao/blob/87514af8f41548dfda705868911fc18a590e6dab/app/actions/pklima.ts#L35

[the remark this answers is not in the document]
ladytrekker: I'd vote for extending this helper here with all possible tokens
https://github.com/KlimaDAO/klimadao/blob/staging/lib/utils/getTokenDecimals/index.ts

and use this everywhere

app/actions/pklima.ts
```diff
@@ -67,10 +68,7 @@ export const changeApprovalTransaction = async (params: {
         provider: params.provider.getSigner(),
       }),
     }[params.action];
-    const value = ethers.utils.parseUnits(
-      "1000000000000000000000000000",
-      "wei"
-    ); // BigNumber
+    const value = ethers.utils.parseUnits(params.value, "wei"); // BigNumber
```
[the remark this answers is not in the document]
ladytrekker: done

app/actions/stake.ts
```diff
@@ -5,11 +5,13 @@ import { addresses } from "@klimadao/lib/constants";
 import { formatUnits, getContract } from "@klimadao/lib/utils";
 
 export const changeApprovalTransaction = async (params: {
+  value: string;
   provider: providers.JsonRpcProvider;
   onStatus: OnStatusHandler;
   action: "stake" | "unstake";
 }): Promise<string> => {
   try {
+    const parsedValue = ethers.utils.parseUnits(params.value, "gwei");
```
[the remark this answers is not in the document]
ladytrekker: Done

app/actions/offset.ts
```diff
@@ -51,61 +54,70 @@ export const getRetirementAllowances = (params: {
...
   onStatus: OnStatusHandler;
 }): Promise<string> => {
   try {
     const contract = getContract({
       contractName: params.token,
       provider: params.provider.getSigner(),
     });
-    const decimals = getTokenDecimals(params.token);
-    const value = ethers.utils.parseUnits("1000000000", decimals);
+    const decimals = getTokenDecimals(params.token); // why here not "gwei" as in all other changeApprovalTransaction?
```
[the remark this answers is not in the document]
ladytrekker: I'd vote for using `getTokenDecimals` everywhere

app/actions/user.ts
ladytrekker: Is changed

lib/utils/getAllowance/index.ts
```diff
@@ -0,0 +1,44 @@
...
+} from "../../types/allowances";
+
+export const isSpenderInAddresses = (spender: string): boolean => {
+  const keys = Object.keys(
+    addresses.mainnet
+  ) as (keyof typeof addresses["mainnet"])[];
+  return keys.includes(spender as keyof typeof addresses["mainnet"]);
+};
+
+export const getAllowance = async (params: {
```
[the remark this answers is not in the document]
ladytrekker: Yeah, I came up with this pattern because two many spenders with the same tokens!
How do you know in the end which BigNumber belonged to which token from which spender?

You pass in spender: 
"wsklima" | "retirementAggregator" | "staking_helper" | "staking" | "pklima_exercise";
for tokens:
"bct" | "mco2" | "nct" | "ubo" | "nbo" | "klima" | "pklima" | "sklima" | "wsklima" | "usdc"

And after that you need to know which BigNumber belonged to which token from which spender.

Let me know if you see an easier solution.

app/actions/user.ts
```diff
@@ -1,171 +1,153 @@
...
+      // domains
+      const domains = [
+        getKns({
+          address: params.address,
+          contract: klimaDomainContract,
+        }),
+        getEns({ address: params.address }),
+      ];
+
+      const [knsDomain, ensDomain] = await Promise.all(domains);
```
Atmosfearful, line 85: nice speedup!

lib/utils/getAllowance/index.ts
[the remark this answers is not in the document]
Atmosfearful: So for that nested `forEach` here https://github.com/KlimaDAO/klimadao/pull/495/files#diff-254c9445c732e1b66b745e7db313501080a4ab658c580ed8584ed29502ecd194R111

we can track the index like `spender[spenderIndex]` and `token[tokenIndex]`

roughly what i have in mind:
```ts
// outer loop
spenders.reduce(state, value, spenderIndex => {

  // create array of strings
  const allowanceValues = tokenNames.map(getAllowance);

  // assign to inner object
  const spenderAllowances = allowanceValues.reduce((prev, value, i) => {
    const token = tokenNames[i]; // use index
    return { ...prev, [token]: value };
  });
  
  // assign the inner object to the final outer obj
  const spender = spenders[spenderIndex];
  return { ...prev, state[spender]: spenderAllowances }
});
```

Maybe its just my style but its more readable than a deep merge to me and preferable because it keeps the getAllowance util simple and predictable returning just a string!

app/actions/user.ts
```diff
@@ -1,171 +1,153 @@
...
+  "staking",
+  "wsklima",
+  "pklima_exercise",
+] as const;
+
+type TokenValueFormatted = {
+  [K in Asset]: string;
+};
+
+const getBalance = async (params: {
```
Atmosfearful, line 42: again i still think its better to do a formatted string rather than an object. the outer fn should always be able to use array index or the `token` string itself to know which value this is

if we return an object we might end up doing a lot of merge boiler plate if we reuse this in a lot of places for simple balance queries

app/actions/user.ts
```diff
@@ -1,171 +1,153 @@
...
+  token: Asset;
+  contract: ethers.Contract;
```
Atmosfearful, line 44: would be cool if this was smart enough to find the correct contract on it's own :)

future enhancement?

app/actions/user.ts
```diff
@@ -1,171 +1,153 @@
...
-        provider: params.provider,
-      });
-      const wsklimaContract = getContract({
-        contractName: "wsklima",
-        provider: params.provider,
-      });
-      const pKlimaContract = getContract({
-        contractName: "pklima",
+      const assetsContracts = createContractsObject({
+        contracts: assets as any,
```
Atmosfearful, line 67: This is risky, what is the type error saying?
it looks like the `assets` array is strongly typed, so perhaps there is something missing in `contracts` type?

app/actions/user.ts
```diff
@@ -1,171 +1,153 @@
...
       // balances
-      const balances = [
-        // CARBON
-        bctContract.balanceOf(params.address),
-        mco2Contract.balanceOf(params.address),
-        nctContract.balanceOf(params.address),
-        uboContract.balanceOf(params.address),
-        nboContract.balanceOf(params.address),
+      const promisesBalance = assets.reduce((arr, asset) => {
+        const contract = assetsContracts[asset];
```
Atmosfearful, line 89: nice, i like this!

app/actions/wrap.ts
```diff
@@ -12,7 +13,7 @@ export const changeApprovalTransaction = async (params: {
       contractName: "sklima",
       provider: params.provider.getSigner(),
     });
-    const value = ethers.utils.parseUnits("1000000000", "gwei"); //bignumber
+    const value = ethers.utils.parseUnits(params.value, 9); // bignumber, is 9 correct here?
```
Atmosfearful, line 16: I had to think this through, a bit confusing:
we only care about decimals of the approving contract, i.e. the token that will be spent. in this case thats sklima.approve() so, 9 is correct

app/components/views/PKlima/index.tsx
```diff
@@ -110,14 +135,16 @@ export const PKlima: FC<Props> = (props) => {
         onStatus: setStatus,
       });
       dispatch(redeemPklima(value));
+      // WHICH ALLOWANCE HAS TO BE DECREMENTED HERE ??
```
Atmosfearful, line 138: pklima exercise contract spending BCT

app/components/views/Stake/index.tsx
```diff
@@ -100,17 +115,33 @@ export const Stake = (props: Props) => {
...
+            spender: "staking_helper",
+            value: approvedValue,
+          })
+        );
       } else {
-        dispatch(setStakeAllowance({ sklima: value }));
+        dispatch(
+          incrementAllowance({
+            token: "sklima",
+            spender: "staking",
```
Atmosfearful, line 138: love this!

Atmosfearful: I guess as far as naming, now the actions should just be called `setAllowance` and `decrementAllowance`
we never actually add two values together (increment)

Atmosfearful (changes requested):

app/components/views/Stake/index.tsx
ladytrekker: Done

app/components/views/PKlima/index.tsx
ladytrekker: Done with 
```TS
 dispatch(
        decrementAllowance({
          token: "bct",
          spender: "pklima_exercise",
          value: approvedValue,
        })
      );
```

app/actions/wrap.ts
ladytrekker: Thanks, comment is removed

app/actions/user.ts
ladytrekker: Done

app/actions/user.ts
ladytrekker: Yeah, let´s discuss a smart way on how to init all contracts in one place in lib
so that both app and site can have access to it

app/actions/user.ts
ladytrekker: It's refactored

lib/utils/getAllowance/index.ts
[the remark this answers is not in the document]
ladytrekker: That´s the thing, 
`getAllowance` is an async function
So you need to wrap the whole reduce function in an async await in order to not loose the information which token allowance value belongs to which spender. If you want to nest this information still.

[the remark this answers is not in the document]
Atmosfearful: ok if you want to take a stab at nested async loops or promise arrays or something please go for it otherwise I guess I can live with this returning an object :)

Atmosfearful (approved):
good to go after we do another round of QA testing @ShimonD-KlimaDAO

app/components/views/PKlima/index.tsx
```diff
@@ -104,20 +129,28 @@ export const PKlima: FC<Props> = (props) => {
...
         provider,
         onStatus: setStatus,
       });
-      dispatch(redeemPklima(value));
+      dispatch(redeemPklima(approvedValue));
+      dispatch(
+        decrementAllowance({
+          token: "bct",
+          spender: "pklima_exercise",
+          value: approvedValue,
```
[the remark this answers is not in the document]
ladytrekker: But this is only true for future users.
All others which have once allowed the 1.000.000 gwei has to decrement their value each time.
Otherwise, if those users would immediately do another transaction, do not have an approval no more. Yes/No?

app/lib/hooks/useParamsSelector.ts
```diff
@@ -0,0 +1,8 @@
+import { useSelector } from "react-redux";
+
+export const useParamSelector = (
```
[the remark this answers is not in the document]
ladytrekker: I removed the hook and typed the useSelector hook instead.

lib/constants/index.ts
```diff
@@ -173,24 +192,14 @@ export const SANITY_STUDIO_API_PROJECT_ID = "dk34t4vc";
...
-  "nct",
-  "mco2",
-  "usdc",
-  "klima",
-  "sklima",
-  "wsklima",
-] as const;
-export type InputToken = typeof inputTokens[number];
+export const offsetInputTokens = allowancesContracts["retirementAggregator"];
+export type OffsetInputToken = typeof offsetInputTokens[number];
```
Atmosfearful, line 196: We can do the assertion here, it's not really safer but at least it lets us remove the `any` from our component code here https://github.com/KlimaDAO/klimadao/pull/495/files#diff-b68c42fa04a583195c0f40a15e663dcebdd59c469a31bac00a34e1848e2a6c06R110

```suggestion
/** Tokens accepted as input for the offset aggregator /#/offset */
export type OffsetInputToken =
  typeof allowancesContracts["retirementAggregator"][number];
export const offsetInputTokens = allowancesContracts[
  "retirementAggregator"
] as unknown as OffsetInputToken[];
```

ladytrekker: Thanks! Nested assertions. Interesting. Typescript is a puzzle sometimes.

lib/utils/getAllowance/index.ts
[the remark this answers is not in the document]
ladytrekker: I´d leave it as is right now.

Let´s improve in a later PR

app/components/views/PKlima/index.tsx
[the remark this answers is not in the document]
Atmosfearful: True, the way the UIs are built now, it doesnt do the approval flow if they already have an approval. And I guess the new transaction modal would check for an existing approval and skip that step if they did it already.

But I'm thinking now that it would be better to move all users to the new "as-needed" approvals. So when we build our new flow, it's a simple check:
```if (allowance !== value) { showApprovalStep() }```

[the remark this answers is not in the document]
Atmosfearful: I updated this issue to clarify that, also added a point about multiple-approvals
https://github.com/KlimaDAO/klimadao/issues/9

Atmosfearful (approved):

ShimonD-KlimaDAO:
@ladytrekker @Atmosfearful  here are my QA results. This is **not** quite ready yet for production.

I tried staking, wrapping, and offsetting.

Here is what I tried for each.
I removed all my allowances, and ensured Approval was triggered for the amount I wanted to use. After approve, I tried to up the amount and ensure Approval was triggered again, incrementing the allowance perfectly. I then tried updating the amount to less than the allowance and to exactly the allowance, and ensured Approval was not triggered as expected. Once the allowance was used up, I tried again to see if the frontend correctly updated the actual allowance I had as expected.

Staking works as expected. 
Bonding can't be tested - everything is sold out. I can circle back when new IB capacity comes online.

**Issues:**
1. Wrapping doesn't work as expected. Approval is triggered at first, but then upping the wrapping amount does not re-trigger it.
My allowance:
image

Trying 0.2:
image

Clicking "wrap" then results in a failure.

2. Offsetting doesn't account for 1% retirement fee. Example below. I approved 0.1 BCT to be retired. The cost is actually 0.101 BCT due to the fee. The approval went through (https://polygonscan.com/tx/0x8f6665d12dded6b9884c271f9a556652a6434a189db44988ab80f72ebca3bdd8), but I can't offset 0.1 BCT. I can offset 0.09 since my allowance also covers the fee.
image


**Notes:**
1. Unwrapping doesn't require approval, but unstaking does. I mention this in case there's code dealing with unwrapping allowances - it's unnecessary.
2. I will file a separate issue for this one. I discovered that bond redeem and autostake doesn't work, but it's unrelated to this PR (not something this PR broke).
3. Token allowances can be revoked on polygonscan here without having to tinker with solidity: https://polygonscan.com/tokenapprovalchecker

ladytrekker:
@ShimonD-KlimaDAO 🙇 

Thanks for the intensive testing !
Very good findings. 💯 
And thanks for the hint with the polygon link for revoking the allowances. I did not know this 😅 
 
> 
> **Issues:**
> 
> 1. Wrapping doesn't work as expected. Approval is triggered at first, but then upping the wrapping amount does not re-trigger it.

=> this is fixed now. 
I even found another bug because the wrap allowance value was not formatted correctly.


> 
> 2. Offsetting doesn't account for 1% retirement fee. Example below. I approved 0.1 BCT to be retired. The cost is actually 0.101 BCT due to the fee. 

=> this is fixed too. Very good find.

> **Notes:**
> 
> 1. Unwrapping doesn't require approval, but unstaking does. I mention this in case there's code dealing with unwrapping allowances - it's unnecessary.

Yes, this is how it is written too. I added more comments in the code to make this clear.

> 2. I will file a separate issue for this one. I discovered that bond redeem and autostake doesn't work, but it's unrelated to this PR (not something this PR broke).

Yes, Bonds are not part of this PR. There is still too much going on. I will refactor Bonds in a later PR.

> 3. Token allowances can be revoked on polygonscan here without having to tinker with solidity: https://polygonscan.com/tokenapprovalchecker

=> super nice! Thanks

PR #498, opened 2022-06-17
update metadescription strings

Atmosfearful:
## Description
Just some quick string changes

### DIFF
--- a/site/components/pages/errors/Custom404.tsx
+++ b/site/components/pages/errors/Custom404.tsx
@@ -13,17 +13,16 @@ export const Custom404 = () => {
     <div className={styles.errorPageWrapper}>
       <PageHead
         title={t({
-          id: "error.page.404.title",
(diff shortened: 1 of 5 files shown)


site/components/pages/errors/Custom404.tsx
```diff
@@ -13,25 +13,24 @@ export const Custom404 = () => {
     <div className={styles.errorPageWrapper}>
       <PageHead
         title={t({
-          id: "error.page.404.title",
+          id: "shared.404",
           message: "404 - Page Not Found",
         })}
         mediaTitle={t({
-          id: "error.page.404.head.title",
+          id: "shared.404",
```
ladytrekker, line 20: What are the rules for using "shared"?
This page is used only once, that is why I am wondering 🤔

ladytrekker (approved):
Thanks

PR #503, opened 2022-06-21
Refactor: Move Contracts and Abi to lib 🧹 

ladytrekker:
## Description

This is another preparation PR for ticket #9 
- map all token contracts to their corresponding Abi JSON files
- replace all Contracts with the new `getContract` lib util

This will make it easier later to create contracts objects like

```TS
const allContracts = createContracts(["bct", "nbo", "klima"]);
const allBalances = getBalancesFromContracts(allContracts);
const allAllowances = getAllowancesFromContracts(allContracts);
...
```
**TODO**
- do the same for Bonds later in another PR

## Related Ticket

Part of #9 

## Checklist

### DIFF
--- a/app/actions/user.ts
+++ b/app/actions/user.ts
@@ -5,7 +5,7 @@ import { addresses } from "@klimadao/lib/constants";
 import {
   formatUnits,
   trimStringDecimals,
-  getContractByToken,
+  getContract,
 } from "@klimadao/lib/utils";
 import {
   setBalance,
@@ -24,47 +24,47 @@ export const loadAccountDetails = (params: {
 }): Thunk => {
   return async (dispatch) => {
     try {
-      const bctContract = getContractByToken({
+      const bctContract = getContract({
         token: "bct",
         provider: params.provider,
       });
-      const nctContract = getContractByToken({
+      const nctContract = getContract({
         token: "nct",
         provider: params.provider,
       });
-      const mco2Contract = getContractByToken({
+      const mco2Contract = getContract({
         token: "mco2",
         provider: params.provider,
       });
-      const uboContract = getContractByToken({
+      const uboContract = getContract({
         token: "ubo",
         provider: params.provider,
       });
-      const nboContract = getContractByToken({
+      const nboContract = getContract({
         token: "nbo",
         provider: params.provider,
       });
-      const usdcContract = getContractByToken({
+      const usdcContract = getContract({
         token: "usdc",
         provider: params.provider,
       });
-      const klimaContract = getContractByToken({
+      const klimaContract = getContract({
         token: "klima",
         provider: params.provider,
       });
-      const sklimaContract = getContractByToken({
+      const sklimaContract = getContract({
         token: "sklima",
         provider: params.provider,
       });
-      const wsklimaContract = getContractByToken({
+      const wsklimaContract = getContract({
         token: "wsklima",
         provider: params.provider,
       });
-      const pKlimaContract = getContractByToken({
+      const pKlimaContract = getContract({
         token: "pklima",
         provider: params.provider,
       });
-      const klimaDomainContract = getContractByToken({
+      const klimaDomainContract = getContract({
         token: "klimaNameService",
         provider: params.provider,
       });
--- a/app/actions/user.ts
+++ b/app/actions/user.ts
@@ -66,12 +66,10 @@ export const loadAccountDetails = (params: {
         token: "pklima",
         provider: params.provider,
       });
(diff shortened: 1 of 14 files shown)


lib/utils/getContract/index.ts
```diff
@@ -0,0 +1,63 @@
...
+} as ContractMap;
+
+export const getContractAbiByToken = (token: Token) => {
+  return contractMap[token as keyof ContractMap];
+};
+
+export const getContractByToken = (params: {
+  token: Token;
+  provider: providers.JsonRpcProvider | Signer;
+}): ethers.Contract => {
```
Atmosfearful, line 55: some naming and type suggestions. Everything in our addresses object is a contract, although we might not have ABIs for all of them.

```suggestion
export const getContract = (params: {
  contractName: keyof typof addresses.mainnet;
  provider: providers.JsonRpcProvider | Signer;
}): ethers.Contract => {
```

app/actions/user.ts
```diff
@@ -24,62 +24,50 @@ export const loadAccountDetails = (params: {
...
+        provider: params.provider,
+      });
+      const pKlimaContract = getContractByToken({
+        token: "pklima",
+        provider: params.provider,
+      });
+      const klimaDomainContract = getContractByToken({
+        token: "klimaNameService",
+        provider: params.provider,
+      });
```
Atmosfearful, line 70: this could be loopified to save a few lines
```ts
const [bctContract, nctContract, ...etc] = ["bct", "nct", ...etc].map(token => getContractByToken({ token, provider}))
```

lib/utils/getContract/index.ts
```diff
@@ -0,0 +1,63 @@
...
+import IERC20 from "../../abi/IERC20.json";
+import WSKLIMA from "../../abi/wsKlima.json";
+import DistributorContractv4 from "../../abi/DistributorContractv4.json";
+import PunkTLD from "../../abi/PunkTLD.json";
+import PairContract from "../../abi/PairContract.json";
+import KlimaRetirementAggregator from "../../abi/KlimaRetirementAggregator.json";
+import ExercisePKlima from "../../abi/ExercisepKLIMA.json";
+import KlimaStakingHelper from "../../abi/KlimaStakingHelper.json";
+import KlimaStakingv2 from "../../abi/KlimaStakingv2.json";
+import KlimaRetirementStorage from "../../abi/KlimaRetirementStorage.json";
```
Atmosfearful, line 13: see my discord dm about tree-shaking...

we **may** need to rethink this to optimize that certain views like `/pledge/` only load the relevant ABIs as all combined are ~150kb raw.

150kb is not the end of the world, though, and is probably minified further during the build/webpack, and nextjs initial pageload should still be fast because its server rendered html.

Soooo what do you think? Worth the effort to optimize?

lib/utils/getContract/index.ts
```diff
@@ -0,0 +1,63 @@
...
+
+type Token = keyof typeof addresses["mainnet"];
+type ContractMap = {
+  [K in Token]: typeof IERC20["abi"] | typeof KlimaStakingHelper["abi"];
+};
+const contractMap = {
```
Atmosfearful, line 19: i think this is a good enough solution

if we do decide to make a single-source of truth config, that would be ideal, but that's also a big refactor because it would lead to us refactoring the `addresses` object entirely in `@klimadao/lib/constants/index.tsx`

we use `addresses.mainnet` in soo many places. So for the sake of avoiding the refactor dominos lets revisit later?
image

Atmosfearful (approved):
some ideas and thoughts

app/actions/user.ts
ladytrekker: This is the refactoring I want to do in a follow-up

```TS
const contracts = createContractsObject("bct", "nct", "ubo"); 
// => {
    bct: ethers.Contract,
    nct: ethers.Contract
    ubo: ethers.contract
...
}

lib/utils/getContract/index.ts
ladytrekker: Yes, let´s do this in a follow-up
This PR is already quite delicate.

lib/utils/getContract/index.ts
ladytrekker: Done

lib/utils/getContract/index.ts
ladytrekker: For the sake of simplicity I would leave the ABIs as is for now.
One contract, one Abi

Let's rather improve step-by-step

PR #526, opened 2022-07-07
Updated klimaDAO logo

### LINKED ISSUE #426: KilmaDAO logo update
Current logo is not brand aligned. Please find attached the right one

Image

biwano:
## Description

Updates klimaDAO logo

## Related Ticket

Resolves #426

## Checklist


lib/components/Logos/styles.ts
```diff
@@ -3,6 +3,7 @@ import breakpoints from "../../theme/breakpoints";
 
 export const logoWithClaim = css`
   max-width: 15rem;
+  padding-right: 1rem;
```
Atmosfearful, line 6:
```suggestion
  padding-inline-end: 1rem;
```
as a habit, always avoid `right` or `left` so that the layout should work when we enable an RTL language

Atmosfearful (changes requested):

Atmosfearful (approved):

PR #541, opened 2022-07-20
DRAFT: Implemented custom transaction options for all transactions

### LINKED ISSUE #519: Better gas estimations
The metamask gas calculation is not always correct, which can cause transactions to fail

Solution:
- when the app loads, use https://polygonscan.com/gastracker to query the gasPrice for "Rapid" transactions
- add 10%
- store this value as state property in redux
- for all transaction method calls, pass this value down as an override
`contract.retireCarbon(...args, { gasPrice })`
 https://docs.ethers.io/v5/api/contract/contract/#Contract-functionsCall
 
# Other ideas
Ping the api before each transaction and keep the value in-scope rather than saving it in redux?

biwano:
## Description

Implemented custom transaction options for all transactions. In the last PR it was implemented only for the stake transaction.

This PR is marked as draft because the following tests should be done (maybe by different persons).

- [x] Bond: Approve
- [x] Bond: Deposit (inverse_usdc)
- [ ] Bond: Deposit (other asset)
- [ ] Bond: Redeem
- [x] Offset: Approve
- [x] Offset: Retire carbon
- [x] Offset: Retire carbon with specific address
- [ ] PKlima: Approve
- [ ] PKlima: Exercise
- [ ] Stake: Approve Stake
- [X] Stake: Approve Unstake
- [X] Stake: Stake
- [X] Stake: Unstake
- [x] Wrap: Approve
- [x] Wrap: Wrap
- [x] Wrap: Unwrap

## Related Ticket

Resolves #519

## Checklist

### DIFF
--- a/app/actions/bonds.ts
+++ b/app/actions/bonds.ts
@@ -11,6 +11,7 @@ import BondCalcContract from "@klimadao/lib/abi/BondCalcContract.json";
 import OhmDai from "@klimadao/lib/abi/OhmDai.json";
 import IERC20 from "@klimadao/lib/abi/IERC20.json";
 import KlimaProV2 from "@klimadao/lib/abi/KlimaProV2.json";
+import { getTransactionOptions } from "@klimadao/lib/utils";
 
 const getBondAddress = (params: { bond: Bond }): string => {
   return {
@@ -336,7 +337,11 @@ export const changeApprovalTransaction = async (params: {
     const approvalAddress = getBondAddress({ bond: params.bond });
     const value = ethers.utils.parseUnits("1000000000", "ether");
     params.onStatus("userConfirmation", "");
-    const txn = await contract.approve(approvalAddress, value.toString());
+    const txn = await contract.approve(
+      approvalAddress,
+      value.toString(),
+      await getTransactionOptions()
+    );
     params.onStatus("networkConfirmation", "");
     await txn.wait(1);
     params.onStatus("done", "Approval was successful");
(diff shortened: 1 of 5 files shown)


Atmosfearful:
@biwano I'm getting second opinions on this from the gigabrains, from my own metamask gas estimations I can't tell if this is really solving the problems we had (or making any difference at all)

biwano:
> @biwano From my own metamask gas estimations I can't tell if this is really solving the problems we had (or making any difference at all) So I'm getting second opinions on this from the gigabrains in private chat, will let you know what they say.

Roger that.

Atmosfearful (changes requested):
@biwano OK I finally found the motivation to dive in and experiment with this.
Indeed we want to be setting maxFeePerGas and maxPriorityFeePerGas, not gasPrice. You can get those values from this url:
https://docs.polygon.technology/docs/develop/tools/polygon-gas-station/
and pass in an object like this:
```
  {
    maxFeePerGas: ethers.utils.parseUnits("60", "gwei"),
    maxPriorityFeePerGas: ethers.utils.parseUnits("60", "gwei"),
  }
```

In this example above I hardcode values, but you can also get these from the polygonscan gasstation too.
I would advise the following:
- [ ] let's only do this for the offset retirement transaction. I've realized the others like Stake and wrap etc. fail much more rarely. Another reason I only want to try on the offset aggregator for now is I noticed values fluctuate wildly from the gas station so I am skeptical that this will really help.
- [ ] For the offset view, fallback to `60 gwei` if the gas station API fails.
- [ ] Keep the 10% increase as before
- [ ] NOTE that the property names are different `maxFee (gas station) -> maxFeePerGas (ethers)`
- [ ] I recommend just close this PR and start fresh from the latest staging (sorry!)
- [ ] be sure to update the other ones that we already changed on production

Issue #489, opened 2022-06-15
Pledge Dashboard Touchups and Feedback


ShimonD-KlimaDAO:
1. (low-medium) I don't see Torus integration like we have in our dapp. If the staging build doesn't have the latest updates, I would consider doing a rebase of the pledge branch onto master to make sure nothing new breaks. 

2.  (low) I'm not a fan of how the cursor flashes in the middle of the search text. This is the first thing users see before connecting their wallets, so stylistically important. Also, capitalization seems incorrect. Finally, when I connect my wallet I expect to be redirected to my pledge page with a search option - not to have to copy-paste my address to get to my page.
![image](https://user-images.githubusercontent.com/99001851/173714722-93fb25b6-46bc-4de2-9ac6-395702c54c6e.png)

3. (low-medium) I don't see my KNS domain nor my KNS profile pic. Can those be pre-filled when the wallet is connected and has a KNS domain? Again this might be related to the branch needing to be rebased onto master.
![image](https://user-images.githubusercontent.com/99001851/173714983-c082fd6e-bace-47c6-8eb1-c323a095e4d3.png)

4. (Medium) See image in (3). I would suggest to stick to the same capitalization style (all caps) across the page. If we don't, the title capitalization is still wrong according to pretty much all grammar style standards ("Offset" not "offset"). Also, the past tense feels a bit off. See example:
https://capitalizemytitle.com/style/MLA
![image](https://user-images.githubusercontent.com/99001851/173715275-8b7ffd3e-d47b-4e7b-b85a-f9cfbe14309d.png)

5. (low-medium) Why is this italicized and in quotations? I would suggest removing the text entirely if there is no pledge/methodology to make it appear cleaner, or at least removing the quotations. Also, the "Methodology" icon of the verified person doesn't really make sense to me.
![image](https://user-images.githubusercontent.com/99001851/173715434-23589280-564d-451a-ae64-4c90b864cac1.png)
![image](https://user-images.githubusercontent.com/99001851/173715721-119abf2a-977d-45a0-a0df-c9354aa87d50.png)

6. (low-medium) Where is UBO, NBO, NCT, and wsKLIMA? Also, I would suggest removing "Holding" or just putting it in the title. When it appears next to each asset it's a bit crowded and also almost looks like there's another option that can be chosen in lieu of "Holding".
![image](https://user-images.githubusercontent.com/99001851/173715617-7968dab8-9798-43d8-8711-7834f37bd4b6.png)

[...]

PR #545, opened 2022-07-21
pledge dashboard tweaks

0xAeterno:
## Description

Addresses the majority of display related feedback raised through QA:
- Changes `methodology` character limit to 1000 (arbitrary number)
- Changes `description` character limit to 500.
- Update methodology card icon,
- Display footprint percent values to 2 decimal points (on chart tooltip and table)
- Footprint category:
  - Made the call to remove text overflow and just wrap the text instead of showing tooltip on hover.
  - Limit category naming to 32 characters, the equivalent of two rows on desktop view. We can revisit these character limit if the need arises.

## Related Ticket

https://github.com/KlimaDAO/klimadao/issues/489

## Changes

| Before  | After  |
|---------|--------|
|image|image|

## Checklist


Atmosfearful (approved):
💅

PR #537, opened 2022-07-19
Use same Modal component

ladytrekker:
## Description

This is another preparation PR for ticket #9 
This PR adds a new generic component for App "Modal" and refactored all occasions where a modal is shown to use this same component.

Please Note: I ignored the ClaimExceededModal as this is only relevant for PKlima view which is not maintained no more.

## Related Ticket

Part of #9 

## Changes

| Before  | After  |
|---------|--------|
| Bildschirmfoto 2022-07-19 um 12 35 08|Bildschirmfoto 2022-07-19 um 12 44 15 |
|Bildschirmfoto 2022-07-19 um 12 49 11 | Bildschirmfoto 2022-07-19 um 13 00 39 | 
| |Bildschirmfoto 2022-07-19 um 13 49 12  | 
| Bildschirmfoto 2022-07-19 um 13 05 18 |
Bildschirmfoto 2022-07-19 um 13 25 58<br>
Bildschirmfoto 2022-07-19 um 13 26 09<br>
Bildschirmfoto 2022-07-19 um 13 26 21<br>
Bildschirmfoto 2022-07-19 um 13 26 32 | 

## Checklist

- [x] Building the site with `npm run build-site` works without errors
- [x] Building the app with `npm run build-app` works without errors
- [x] I formatted JS and TS files with running `npm run format-all`

PR #546, opened 2022-07-22
Add InvalidNetworkModal to site/pledge

### LINKED ISSUE #507: [pledge] add 'wrong network modal' to site
Good catch from @ShimonD-KlimaDAO we haven't migrated this modal yet. It should be easy to port over.

0xAeterno:
## Description

Add modal to `/pledges` to prompt user to switch to the polygon network when they're on the wrong network.

#### Tech debt
The code is more a less a copy of the `InvalidNetworkModal` on app but uses the modal component created on site.
We should combine and create a generic modal component in lib that can be used across both sites. Issue raised here (todo)

## Related Ticket

Resolves #507 

## Changes

image

## Checklist


Atmosfearful:
This will conflict with https://github.com/KlimaDAO/klimadao/pull/537, so i'll wait until we merge that

0xAeterno:
@Atmosfearful pr rebased onto staging

Atmosfearful (approved):
I'm not opposed to copypasta but if you want to make this a shared lib component, I don't see why not.
You just have to pass in the lingui string as prop, since `lib` doesnt have lingui

PR #552, opened 2022-07-27
Add Poppins font to all headings

### LINKED ISSUE #424: Add Poppins Google font tag for headings
Designers want us to use Poppins

Design system here:
https://www.figma.com/file/GtTkd4y6ksWyrDwg5g5YDd/%5B-Design-System-%5D?node-id=1126%3A639

This is required for the KI page #491 #493 #540 #550

* [ ] add the google font script tag to support Poppins and remove any unused sizes for Inter
* [ ] update the typography tags

Atmosfearful:
## Description
Matches the Figma design system now (source of truth)

Resolves #424

## Checklist


lib/theme/typography.ts
```diff
@@ -2,6 +2,7 @@ import { css } from "@emotion/css";
 import breakpoints from "./breakpoints";
 
 export const h1 = css`
+  font-family: "Poppins";
```
0xAeterno, line 5: What do you think of setting `font-family` as a css var?

And probably a question for @Go-Mori, is it worth having a fallback font?

0xAeterno (approved):
LGTM :)
Nothing looked out of place on visually scanning through the pages 🚀

PR #559, opened 2022-07-29
fix: do not translate hard coded values

ladytrekker:
## Description

The infinity card values are all hard coded.
No need to wrap them in translation tags.

## Checklist


Atmosfearful (approved):
ty ty ty

PR #561, opened 2022-07-30
site/pledges: extract strings for i18n

0xAeterno:
## Description

Extract strings from `/pledges/*` for translations.

Translating form error message strings is not as straight forward as utilising lingui macros so I'm going to pull that out into a separate ticket to be worked on :)

## Related Ticket

Resolves #<issue-number>

## Checklist

### DIFF
--- a/site/components/pages/Pledge/PledgeDashboard/Profile/index.tsx
+++ b/site/components/pages/Pledge/PledgeDashboard/Profile/index.tsx
@@ -1,4 +1,6 @@
 import React, { FC, useEffect, useState } from "react";
+import { useRouter } from "next/router";
+import { Trans } from "@lingui/macro";
 import { Text } from "@klimadao/lib/components";
 import {
   concatAddress,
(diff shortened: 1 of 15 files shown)


site/components/pages/Pledge/PledgeDashboard/Profile/index.tsx
```diff
@@ -46,24 +50,26 @@ export const Profile: FC<Props> = (props) => {
...
 
   const renderPledgeProgress = () => {
     if (!displayPledgeProgress) return null;
 
     if (pledgeProgress >= 100) {
       return (
         <Text t="h4" className={styles.pledgeProgress}>
-          &ge; 100% of Pledge Met
+          <Trans id="pledges.dashboard.profile.pledge_met">
+            &ge; 100% of Pledge Met
```
ladytrekker, line 62: Interesting, why not use "≥". instead?

ladytrekker (approved):
Thanks

site/components/pages/Pledge/PledgeDashboard/Profile/index.tsx
0xAeterno: Oh yeah good one, not sure why I didn't go for the simple approach 😆

Atmosfearful (approved):

PR #564, opened 2022-08-01
added lifi link to offset and updated copy

### LINKED ISSUE #440: Link to Li.finance for cross-chain staking and offsetting
* [x] add to stake page
* [ ] add to offset page (same as above)
* [ ] add a click tracker for our analytics

goatsOnDrugs:
## Related Ticket

Resolves #440 

## Checklist

### DIFF
--- a/app/components/views/Offset/index.tsx
+++ b/app/components/views/Offset/index.tsx
@@ -62,6 +62,7 @@ import KLIMA from "public/icons/KLIMA.png";
 import USDC from "public/icons/USDC.png";
 import UBO from "public/icons/UBO.png";
 import NBO from "public/icons/NBO.png";
+import FiberNewRoundedIcon from "@mui/icons-material/FiberNewRounded";
 
 import * as styles from "./styles";
 import { cx } from "@emotion/css";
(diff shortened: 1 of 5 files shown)


app/components/views/Offset/index.tsx
```diff
@@ -448,27 +449,16 @@ export const Offset = (props: Props) => {
...
-              </A>{" "}
-              (NBO), with more coming soon.{" "}
+              environmental benefit of the carbon offset. Learn more about
+              carbon tokens in our <A href={urls.officialDocs}>docs</A>.
+            </Trans>
+          </Text>
+          <Text t="caption" color="lightest">
+            <FiberNewRoundedIcon className="new-releases-icon" />
+            <Trans id="offset.lifi">
+              Cross-chain staking is now available through{" "}
```
Atmosfearful, line 459: see the text i sent in dm, it's a little different

Atmosfearful (changes requested):

app/components/views/Offset/index.tsx
```diff
@@ -448,27 +449,16 @@ export const Offset = (props: Props) => {
...
-              <A href="https://docs.c3.app/c3-pool-specifications-and-carbon-methodologies/nbo">
-                Nature Based Offsets
-              </A>{" "}
-              (NBO), with more coming soon.{" "}
+              environmental benefit of the carbon offset. Learn more about
+              carbon tokens in our <A href={urls.officialDocs}>docs</A>.
+            </Trans>
+          </Text>
+          <Text t="caption" color="lightest">
+            <FiberNewRoundedIcon className="new-releases-icon" />
```
ladytrekker, line 457: Just a note: I prefer direct exported styles

so Like

```TS
export const newReleaseIcon = css` 
...
```

- because easier to extract if you need the component elsewhere too
- the styles are only for this component and are not inheriting any styles from parent
- and in your editor you can directly jump to the styles.

ladytrekker (approved):
Thanks

app/components/views/Offset/index.tsx
goatsOnDrugs: good point! updated.

app/components/views/Offset/index.tsx
goatsOnDrugs: updated

Atmosfearful (approved):

Issue #568, opened 2022-08-02
Retirement receipt 404


Atmosfearful:
From riitaka. Probably subgraph delay.
> Hi guys, after I executed this retirement using MetaMask on mobile: https://polygonscan.com/tx/0xf9242c93dc91c0a4bd3bad0616ae69fd097040a15bde997dc7bc772912ea32e5
I got the attached error when clicking to view and share my retirement page + the retirement doesn’t appear on the aggregated overview page https://www.klimadao.finance/retirements/rittaaka.klima
EDIT: the tx now appears on the retirement overview but when I click it, I get another 404 

- can we improve subgraph speed?
- should we add a fallback state?
in theory the RPC node should have data before the subgraph, so we could query the rpc and see if a retirement exists. If it does, we can show something like 'On-chain data is being indexed, this may take a few minutes.'

ladytrekker:
Not sure how to increase the speed of the subgraph.

What I would do now:

- divide fetching data from RPC and subgraph (currently [wrapped in promise.all](https://github.com/KlimaDAO/klimadao/blob/staging/site/pages/retirements/%5Bbeneficiary%5D/%5Bretirement_index%5D.tsx#L90-L102))
- fill missing data with "Loading data..." and the text 'On-chain data is being indexed, this may take a few minutes.'
- in `useEffect` start a timer for fetching the data from subgraph every XY seconds

PR #571, opened 2022-08-02
Retirements: Check for empty subgraph data

ladytrekker:
## Description

Users reported that the Retirement Receipt page returns a 404 after a successful Offset.

Under the assumption that this error occurs **when the subgraph is querying the data not fast enough**
this PR does the following:

- ensure that all queries return `false` on empty query results (Subgraph returns empty array on not found!)
- If the data is `false` => pass `null` to the Page props
- During rendering add fallback for empty data or render nothing

## Related Ticket

Closes https://github.com/KlimaDAO/klimadao/issues/568

## SCREENSHOTs

Spot the difference 😄 

| With Subgraph Data  | Without Subgraph Data  |
|---------|--------|
| |   |
|  |  |

## Checklist

### DIFF
--- a/lib/utils/subgraph/queryPolygonBridgedCarbon.ts
+++ b/lib/utils/subgraph/queryPolygonBridgedCarbon.ts
@@ -55,7 +55,7 @@ export const queryKlimaRetireByIndex = async (
 
 export const queryKlimaRetiresByAddress = async (
   beneficiaryAddress: string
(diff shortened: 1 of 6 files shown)


Atmosfearful (approved):
good fix, lets discuss whether we want to enhance this with a full page loading state as I mentioned in DMs

PR #601, opened 2022-08-12
Goats/grid

### LINKED ISSUE #532: New Nav header + mobile menu
The current header is not scaling as we increase the number of products and landing pages.
@Go-Mori has completed a design  https://www.figma.com/file/GtTkd4y6ksWyrDwg5g5YDd/%5B-Design-System-%5D?node-id=1133%3A2881

This is a blocker for completing work on the upcoming Resources section #534, but we can start on both in parallel.

Edit: This is ready to go! We are not implementing this in the dapp or /pledge pages yet.

### LINKED ISSUE #563: [app ui] Utilize dead space at top
image

Currently, the content is placed in the center row and below. Can we move it up to the top row?
image
 This may require a rethinking of the grid layout that we use for the dapp.
 
Please make yourself aware of a past issue we had, to make sure this is solved as well: https://github.com/KlimaDAO/klimadao/issues/273
https://github.com/KlimaDAO/klimadao/issues/264

goatsOnDrugs:
## Description
Grid updated to use dead space at the top. Some cards now scroll. UX feels the same or better in my opinion because you can see the other cards and the nav while you scroll  Let me know what you think. I didnt add images because it effects every page and that would be a lot of images hah.

## Related Ticket

Resolves #563 

## Checklist

### DIFF
--- a/app/components/views/Bond/styles.ts
+++ b/app/components/views/Bond/styles.ts
@@ -33,7 +33,7 @@ export const bondCard = css`
 
   ${breakpoints.desktop} {
     grid-column: cardsleft;
-    grid-row: 2 / span 3;
+    grid-row: 1 / span 3;
     gap: 4.8rem;
     grid-template-rows: 1fr 1fr 1fr;
     align-items: start;
--- a/app/components/views/Buy/styles.ts
+++ b/app/components/views/Buy/styles.ts
@@ -11,7 +11,13 @@ export const buyCard = css`
   border-radius: 1.2rem;
   padding-top: 2.4rem;
   gap: 2.4rem;
-
+  overflow: scroll;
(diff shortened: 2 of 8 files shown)


0xAeterno (changes requested):
I think we should get the green light from @theprofessora before moving this ticket along.
IMO I think the misalignment of the two columns looks strange, we can explore a better way to utilise the space when we prioritise revisiting the apps design/

Other feedback:
There shouldn't be whitespace between the text on `/info`
image

Atmosfearful:
Indeed I highlighted some of the issues that this layout presents, and we may need to revisit when we upgrade the Offset tool, because this will probably get some more elaborate confirmation screens that remove the columns on the right. I will close the PR for the time being until we know what direction this layout is going!

PR #617, opened 2022-08-17
CMS: Update lock files

ladytrekker:
## Description

After `cd cms  && npm install` these files are changed.

- Is this correct?
- Does the same happen on your machine?


Atmosfearful:
@ladytrekker check out the repo README I mention there that the cms is a bit different:

```
npm install -g @sanity/cli
cd cms && sanity install
```

Issue #531, opened 2022-07-18
Create KlimaDAO branded PDF receipts for retirements


Atmosfearful:
# Summary
Customers are asking for old-school PDF certificates similar to those created by the Verra registry.
The original plan was to get these from the underlying bridge (c3 or toucan). But we don't want to wait anymore.
Now we think it makes sense to generate a KlimaDAO specific one.

# PDF Requirements
See design https://www.figma.com/file/RtYPQewQkz4RENj81ugUnc/PDF-Certificate?node-id=0%3A1

# Technical Requirements
- Nextjs API endpoint that accepts `address` and `index` and responds with a PDF download
- Cache response for 1 day
- Use a lightweight JavaScript PDF generator library for server-side pdf generation, like `JSPDF` or `PDFMake` or `PDFKit`
- "Download PDF" button on all retirement pages (where the "coming soon" button is)

jabby09:
@Atmosfearful here is the figma file: https://www.figma.com/file/RtYPQewQkz4RENj81ugUnc/PDF-Certificate

PR #618, opened 2022-08-17
Root: Update package.lock

ladytrekker:
## Description

Removing `node_modules` folder and do a fresh install with `npm install` these changes appear in package.lock.

- Is this correct?
- Does the same happen on your machine?


Atmosfearful (approved):
yup looks like @0xAeterno did something funky in #531

Issue #615, opened 2022-08-16
Do extra 1% approval for retireCarbon transaction


Atmosfearful:
as discussed in DMs, it's possible the price can slip upward between the time we call `offsetConsumption()` to query the price, submit approval, and finally submit the final transaction. We need to approve a little bit extra similar to what sushiswap does.

PR #620, opened 2022-08-17
Add 1% to approval value for offset

ladytrekker:
## Description

As described in the ticket https://github.com/KlimaDAO/klimadao/issues/615 the price between approval and final transaction can change. To not create a bad user experience which would result in errors in metamask, let's rather increase the approval amount.

The TXN modal is extended now to accept `approvalValue` for the first screen. 
And falls back to the actual `value` if missing.

## Related Ticket

Resolves https://github.com/KlimaDAO/klimadao/issues/615

## Checklist

### DIFF
--- a/app/components/views/Offset/index.tsx
+++ b/app/components/views/Offset/index.tsx
@@ -244,9 +244,6 @@ export const Offset = (props: Props) => {
       const token = selectedInputToken;
       const spender = "retirementAggregator";
 
-      console.log("COST", cost);
-      console.log("APPROVAL", getApprovalValue());
-
       const approvedValue = await changeApprovalTransaction({
         value: getApprovalValue(),
         provider: props.provider,
--- a/app/components/views/Offset/index.tsx
+++ b/app/components/views/Offset/index.tsx
@@ -49,6 +49,8 @@ import {
   urls,
 } from "@klimadao/lib/constants";
 
+import { getTokenDecimals } from "@klimadao/lib/utils";
+
 import { tokenInfo } from "lib/getTokenInfo";
 
 import { CarbonTonnesRetiredCard } from "components/CarbonTonnesRetiredCard";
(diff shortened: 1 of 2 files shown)


app/components/views/Offset/index.tsx
```diff
@@ -226,20 +231,30 @@ export const Offset = (props: Props) => {
...
   const handleApprove = async () => {
     try {
       if (!props.provider) return;
-      const value = cost.toString();
+
       const token = selectedInputToken;
       const spender = "retirementAggregator";
 
+      console.log("COST", cost);
+      console.log("APPROVAL", getApprovalValue());
```
Atmosfearful, line 248: loggies

app/components/views/Offset/index.tsx
```diff
@@ -226,20 +231,30 @@ export const Offset = (props: Props) => {
     setBeneficiaryAddress(address);
   };
 
+  const getApprovalValue = (): string => {
+    const costAsNumber = Number(cost);
+    const costPlusOnePercent = costAsNumber + costAsNumber * APPROVAL_SLIPPAGE;
+    return costPlusOnePercent.toFixed(18); // ethers does not like more than 18 digits and returns with "underflow" otherwise
```
Atmosfearful, line 237: this depends on the token, I believe USDC approval would still throw underflow at 18 decimals, since its a 6 decimal token

Atmosfearful (changes requested):

app/components/views/Offset/index.tsx
ladytrekker: Uupsi, removed

app/components/views/Offset/index.tsx
ladytrekker: Thanks! Good to know.

I pushed the fix

Atmosfearful (approved):

PR #642, opened 2022-08-23
card font update

goatsOnDrugs:
## Description
Updating the font size for cards on infinity page. I am over-riding the styles set in Text bc none of the `body` styles will work. They are too big on desktop and too small on mobile so thats why `!important` is used. I can also add another `body` var if that would be better.

## Related Ticket

Resolves #<issue-number>

## Checklist

### DIFF
--- a/site/components/pages/Infinity/Cards/Card.tsx
+++ b/site/components/pages/Infinity/Cards/Card.tsx
@@ -54,9 +54,7 @@ export const Card: FC<Props> = (props) => {
         </Anchor>
       </div>
       <div className={styles.cardContent}>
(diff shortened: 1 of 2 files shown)


goatsOnDrugs:
Professor approved

Atmosfearful (approved):
works for now thanks!
but we should be 100% consistent with our typography system @theprofessora could you add the mobile/desktop sizes somewhere in here and give them a name: https://www.figma.com/file/GtTkd4y6ksWyrDwg5g5YDd/%5B-Design-System-%5D?node-id=1126%3A639
Then @goatsOnDrugs can add them to our typography file and use it here

theprofessora:
done

Issue #534, opened 2022-07-18
New /resources, blog, and podcast view


Atmosfearful:
# Summary
As the amount of content we create increases, and the complexity of our product grows, we need a new integrated design that exposes users to all of the various kinds of content and guides that we are creating. Creative team has come up with a nice new home for all of our content. This feature includes:
- A `tag` property for all documents in our CMS that lets us do more advanced querying
- A Sidebar that lets users select which tags to filter by in the UI
- A simple text search

### DESIGN
https://www.figma.com/file/NQTGQDrS7PEfB3QOzYMBN3/RESOURCES?node-id=5%3A833

### PREVIEW LINK

STAGING (merged code status): https://staging-site.klimadao.finance/resources
DEVELOPMENT (unmerged status): https://preview-cms.vercel.app/resources

---

## Todos for GO LIVE :tada:

=> moved to new ticket https://github.com/KlimaDAO/klimadao/issues/760

----

Completion is blocked by #532 because we need to make sure we don't lose access to our /community and /contact pages.

# Implementation Plan
Separate PRs:
* [x] Enhance the CMS to support "tags" property on Podcast and Blog documents => [PR](https://github.com/KlimaDAO/klimadao/pull/609)
- create a new top-level content type called `tag`. A `tag` is a document schema that has these properties:
```ts
interface Tag {
  // slug, not string, because Sanity Slugs can enforce uniqueness https://www.sanity.io/docs/slug-type
  value: Slug;
  // label shown in the nav UI (hardcoded for v1, might be queried from cms in the future)
  label_en: string;
  // just a string for notes, but maybe we can show this in the UI in the far future.
  description_en: string;
}
```
-  [x] for Podcast and Post documents, add a property to the schema: `tags: Reference[]` that is an array of references to tag documents. Sanity Studio users should be able to select from a dropdown of existing tags. See the docs for help on references. => [PR](https://github.com/KlimaDAO/klimadao/pull/609)

https://www.sanity.io/docs/reference-type

---

### SANITY
- [x] Please deploy the new tags to Sanity.io (already merged to Staging)
- [x] Get a list of all Tags the content editor want to use (lists needs to be hard coded in site)
- [x] We need another checkbox for the "FEATURED SLIDER" for Posts, like "is featured article" => [PR](https://github.com/KlimaDAO/klimadao/pull/644)
[...]

PR #650, opened 2022-08-26
Resources: Add Featured Articles slider

ladytrekker:
## Description

This is the next part of the new Resources page:

- build new slider for Resources Page
- query posts with flag "isFeaturedArticle"
- pass these articles to Slider

## Related Ticket

Part of https://github.com/KlimaDAO/klimadao/issues/534

## Preview

See https://klimadao-site-git-resources-featured-articles-slider-klimadao.vercel.app/resources

## Checklist

### DIFF
--- a/site/components/pages/Resources/FeaturedArticles/ArticlesSlider.tsx
+++ b/site/components/pages/Resources/FeaturedArticles/ArticlesSlider.tsx
@@ -83,10 +83,10 @@ export const ArticlesSlider: FC<Props> = (props) => {
   };
 
   return (
-    <div className={styles.cardsSliderContainer}>
-      <div className={styles.rerouselContainer}>
+    <div className={styles.container}>
+      <div className={styles.sliderContainerOuter}>
         <div
-          className={styles.sliderWrapper}
+          className={styles.sliderContainerInner}
           ref={sliderRef}
           onTouchStart={stopAutoSlide}
         >
--- a/site/components/pages/Resources/FeaturedArticles/ArticlesSlider.tsx
+++ b/site/components/pages/Resources/FeaturedArticles/ArticlesSlider.tsx
@@ -1,0 +1,127 @@
+import React, { FC, useEffect, useState, useRef } from "react";
+import ArrowForwardIosIcon from "@mui/icons-material/ArrowForwardIos";
+import ArrowBackIosNewIcon from "@mui/icons-material/ArrowBackIosNew";
+import { Article } from "./Article";
+
+import { useElementWidth } from "hooks/useElementWidth";
+import { Document } from "lib/queries";
+
+import * as styles from "./styles";
+
+type Props = {
+  articles: Document[];
+};
+
+export const ArticlesSlider: FC<Props> = (props) => {
+  const cardRef = useRef(null);
+  const sliderRef = useRef<HTMLDivElement>(null);
+
+  const [cardWidth] = useElementWidth(cardRef);
+  const [sliderWidth] = useElementWidth(sliderRef);
+  const [, setScrollInterval] = useState<NodeJS.Timeout>();
+  const [currentScrollLeft, setCurrentScrollLeft] = useState<number>(0);
+  const cardsLength = props.articles.length;
+
+  const hasReachedEnd = () => {
+    if (!!cardWidth && !!sliderWidth) {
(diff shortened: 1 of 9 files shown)


site/components/pages/Resources/FeaturedArticles/ArticlesSlider.tsx
```diff
@@ -0,0 +1,127 @@
...
+  articles: FeaturedPost[];
+};
+
+export const ArticlesSlider: FC<Props> = (props) => {
+  const articleRef = useRef(null);
+  const sliderRef = useRef<HTMLDivElement>(null);
+
+  const [articleWidth] = useElementWidth(articleRef);
+  const [sliderWidth] = useElementWidth(sliderRef);
+  const [, setScrollInterval] = useState<NodeJS.Timeout>();
```
Atmosfearful, line 21: I think `useRef` is a better fit for this, plus then it doesn't cause re-renders (unless you want re-render?)

site/hooks/useElementWidth/index.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  useEffect(() => {
+    firstUpdateWidth();
+    window.addEventListener("resize", updateWidth);
+    return () => {
+      window.removeEventListener("resize", updateWidth);
+    };
+  }, [updateWidth]);
+
+  return [width];
```
Atmosfearful, line 41: would be neat if you also manage the ref inside this hook, so the parent doesnt need to pass anything in
```suggestion
  return [width, ref];
```

```
const [articleWidth, articleRef] = useElementWidth()
```

site/components/pages/Resources/index.tsx
```diff
@@ -9,14 +9,23 @@ import { Navigation } from "components/Navigation";
...
+import { Document, FeaturedPost } from "lib/queries";
 
 export interface Props {
   documents: Document[];
+  featuredArticles: FeaturedPost[];
 }
 
-export const Resources: NextPage<Props> = ({ documents }) => {
+export const Resources: NextPage<Props> = ({ documents, featuredArticles }) => {
+  console.log("WORKS", featuredArticles);
```
Atmosfearful, line 22: loggies

site/hooks/useElementWidth/index.ts
```diff
@@ -0,0 +1,42 @@
...
+      setWidth(width);
+    }
+  }, [elementRef]);
+
+  const firstUpdateWidth = useCallback(() => {
+    if (elementRef && elementRef.current) {
+      let { width } = elementRef.current.getBoundingClientRect();
+      width =
+        width -
+        parseInt(
```
Atmosfearful, line 18: just a personal preference:

```suggestion
      const rect = elementRef.current.getBoundingClientRect();
      const width =
        rect.width -
        parseInt(
```

site/lib/queries.ts
```diff
@@ -94,6 +108,8 @@ export type PostDetails = {
   imageUrl?: string;
 };
 export type AllPosts = PostDetails[];
+export type FeaturedPost = PostDetails & { isFeaturedArticle: true };
```
Atmosfearful, line 111: do we support featured Podcasts up there too? Might be worth supporting if it's not too hard.

Atmosfearful (approved):
smol things! looks great!
Might have design nitpicks but let's get some good covers (without overlapping text) so we can enjoy it in it's true glory first

site/components/pages/Resources/index.tsx
ladytrekker: Yes these were there for you 😄 

Removed now as "real" featured articles are queried now

site/lib/queries.ts
ladytrekker: Not according to any design or concept.
I added it to the "Open UI questions" lists here https://github.com/KlimaDAO/klimadao/issues/534

site/hooks/useElementWidth/index.ts
ladytrekker: The hook is used for the infinity slider too.
Will leave it as is right now and let´s do this in a follow-up

site/components/pages/Resources/FeaturedArticles/ArticlesSlider.tsx
ladytrekker: Done

Issue #475, opened 2022-06-07
Update price calculations to avoid BCT/USDC pool


0xAurelius:
We are planning to deprecate the BCT/USDC pool, so any price calculations done using that pool need to be switched over to the BCT/KLIMA => KLIMA/USDC route instead

Atmosfearful:
This is higher prio because we want to deprecate the pools

kaxline:
Still kind of exploring this part of the app, but it looks like this would be the method to be concerned about, correct?

https://github.com/KlimaDAO/klimadao/blob/eaff5d85bb03d378e2c50803e36f9750fb1435cc/app/actions/bonds.ts#L94

And that looks like it's using the KLIMA/BCT address.

And then it looks like even if you input `bct_usdc_lp` into the `getMarketPrice` method, it still routes to that same method and so uses KLIMA/BCT as well:

https://github.com/KlimaDAO/klimadao/blob/eaff5d85bb03d378e2c50803e36f9750fb1435cc/app/actions/bonds.ts#L206

Do I have that right? Anywhere this may be a concern?

0xAurelius:
hmm yeah it looks like this has already been addressed, i can't think of anywhere else it might appear and searching for "usdc" in the repo doesn't seem to lead to any other uses of the BCT/USDC pool

we should just open a PR to remove the reference to the `bct_usdc_lp` since we won't need it anymore and the pool will be tiny after all DAO LP is pulled

Atmosfearful:
@kaxline go ahead and remove all references to bct_usdc_lp including the bond!

kaxline:
@0xAurelius @Atmosfearful Should we wait to remove the BCT portion of this method until the LP has been pulled:

https://github.com/KlimaDAO/klimadao/blob/17fcfad12c9ca0695d3518931e93a3db5a362be1/lib/utils/getTreasuryBalance/index.ts#L41

0xAurelius:
Ah yeah let's maybe do a separate PR for deprecating this, or just leave this PR unmerged until we have pulled the LP

I just didn't want us to move forward with dissolving the LP unless we are sure nothing will break when we do. This particular usage shouldn't break because the pool will still exist, there just won't be any LP held by the treasury so this will go to 0.

kaxline:
@0xAurelius Ok cool, I just created what I think is a comprehensive merge that can wait until the treasury is ready.

Atmosfearful:
[...]

PR #651, opened 2022-08-26
Removing all mentions and uses of BCT/USDC LP

kaxline:
## Description

Removing all mentions and uses of the BCT/USDC LP in preparation for the treasury no longer holding it.

## Related Ticket

#475 

Resolves #<issue-number>

## Checklist


lib/utils/getTreasuryBalance/index.ts
```diff
@@ -48,9 +48,8 @@ export const getTreasuryBalance = async (
     const nakedBCT = getInteger(
       await bctContract.balanceOf(addresses.mainnet.treasury)
     );
-    const bctUSDC = await getOwnedBCTFromSLP({ adr: "bctUsdcLp", provider });
     const bctKLIMA = await getOwnedBCTFromSLP({ adr: "klimaBctLp", provider });
-    const sum = nakedBCT + bctUSDC + bctKLIMA;
```
Atmosfearful, line 53: ohhhh this is one we didn't think about. We need to keep it, as our treasury still has LP and we want that shown in our total tonnes on the landing page.

I believe all you have to do is revert this line then add back the address in `lib/constants` and everything should work again

Atmosfearful (changes requested):
Next time `npm run extract-strings:dev` will only update en and pseudo which makes it a bit easier for me to review

Atmosfearful (approved):

PR #656, opened 2022-08-29
519 better gas price estimations

### LINKED ISSUE #519: Better gas estimations

biwano:
## Description

Fetch transaction options from polyscanGasStation (maxFeePerGas & maxPriorityFeePerGas) for offset retirement transactions only.

From @Atmosfearful in https://github.com/KlimaDAO/klimadao/pull/541
- [X] let's only do this for the offset retirement transaction. I've realized the others like Stake and wrap etc. fail much more rarely. Another reason I only want to try on the offset aggregator for now is I noticed values fluctuate wildly from the gas station so I am skeptical that this will really help.
- [X] For the offset view, fallback to `60 gwei` if the gas station API fails.
- [X] Keep the 10% increase as before
- [X] NOTE that the property names are different `maxFee (gas station) -> maxFeePerGas (ethers)`
- [X] I recommend just close this PR and start fresh from the latest staging (sorry!)
- [X] be sure to update the other ones that we already changed on production

## Related Ticket

Resolves #519

## Checklist

### DIFF
--- a/app/actions/offset.ts
+++ b/app/actions/offset.ts
@@ -1,6 +1,7 @@
 import { ethers, providers } from "ethers";
 import { Thunk } from "state";
 import { setCarbonRetiredBalances, updateAllowances } from "state/user";
+import { getTransactionOptions } from "@klimadao/lib/utils";
 
 import {
   addresses,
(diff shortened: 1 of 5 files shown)


Atmosfearful (approved):
Awesome! aced it!

PR #665, opened 2022-08-30
Goats/646 header cleanup

### LINKED ISSUE #646: Nav header cleanups
Merge #591 first and do these in a follow-up PR

# Related to
#591 
#532 

# Desktop issues
These first three can be solved by using tippy.js around a natively-focusable element like `button`
* [x] 1. Desktop dropdowns can not be reached with the Tab key
* [x] 2. When a dropdown is open, items in an open menu should also be reachable with the tab key
* [x] 3. We need aria labels to indicate to screen-readers that a popper is open. `aria-expanded` and `aria-describedby`  (tippy does by default)

# Other issues
* [x] Mobile menu items which are expandable should have `aria-expanded="false"` by default (`true` when expanded)
* [x] Mobile menu on narrow desktop Windows 10 screens has visible X and Y scrollbars. Probably using overflow: scroll, when you should be using overflow: auto?
image
* [x] When clicking a menu item in the mobile menu, it should not close the other one. This causes a jarring jump in the content which is disorienting. A menu should only close itself when you click it, or when the other
* [x] Remove the "mousedown" / click animation on the button. I always thought it was a bit overkill and it is particularly jarring on this new mobile menu
* [x] add 2.4rem padding-top to `hero_container` on mobile views only, because the white News banner clashes with the white sticky header
Before: 
image
After:
image

* [x] especially for string ids starting with `shared`, but as a general best-practice, the English source strings should not be capitalized, because we might use it somewhere that doesn't want it capitalized. Use `text-transform: uppercase` in css instead
For example:
```ts
// 🔴 bad
t({
    message: "COMMUNITY",
    id: "shared.community",
})}

// 🟢 good
t({
    message: "Community",
    id: "shared.community",
})}
```

# Stretch Goals
* [ ] Mobile menu should also support tab-navigation. This would require the creation of a "focus trap" so that focus can't leave the mobile menu until the user closes it.
* [x] When the mobile menu is open, the main content and/or `body` should have a scroll lock to prevent un-knowingly scrolling the content underneath when you are trying to scroll the menu list

goatsOnDrugs:
## Description
- Tippy was implemented on desktop for aria labels and tab support.
- aria-expanded added to mobile nav items
- scroll bar is hidden when mobile nav is open and the body scroll is locked. There is an issue of when the user closes the nav menu they jump back to the top of the screen. In order for this not to happen we need to set `position` and `top` on the body element when opening and closing the modal which I was unsuccessful in doing.
- mouse down event removed and replaced with the green state from figma
- mobile tabs dont close if you open another one
- tab trap for mobile was taking too long. I think it has something to do with next bc I was trying to use a ref and do refElement.current.focus() but it was always undefined for some reason I could not figure out.
-  2.4rem padding-top added to `hero_container`
-  all caps messages uncapped
- podcast link added and carbon dashboards -> carbon dashboard

## Related Ticket

Resolves #646 

## Checklist


site/components/Navigation/index.tsx
```diff
@@ -40,9 +41,6 @@ export const Navigation: FC<Props> = ({
   activePage,
   showThemeToggle = true,
 }) => {
-  const [toggledNavItemID, setToggledNavItemID] = useState<
-    NavItemMobileID | undefined
-  >(undefined);
```
ladytrekker, line 45: 👍

site/components/Navigation/NavItemMobile.tsx
```diff
@@ -8,34 +8,22 @@ interface Props extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
...
+    <div className={cx(styles.navMain_MobileItem)} aria-expanded={isToggled}>
       <ButtonPrimary
         href={props.url}
         className={cx(styles.navMain_MobileLink, {
-          active: isToggled,
+          activeItem: isToggled,
         })}
         label={props.name}
-        onClick={onToggleSubmenu}
+        onClick={() => setIsToggled(!isToggled)}
```
ladytrekker, line 26: better: `onClick={() => setIsToggled(current => !current)}`
to read the previous state

site/components/Navigation/NavItemDesktop.tsx
```diff
@@ -32,16 +32,24 @@ export const NavItemDesktop: FC<Props> = ({
...
+              <div className={styles.navMain_DesktopSubMenu}>{subMenu}</div>
+            )
+          }
+          interactive={true}
         >
-          {name}
-        </div>
+          <button
+            className={cx(styles.navMain_DesktopLink, {
+              dropdown: subMenu !== undefined,
```
ladytrekker, line 45: better : `dropdown: !!subMenu,`

site/components/Navigation/index.tsx
```diff
@@ -279,50 +263,48 @@ export const Navigation: FC<Props> = ({
             />
             <NavItemMobile
               name={t({ message: "App", id: "shared.app" })}
-              toggledNavItemID={toggledNavItemID}
-              setToggledNavItemID={setToggledNavItemID}
               id="App"
```
ladytrekker, line 266: You can now remove all `id` from every `NavItemMobile` here
remove the prop too
and delete the type `NavItemMobile`

site/components/Navigation/index.tsx
```diff
@@ -332,57 +314,21 @@ export const Navigation: FC<Props> = ({
             />
             <NavItemMobile
               name={t({ message: "Infinity", id: "shared.infinity" })}
-              toggledNavItemID={toggledNavItemID}
-              setToggledNavItemID={setToggledNavItemID}
               url="/infinity"
               id="Infinity"
```
ladytrekker, line 318: Delete `id`

site/components/Navigation/index.tsx
```diff
@@ -332,57 +314,21 @@ export const Navigation: FC<Props> = ({
...
-              //   url="/blog"
-              // />,
-              //   ]
-              // }
             />
             <NavItemMobile
               name={t({ message: "Love Letters", id: "shared.loveletters" })}
-              toggledNavItemID={toggledNavItemID}
-              setToggledNavItemID={setToggledNavItemID}
               id="Love Letters"
```
ladytrekker, line 322: Delete `id`

site/components/Navigation/index.tsx
```diff
@@ -332,57 +314,21 @@ export const Navigation: FC<Props> = ({
...
               url={urls.loveletter}
             />
             <NavItemMobile
               name={t({ message: "Resources", id: "shared.resources" })}
-              toggledNavItemID={toggledNavItemID}
-              setToggledNavItemID={setToggledNavItemID}
               id="Resources"
```
ladytrekker, line 327: Delete `id` ... and so on

ladytrekker (approved):
Thanks Goats

I tested in Chrome dev tools, looks good IMO

See some minor comments below

site/components/Navigation/NavItemMobile.tsx
goatsOnDrugs: TY, noted for the future. Removed onClickfor now.

site/components/Navigation/NavItemDesktop.tsx
goatsOnDrugs: Oh, `!!` I like this, I never use it. 👍

site/components/Navigation/index.tsx
goatsOnDrugs: removed

site/components/Navigation/index.tsx
goatsOnDrugs: All, ids deleted. I'm not sure its showing that they are there bc they were deleted in the commit a few days ago. 🤷 they are deleted tho haha.

goatsOnDrugs:
@ladytrekker thanks for the review!

site/components/Header/HeaderMobile.tsx
```diff
@@ -35,6 +35,17 @@ export const HeaderMobile: FC<Props> = (props) => {
...
+        styles={
+          isToggled &&
+          css`
+            body {
+              position: fixed;
+              height: 100vh;
+            }
+          `
+        }
+      />
```
Atmosfearful, line 48: All we need is `overflow: hidden`, which also solves the issue of losing your scroll position.

But the downside with that is it hides the scrollbar, which causes a tiny layout shift. No way around it.

lastly, nitpick, I would just move the conditional around the entire tag so we don't render it unless its being used:

```suggestion
    {isToggled && (
      <Global
        styles={css`
            body {
              overflow-y: hidden;
            }
          `
        }
      />)}
```

site/components/Navigation/NavItemDesktop.tsx
```diff
@@ -23,7 +23,7 @@ export const NavItemDesktop: FC<Props> = ({
         <Link href={url}>
           <a
             className={cx(styles.navMain_DesktopLink, {
-              dropdown: subMenu !== undefined,
+              dropdown: !!subMenu,
```
Atmosfearful, line 26: these url links don't have a dropdown/submenu, so I assume this and `data-active` can be removed

site/components/Navigation/NavItemDesktop.tsx
```diff
@@ -23,7 +23,7 @@ export const NavItemDesktop: FC<Props> = ({
...
             })}
             data-active={active?.toString()}
             {...props}
```
Atmosfearful, line 29: ah see this is a bit weird now because the `props` are typed to inherit `HTML Anchor` props but the anchor tag isn't always rendered. Also per a recent discussion in one of ladytrekkers recent PRs, I've come to dislike spreading `...props` because it can lead to bugs as its easy to accidentally pass down wrong props.

I suggest breaking this whole thing into two separate components, and then narrowing down the props / removing the unnecessary ones
* [ ] `LinkItemDesktop` which renders a nextjs `<Link>`
* [ ] `DropdownItemDesktop` which uses tippy and renders `props.children` in the dropdown
```
<DropdownItemDesktop>
  <LinkItemDesktop />
  <LinkItemDesktop />
</DropdownItemDesktop>
```
If it's too hard to get `LinkItemDesktop` to work as tippy child feel free to make a third component called `DropdownLink`
* [ ] Don't destructure props or spread `...props` down either, don't do `Props extends HTMLAttributes` just be explicit about which props it each one takes and pass them down as needed 🙏

Atmosfearful: Feel free to do this in a follow up PR btw, just make a ticket and tag yourself if you decide to do it separate

Atmosfearful (changes requested):

site/components/Header/HeaderMobile.tsx
goatsOnDrugs: Nice, overflow-y works much better.
We could customize the scroll bar for site the same way we did for app and give the body a padding-right of the scroll bar width to avoid the layout shift.

site/components/Navigation/NavItemDesktop.tsx
goatsOnDrugs: yes, you're right. removed.

site/components/Navigation/NavItemDesktop.tsx
goatsOnDrugs: Sweet, good ideas. Feels cleaner having them separate. TY

Atmosfearful (approved):

PR #667, opened 2022-08-31
pledge: Gnosis safe multisig support

### LINKED ISSUE #629: 500 Error on Pledge Creation
image

Unclear what the underlying issue is, but it seems to be related to the Vercel Node backend since it's a generic 500 error on the `/api/pledge` endpoint

We have the user available in Discord for a troubleshooting session

0xAeterno:
## Description

Support gnosis safe smart contract wallets on pledge dashboard.
Also adds:
- Form submitting state.

### Issue
Gnosis safe is a smart contract based wallet with multisig support. It utilises a different mechanism to verify a user/s signed signature and therefore affects user's ability to create and update pledges attached to a gnosis safe.

### How did we add support?
Users are prompted via gnosis safe app to sign a message when they confirm to manage their pledge.
This requires a transaction to be submitted in order to update the smart contract wallets (potentially multisig) signedMessages hash map. Signed messages can be verified by calling `isValidMessage` on gnosis safes smart contract.
However, there are two issues with this approach. 
- Each time a gnosis safe based wallet manages their pledge, it will require two transactions. One to sign the message and one to call `isValidMessage`. This results in unnecessary UX friction and gas usage to interact with our app.
- `isValidMessage` also requires a signer which is only available client side. This detracts from our pattern of verifying signatures server side.

The implemented solution is listen for a [`SignMsg`](https://github.com/safe-global/safe-contracts/blob/ee92957307653ae6cf7312bbcb1a13c6884ea6ea/contracts/libraries/SignMessageLib.sol#L23) event with the hashed signature as a param, which is emitted on when the signature is signed.
```
function signMessage(bytes calldata _data) external {
        bytes32 msgHash = getMessageHash(_data);
        signedMessages[msgHash] = 1;
        emit SignMsg(msgHash);
    }
```
This allows us to verify the signature server side by listening to the event and does not require additional transactions submitted to the blockchain. If the event is not emitted by the gnosis safes smart contract in 10 minutes, then the pledge will not be updated. Since it might take some time for the new blocks to be mined, confirming the signature, we've added a form submitting state to represent the request currently in flight, and to prevent further submissions until the current request has completed.

Questions:
- Is ten minutes enough for multi sig gnosis safes to sign a message?

## Related Ticket

Resolves #629 

## Changes

| Before form submission  | During form submission      |
|---------|---------|
|image|image|

## Checklist


ShimonD-KlimaDAO:
I tried creating a pledge first, and it worked flawlessly.
I then tried to edit the same page, and the page is stuck on "SUBMITTING", yet I don't see any signature requests in my gnosis safe.

Here is the transaction hash that went through successfully for the original pledge creation:
0x3b1b0c2f491553fae3850ad64c369c713b7e037d527213ce0454e91a36816978

image
image

site/components/pages/Pledge/lib/verifyGnosisSafeMultisig.ts
```diff
@@ -0,0 +1,32 @@
+import { ethers } from "ethers";
+import { getJsonRpcProvider } from "@klimadao/lib/utils";
+import GnosisSafeSignMessageLib from "@klimadao/lib/abi/GnosisSafeSignMessageLib.json";
+
+const ONE_HOUR = 60 * 60 * 1000;
```
0xAeterno, line 5: We can bump the wait time higher as needed.

ShimonD-KlimaDAO:
Not submitting a new issue, since I think the following two are issues that only arise due to:
(1) Something not yet deployed to prod which checks for those types of pledges.
(2) Local caching of the pledge while the graph, etc, update. 
If that's the case, everything looks good on my end!


1. I don't see my original pledge when I go to the production link. Why are those not the same?

What I submitted in the past: https://klimadao-site-git-pledges-gnosis-safe-support-klimadao.vercel.app/pledge/0x2703ab6ECF88B320aB6B30f8a8AAf07C2967bf5d
What I see now: https://www.klimadao.finance/pledge/0x2703ab6ECF88B320aB6B30f8a8AAf07C2967bf5d

2. I tried editing my pledge. It worked well. However, after I edited I tried disconnecting the gnosis safe wallet, and I immediately saw that my edits disappeared.

Before:
image
After: 
image

0xAeterno:
@ShimonD-KlimaDAO appreciate you taking the time to QA this!

1) Data across prod and non-prod (staging, preview builds) environments have separate databases :)

2) Yup you're correct in saying that data on pages are cached (if you visit the your pledge now you'll see your update). The app will refresh the data approximately every ~5 mins. I'm currently working on a task that will force an update on the cached page, when there is a successful update to the pledge.

site/components/pages/Pledge/lib/verifyGnosisSafeMultisig.ts
```diff
@@ -0,0 +1,32 @@
...
+  const signedEvent = gnosisSafeContract.filters.SignMsg(getMessageHash);
+
+  const waitForSignedEvent = (): Promise<void> =>
+    new Promise((resolve, reject) => {
+      setTimeout(() => {
+        gnosisSafeContract.removeListener(signedEvent, () => resolve());
+        reject(new Error("Gnosis safe signature verification timed out"));
+      }, ONE_HOUR);
+
+      gnosisSafeContract.once(signedEvent, () => resolve());
```
Atmosfearful, line 28: does this only succeed on future events? so we aren't able to do a flow like this:
1. wait clientside for event
2. pass signature to server
3. look back at recent events and see if it exists

?

site/components/pages/Pledge/lib/verifyGnosisSafeMultisig.ts
```diff
@@ -0,0 +1,32 @@
...
+  const messageHash = ethers.utils.hashMessage(params.signature);
+  const getMessageHash = await gnosisSafeContract.getMessageHash(messageHash);
+  const signedEvent = gnosisSafeContract.filters.SignMsg(getMessageHash);
+
+  const waitForSignedEvent = (): Promise<void> =>
+    new Promise((resolve, reject) => {
+      setTimeout(() => {
+        gnosisSafeContract.removeListener(signedEvent, () => resolve());
+        reject(new Error("Gnosis safe signature verification timed out"));
+      }, ONE_HOUR);
```
Atmosfearful, line 26: Since this is waiting for signatures server-side, the lambda should time out at 60 seconds

image

site/components/pages/Pledge/lib/verifySignature.ts
```diff
@@ -7,9 +7,13 @@ interface Params {
   nonce: string;
 }
 
-export const verifySignature = (params: Params): void => {
+export const verifySignature = async (params: Params): Promise<void> => {
+  // Gnosis multisig wallets are already validated client side
+  if (params.signature === "0x") return;
```
Atmosfearful, line 12: Oh I see, in my comment above I assumed you are using the method server side, but you are not.
Unfortunately we need to.
Our API endpoint is public, since we aren't using session tokens or api keys or anything like that. The signature verification is our access token.

The way this is written, anyone can send `{ signature: "0x", address: marcCubansAddress }` and get full write permission

Atmosfearful (changes requested):

site/components/pages/Pledge/lib/verifyGnosisSafeMultisig.ts
```diff
@@ -0,0 +1,32 @@
+import { ethers } from "ethers";
+import { getJsonRpcProvider } from "@klimadao/lib/utils";
+import GnosisSafeSignMessageLib from "@klimadao/lib/abi/GnosisSafeSignMessageLib.json";
+
+const ONE_HOUR = 60 * 60 * 1000;
+
+export const verifyGnosisSafeMultisig = async (params: {
+  signature: string;
```
Atmosfearful, line 8: I realized the naming here can be confusing. This is not actually a signature, it's what the ethers docs call a "message" (unsigned string)

Atmosfearful (changes requested):

site/components/pages/Pledge/lib/verifySignature.ts
```diff
@@ -1,19 +1,42 @@
...
     params.signature
   );
 
   if (decodedAddress.toLowerCase() !== params.address.toLowerCase()) {
     throw new Error("Invalid signature");
   }
 };
+
+export const verifySignature = async (params: Params): Promise<void> => {
+  // Expected signature generated server-side
```
0xAeterno, line 26:
```suggestion
```

site/components/pages/Pledge/lib/verifyGnosisSafeMultisig.ts
```diff
@@ -0,0 +1,60 @@
...
+  address: string;
+}) => {
+  const provider = getJsonRpcProvider();
+  const gnosisSafeContract = new ethers.Contract(
+    params.address,
+    GnosisSafeSignMessageLib.abi,
+    provider
+  );
+  const messageHash = ethers.utils.hashMessage(params.message);
+  const getMessageHash = await gnosisSafeContract.getMessageHash(messageHash);
```
Atmosfearful, line 47: RPC can be slow and flaky, so we should calculate this inline using ethers rather than calling the contract

Atmosfearful (changes requested):
lgtm just one change

When that is in we can call on @ShimonD-KlimaDAO to test again, and finally get this on prod!

site/components/pages/Pledge/lib/verifyGnosisSafeMultisig.ts
0xAeterno: Ah gotcha, so that's why code was pulled in from the other gnosis' repo to leverage ethers.
Makes sense :)

Atmosfearful (approved):

ShimonD-KlimaDAO:
When trying to edit the pledge, I received a confirmed successful transaction, but the Pledge page is stuck on "Submitting".

TX https://polygonscan.com/tx/0x345d52b3e8d738941f4ee58d8d0d2c0d35ad271b125792b55a59ccfe6c5bee34

image
:

site/components/pages/Pledge/lib/verifyGnosisSafeMultisig.ts
```diff
@@ -0,0 +1,81 @@
...
+    params.message
+  );
+
+  const signedEvent = gnosisSafeContract.filters.SignMsg(getMessageHash);
+  // signature event must be in the last 5 blocks (10 seconds)
+  const currentBlock = await provider.getBlockNumber();
+  const events = await gnosisSafeContract.queryFilter(
+    signedEvent,
+    currentBlock - 5
+  );
```
0xAeterno, line 76:
```suggestion
  const events = await gnosisSafeContract.queryFilter(
    signedEvent,
    -10
  );
```

Bumped to look at the last 10 blocks.
https://docs.ethers.io/v5/getting-started/#getting-started--history

0xAeterno:
Can we give this another test run @ShimonD-KlimaDAO?
I was able to successfully create and update a pledge with a test gnosis safe.

I've tweaked the server verification step to look for the signed message over a greater range of blocks (20seconds/10 blocks)

ShimonD-KlimaDAO:
I tried a few times, and it mostly went through smoothly. I tried to execute the transaction fast and slow. The second time I tried I waited only a few seconds, and I got the following. Also pasting in screenshot of successful transaction. After refreshing the pledge, I didn't see my update. I assumed that my update would appear anyway regardless of the error, since the transaction was successful.

image
image

ShimonD-KlimaDAO:
@0xAeterno @Atmosfearful

0xAeterno:
@ShimonD-KlimaDAO Could you document your testing approach for me to run? 🙏

Atmosfearful:
@0xAeterno I did a final test and pushed 4 commits. When you approve these I'll merge to prod!

https://github.com/KlimaDAO/klimadao/pull/667/commits/6b5101ff8d634d7a0ba8c1c1155a4c9baedfa737

Atmosfearful:
oops fatfingered the close button

PR #671, opened 2022-08-31
add decimals to staking rewards displayed

### LINKED ISSUE #661: Fix 0% rounding issue on main landing page
image

This is rounding to 0 places, but now that AKR is lower weekly rewards are below 1%

We should either round to more places (probably need 2) or do something more dynamic like rounding to more places if the rate is below a certain amount (e.g. round to 1 place between 1 and 0.1, round to 2 places between 0.1 and 0.01, etc.

goatsOnDrugs:
## Description
adds 2 decimal places to the staking rewards and change from weekly to monthly

I dont think so. smol change.

## Related Ticket

Resolves #661 

## Checklist

### DIFF
--- a/site/pages/index.tsx
+++ b/site/pages/index.tsx
@@ -15,7 +15,7 @@ export const getStaticProps: GetStaticProps<Props> = async (ctx) => {
   const latestPost = await fetchCMSContent("latestPost");
   const translation = await loadTranslation(ctx.locale);
   const blockRate = await fetchBlockRate();
-  const weeklyStakingRewards = await getStakingRewards({
+  const monthlyStakingRewards = await getStakingRewards({
     days: 31,
     blockRate,
     providerUrl: infuraURL,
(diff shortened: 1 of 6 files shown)


site/pages/index.tsx
```diff
@@ -16,7 +16,7 @@ export const getStaticProps: GetStaticProps<Props> = async (ctx) => {
   const translation = await loadTranslation(ctx.locale);
   const blockRate = await fetchBlockRate();
   const weeklyStakingRewards = await getStakingRewards({
```
Atmosfearful, line 18:
```suggestion
  const monthlyStakingRewards = await getStakingRewards({
```

Atmosfearful (changes requested):
thanks, just a variable name nitpick:

site/pages/index.tsx
goatsOnDrugs: good catch, updated the translation as well.

Atmosfearful (approved):
ty ty

PR #689, opened 2022-09-08
Resources: Add filters and sortings queries and UI elements

ladytrekker:
## Description

This PR adds a filterable and sortable list of CMS documents.

~Please note that the UI works **ON DESKTOP ONLY**.~
~The UI is not optimized to smaller screens yet. See the open to-dos below.~=> ALL DONE ✅ 

## PREVIEW LINK

https://preview-cms.vercel.app/resources

## TODOs

- [x] make all checkboxes unchecked on initial state
- [x] after a text search: set all checkboxes to initial unchecked state again (always search in all resources)
- [x] Fix: Reset to initial documents when all checkboxes are unchecked again
- [x] Fix: SortBy "publishedAt desc" and "publishedAt asc" are not working correctly, sorting is always chaotic
- [x] Search: If search term => filter and sort inside search results (extend query "searchByText") => No
- [x] Refactor: combine all queries into one (make queries understand `undefined` params)
- [x] Limit the documents results to max 20 items and add pagination UI elements and queries (see the [docs](https://www.sanity.io/docs/paginating-with-groq)) => no priority
- [x] Fix: better styles for SortByDropDown
- [x] Show Loading Spinner during requests
- [x] Mobile view for SortByDropDown and FilterCheckboxes (wrap all in another Dropdown again, see designs)
- [x] Mobile: Hide current Filter Checkbox SideBar on small screens
- [x] Fix: better styles for Search Bar (make bigger on bigger screens)
- [x] Follow-Up: pass ID as separated prop to Input elements (see [comment](https://github.com/KlimaDAO/klimadao/pull/689#discussion_r967087535) below) => [PR](https://github.com/KlimaDAO/klimadao/pull/737)
- [x] Follow-Up: remove all ForwardRef from Form inputs in `site/components/Form` (see [comment](https://github.com/KlimaDAO/klimadao/pull/689#discussion_r966225134) below) => will keep the ref for now, see comments below
- [ ] CrossBrowser Testing 
- - [x] Chrome Mac
- - [x] FF Mac
- - [x] Safari Mac
- - [x] iOS iPad
- - [x] iOS iPhone SE

## Related Ticket

Main part of ticket https://github.com/KlimaDAO/klimadao/issues/534

## Checklist


site/components/Form/Checkbox/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+      {
+        [styles.errorStyles]: !!props.errorMessage,
+      },
+      props.inputProps.className
+    );
+
+    return (
+      <div className={styles.checkboxContainer}>
+        <input
+          id={props.inputProps.id}
```
Atmosfearful, line 27: is this prop required? what happens if you don't pass an id? I don't think typescript would enforce it as required the way this is right now

site/components/Form/Checkbox/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+import * as styles from "./styles";
+
+interface Props {
+  inputProps: InputHTMLAttributes<HTMLInputElement>;
+  label: string;
+  errorMessage?: string;
+}
+
+export const Checkbox = React.forwardRef<HTMLInputElement, Props>(
```
Atmosfearful, line 14: remind me, why do we need forwardRef for these?

site/components/Form/Checkbox/styles.ts
```diff
@@ -0,0 +1,93 @@
...
+`;
+
+export const baseStyles = css`
+  ${typography.caption};
+  display: grid;
+  grid-template-columns: 1em auto;
+
+  /* Add if not using autoprefixer */
+  -webkit-appearance: none;
+  /* Remove most all native input styles */
```
Atmosfearful, line 21: thanks for the 'why' comment, its helpful 👍

Atmosfearful (commented):
Some things to note that I don't see in your pr description. Perhaps these are issues with react-hook-form?

* [ ] checking a single category, then unchecking it -> should re-query but right now it doesn't
* [ ] The designs had everything checked by default, and I think that makes sense. But it requires the UX to be flipped a little bit from what you have now. The buttons would actually have two behaviors.

I try to illustrate what I mean:

User journey 1: page load -> everything is selected by default -> click a **label** to select that category and unselect everything else

User journey 2:  page load -> everything is selected by default -> click a *checkbox* to *unselect* that category and everything else stays selected

ladytrekker:
> * [ ]  The designs had everything checked by default, and I think that makes sense. But it requires the UX to be flipped a little bit from what you have now. The buttons would actually have two behaviors.
> 
> I try to illustrate what I mean:
> 
> User journey 1: page load -> everything is selected by default -> click a **label** to select that category and unselect everything else
> 
> User journey 2: page load -> everything is selected by default -> click a _checkbox_ to _unselect_ that category and everything else stays selected

Let's move these kind of concept discussions to the ticket. 
This needs more brain food then just talking about a code review.
Please see my comment here => https://github.com/KlimaDAO/klimadao/issues/534#issuecomment-1241620148

site/components/Form/Checkbox/index.tsx
ladytrekker: Hm, I followed the implementation of the other Pledge Inputs.
I think, these are not needed at all indeed.
Another follow-up refactoring PR? I added this to the list above.

ladytrekker: https://react-hook-form.com/faqs#Howtosharerefusage

0xAeterno: Here's another example from their documentation for further context.
At its core, we need to pass a ref onto the input elements so the library can work its magic.
It's up to us to agree on the pattern on how to do that moving forward.
https://react-hook-form.com/get-started#Integratinganexistingform

I think something to keep in mind here is that we're creating reusable form input elements to be used across the app. So imo, the example in the link above doesn't feel like the most scalable solution, but also the example is an extension on refs if we need to use refs for other purposes (eg. scroll into view)

site/components/pages/Resources/ResourcesList/styles.ts
```diff
@@ -46,3 +49,93 @@ export const searchInputSubmit = css`
...
+
+  ${breakpoints.large} {
+    display: flex;
+    flex-direction: row;
+    gap: 1.2rem;
+    align-items: center;
+  }
+`;
+
+export const sortByButtons = css`
```
Atmosfearful, line 94: on mobile these need justify-content: start and gap: 1.6rem

before fix:
image


after fix:

image

Atmosfearful (changes requested):

site/components/pages/Resources/ResourcesList/index.tsx
```diff
@@ -17,27 +33,66 @@ export interface Props {
...
+  sortedBy: "",
 };
 
 export const ResourcesList: FC<Props> = (props) => {
   const [visibleDocuments, setVisibleDocuments] = useState<Document[] | null>(
     props.documents
   );
+  const [isLoading, setIsLoading] = useState(false);
+  const [isError, setIsError] = useState(false);
+  const [showMobileMobile, setShowMobileModal] = useState(false);
```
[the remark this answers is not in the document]
ladytrekker: Done

site/components/pages/Resources/ResourcesFilters/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+        ))}
+      </div>
+      <Text
+        t="caption"
+        align="center"
+        role="button"
+        onClick={props.onResetFilters}
+        style={{ cursor: "pointer" }}
+      >
+        X <Trans id="resources.form.filters.clear_all">Clear all filters</Trans>
```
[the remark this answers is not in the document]
ladytrekker: Done

site/components/pages/Resources/ResourcesList/styles.ts
ladytrekker: Done

Atmosfearful (approved):

PR #691, opened 2022-09-08
useFocusTrap & fix button outlines while tabbing

### LINKED ISSUE #684: Modal fixes & useFocusTrap
Forgot to test a11y when @ladytrekker made the modal. I think I found some issues:
* [ ] Close button isn't tabbable. Not sure why not. Maybe the focus outline is just hidden?
* [ ] Focus trap. Make a hook called `useFocusTrap()`.
This should use a [callback ref](https://reactjs.org/docs/refs-and-the-dom.html#callback-refs) which, if the element is rendered, parses the html to find focusable elements, then sets a blur event listener on the container to identify when the last element in the modal has been blurred.
```jsx
const ModalDemo = (props) => {
  const focusTrapRef = useFocusTrap();
  
  if (!props.showModal) return null;
  return (
  <div className="modalContainer" ref={focusTrapRef}>
     <button className='close' />
     {props.children}
  </div>
)
```
There are probably some edge cases to handle for the focus trap that I'm not remembering :)
here is some sample code for how to query-select focusable elements. Instead of using the container id, you would use the ref https://uxdesign.cc/how-to-trap-focus-inside-modal-to-make-it-ada-compliant-6a50f9a70700

* [ ] To improve tab navigation, we might need to render modals immediately below the button that opens them. That way when the user clicks (focuses) the button, and presses tab, the first item in the modal is focused. Another option is to force-focus the modal when it opens. Which could be accomplished with `useFocusTrap` as well
```ts
<button onClick={toggleModal} />
<modal />
```

Atmosfearful:
## Description

Along the way I discovered that outlines werent working on any button, because we never brought over the `user-is-tabbing` logic from the `app`

useFocusTrap ended up being too painful to use on mobile menu, because we can only trap focus there when its open. But I think that's OK because I can still tab through the menu pretty easily.

Modals seem to be working great!

## Related Ticket
Resolves #684

### DIFF
--- a/app/components/views/Home/index.tsx
+++ b/app/components/views/Home/index.tsx
@@ -122,30 +122,6 @@ export const Home: FC = () => {
     }
   }, [web3.isConnected]);
 
-  /**
-   * Outline manager for a11y
-   * So we can hide outlines when clicking, only show them when tabbing
-   */
-  useEffect(() => {
-    const handleMousedown = () => {
-      document.body.removeEventListener("mousedown", handleMousedown);
-      document.body.classList.remove("user-is-tabbing");
-      document.body.addEventListener("keydown", handleKeydown);
-    };
-    const handleKeydown = (e: KeyboardEvent) => {
-      if (e.keyCode === 9) {
-        document.body.removeEventListener("keydown", handleKeydown);
-        document.body.classList.add("user-is-tabbing");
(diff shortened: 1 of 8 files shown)


goatsOnDrugs (approved):
Neat!

PR #708, opened 2022-09-19
pledge: on-demand revalidation

### LINKED ISSUE #466: [Pledge] Trigger on-demand revalidation when pledge is updated.
https://vercel.com/docs/concepts/next.js/incremental-static-regeneration#using-on-demand-revalidation

From within out nextjs API endpoint `/api/pledge`, we can ping this webhook. Some setup required but it should be a straightforward task.

```
    await res.unstable_revalidate('/path-to-revalidate');
```

# Question

Can we do this from within the endpoint directly, or do we need to set up a second `api/revalidate` endpoint as per the doc examples?

0xAeterno:
## Description

Implements on-demand revalidation users successfully updates their pledge.

#### What does this solve?
Pages are cached globally and are regenerated at an interval configured in our app (for `/pledges/[address].tsx` it's approximately 4 mins).
This means an out of date page can be served until the page is regenerated.

## Related Ticket

Resolves #466 

### DIFF
--- a/site/components/pages/Pledge/PledgeForm/index.tsx
+++ b/site/components/pages/Pledge/PledgeForm/index.tsx
@@ -59,7 +59,7 @@ type Props = {
 
 export const PledgeForm: FC<Props> = (props) => {
   const [serverError, setServerError] = useState(false);
-  const { asPath } = useRouter();
+  const { pathname } = useRouter();
   const { signer } = useWeb3();
   const { control, register, handleSubmit, formState, reset, setValue } =
     useForm<PledgeFormValues>({
@@ -87,7 +87,7 @@ export const PledgeForm: FC<Props> = (props) => {
         pageAddress: props.pageAddress,
         pledge: values,
         signature,
-        urlPath: asPath,
+        urlPath: pathname,
       });
       const data = await response.json();
 
--- a/site/components/pages/Pledge/PledgeForm/index.tsx
+++ b/site/components/pages/Pledge/PledgeForm/index.tsx
@@ -59,7 +59,7 @@ type Props = {
 
 export const PledgeForm: FC<Props> = (props) => {
   const [serverError, setServerError] = useState(false);
-  const { pathname } = useRouter();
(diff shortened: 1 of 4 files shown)


jabby09:
@ShimonD-KlimaDAO this is ready for QA

site/pages/api/pledge/index.ts
```diff
@@ -26,6 +26,7 @@ export default async function handler(
           );
         }
 
+        await res.unstable_revalidate(req.body.urlPath);
```
Atmosfearful, line 29: nice, this is rad. We will have to circle back after we address #714 because it was made stable in 12.2 🎉

site/components/pages/Pledge/PledgeForm/index.tsx
```diff
@@ -86,6 +87,7 @@ export const PledgeForm: FC<Props> = (props) => {
         pageAddress: props.pageAddress,
         pledge: values,
         signature,
+        urlPath: asPath,
```
Atmosfearful, line 90: Did you test paths with query params or hash routes? `asPath` includes both, might be safer to use `pathname`

But on second thought we don't even need to pass the URL, it's always `/pledge/${pageAddress}` right?
Is `pageAddress` the kns/ens domain or is it always the 0x address?

Atmosfearful (changes requested):

site/pages/api/pledge/index.ts
0xAeterno: Sweet, I noticed that too because `res.revalidate` wouldn't work 😆

site/components/pages/Pledge/PledgeForm/index.tsx
0xAeterno: Update: `pathname` gives us `/pledge/[address]`. So I went with constructing the url path using `query`

Atmosfearful (approved):

ShimonD-KlimaDAO:
Works great!

PR #729, opened 2022-09-29
Goats/658 multi wallet pledge

### LINKED ISSUE #658: Multi-wallet pledges
User's need the ability to associate several wallets to a pledge.

## Push vs Pull pattern
These have subtle engineering and UX implications, but it is worth discussing. On our original call we leaned towards "push" but I've decided that "Pull" is a better pattern.

"Push" pattern:
  1. Secondary wallet owner visits the pledge, connects and "pushes" their address onto it
  2. The pledge owner wallet accepts or denies any pushed addresses
  
"Pull" pattern:
  1. Pledge owner enters the secondary addresses they would like to "pull" in.
  2. The pulled wallets then connects and visits the pledge, and are prompted to accept or deny the pull.
  
 I believe that **Pull** is **better** for three reasons:
 1. With the Push pattern, we would have to prompt **every** connected user who visits *any* pledge: 'would you like to pin your address to this pledge?'. This is irrelevant for 99.9% of wallets. Whereas with the Pull pattern, **only pulled wallets see the prompt 👌**.
 2. With the Push pattern, I can write a few lines of code that spams a pledge with randomly created addresses.
 3. It's somewhat easier (engineering and UX wise) to enforce that a wallet not be Pulled into to multiple pledges, it's a minor tweak to the Edit Pledge API. Pushing would require a separate API.
  
 # Core UX Requirements
 * [x] Impermanence: the pledge owner has the power to unpin any wallet at any time
 * [x] Secondary wallets also reserve the power to unpin themselves at any time
 * [x] When accepting the "Pull" into another pledge, we must prompt users that any existing pledges under their address will be redirected to the new pledge.
 
  ## Other engineering requirements
* [x] If someone creates a pledge, then later decides to pin themselves to a different pledge, we need to update existing pledge documents in the db to record this state.
 ```json
 {
   "pinned_to": {
      "pledge_id": "[database_id_of_pledge]",
      "pledge_owner_address": "[0x_address]",
    }
 }
 ```

goatsOnDrugs:
## Description
This is the base branch I will be making PRs to for the multi-wallet pledges. Error handling and messages will be handled in a follow up PR. 
Overview: This PR adds the functionality to pin "secondary wallet address" to a pledge so the retirements count towards that pledge. So the flows are 
1. pledge owner adds or removes a wallet in the pledge form modal. 
2. invited wallet visits pledge page and sees the invitation modal. user rejects and sets status to "rejected" or accepts and sets status to "verified".
3. Invited user removes themself from the pledge via modal  (pledge dashboard -> click "edit pledge")

Test cases:
- [ ] pledge owner can add valid wallet and submit
- [ ] user sees error message for invalid wallet address
- [ ] user sees error for duplicate wallet on submit
- [ ] user should only be able to be adding one wallet at a time ("add wallet" button is disabled when adding a wallet)
- [ ] user should not see rejected wallets in list but be able to re-add them
- [ ] if you are pinned to a pledge when a user visits pledge/youraddress then they should be redirected to the parent pledge
- [ ] user should be able to accept invitation
- [ ] if a user is already pinned to another pledge they should see an error
- [ ] user should be able to remove themselves from pledge
- [ ] retirements should be added up from all wallets on parent pledge
- [ ] on retirements dropdown user should only see verified wallets and owner. if user owns pledge they should see pending wallets

In order to test this you will need to create a pledge and invite your secondary wallet. QA will also need access to firebase or we can do a screen share. We would need to manually set the status back from "rejected" if you need to test again after rejecting

## Related Ticket

Resolves #658 

## Checklist


site/components/pages/Pledge/lib/firebase.ts
```diff
@@ -53,21 +58,99 @@ export const findOrCreatePledge = async (
...
+      .where(
+        `wallets.${params.secondaryWalletAddress}.status`,
+        "==",
+        "verified"
+      )
+      .get();
+    // if its already add it then un add from other pledge?
+    const isNotAlreadyAdded = snapshot.empty;
+    // check if secondary wallet is signing to accept
+    if (invitedAddress && isNotAlreadyAdded && params.secondaryWalletAddress) {
```
0xAeterno, line 86: All this branching logic is pretty hard to follow :(
It feels like this function is responsible many actions, is there a reason why we opted to add onto the existing endpoint instead of creating an each endpoint for handling adding and removing secondary wallets?

0xAeterno (commented):
Great work so far @goatsOnDrugs.
This definitely is not an easy ticket to tackle.
This is my first pass at reviewing the PR, it's a sure is a big one.

To help with gaining context for reviewing. Would we be able to get a summary of the features at a high level?

0xAeterno:
@goatsOnDrugs and I had a meeting on this feature, we've identified and addressed small improvements to the PR.
Happy enough to move this ticket along towards release and iterate on on it further post release.
Some improvements to explore are documented in issue #773

@Atmosfearful what do you think? Feel free to add to the issue larger improvements to be addressed

site/components/pages/Pledge/InvitationModals/AcceptModal/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+        pledge: pledgeFormAdapter(props.pledge),
+        signature,
+        urlPath: `/pledge/${props.pageAddress}`,
+      });
+      props.setShowAcceptModal(false);
+    } catch (e: any) {
+      console.log("error:", e);
+      setStep("error");
+      setErrorMessage(
+        e.message ??
```
Atmosfearful, line 72: do we translate the message thrown by `putPledge` or the server? (I didn't get to that part of the code yet)

site/components/pages/Pledge/InvitationModals/RemoveModal/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+        id: "shared.loading",
+        message: "Loading",
+      }),
+    }[step]);
+
+  const { signer } = useWeb3();
+  const handleSubmit = async (params: { message: string }) => {
+    try {
+      if (!signer) return;
+      const address = await signer.getAddress();
```
Atmosfearful, line 51: same here, we have this already from useWeb3

site/components/pages/Pledge/InvitationModals/RemoveModal/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+      const signature = await signer.signMessage(params.message);
```
Atmosfearful, line 52: same here, need to support gnosis

site/components/pages/Pledge/PledgeForm/index.tsx
```diff
@@ -56,26 +66,51 @@ type Props = {
...
 
 export const PledgeForm: FC<Props> = (props) => {
   const [serverError, setServerError] = useState(false);
+  const [selectedAddress, setSelectedAddress] =
+    useState<{ address: string; index: number }>();
   const [submitting, setSubmitting] = useState(false);
   const { query } = useRouter();
   const { signer } = useWeb3();
+  // any type is used here because the pledge will only be this shape once before we format the wallets so it is not a type of Pledge or PledgeFormValues
+  const formattedPledge: any = { ...props.pledge };
```
Atmosfearful, line 81: note to self: this is confusing i need to re read this after coffee tmr

Atmosfearful (changes requested):
Niiice looking good so far. I didn't make it all the way through. But these are things you can jump on while I finish the rest tmr
* [x] a .prettierrc is still being committed even though its in gitignore, need to delete that.

* [ ] major blocker is we should support gnosis before launch. Shouldn't be too hard to support.

* [ ] We probably also need to add a helpful message while waiting for signers, like I did [here](https://github.com/KlimaDAO/klimadao/pull/667#issuecomment-1291180072), because It's not apparent to newbies (gnosis or not) hat they need to "sign" their edit.

* [ ] we need better validation (ethers.utils.isAddress()) for the provided wallets 
image

* [ ] minor layout cleanups: the grid parent for this input is funking up because there are 4 elements as child. Maybe wrap them in another div. If you can center that green floppydisk on the input that would be rad too (maybe align-self: start w/ some padding-top? idk)
image

* [x] prevent this awkward irreversible situation by disabling the "add wallet" button until they have clicked `save` on the previous entry 
image

goatsOnDrugs:
@Atmosfearful for the validation I am having trouble figuring out how to do custom validation with a function. We can check it against a regex but I dont see a way to pass yup a custom function. We should be able to pass rules to fieldArray but it is not working and I'm not sure why. (```useFieldArray({
  rules: { minLength: 4 }
})```
For the formatting issue that is inside of TextInput. Should we create a separate PR for that fix?

0xAeterno:
@goatsOnDrugs just pushed a PR (https://github.com/KlimaDAO/klimadao/pull/780) to help out with address validation on a string.
Hope this helps out!

site/components/pages/Pledge/PledgeForm/index.tsx
```diff
@@ -111,13 +112,34 @@ export const PledgeForm: FC<Props> = (props) => {
...
+          wallets[i].address === wallet.address &&
+          wallet.status === "pending" &&
+          wallets[i].status === "pending"
+        ) {
+          isDuplicate = true;
+        }
+      }
+    });
+    return isDuplicate;
+  };
```
0xAeterno, line 129: Does this need to check for duplication of wallet address _and_ status?
Checking for duplicate wallet address seems sufficient to me unless I'm missing something.

Would you find it useful to extend the yup schema similar to the other PR. Something like `yup.array().unique(attribute)?` I can give it a crack.

site/components/pages/Pledge/InvitationModals/RemoveModal/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+  const { signer, address } = useWeb3();
+  const handleSubmit = async (params: { message: string }) => {
+    try {
+      if (!signer) return;
+      setStep("signing");
+      const signature = await signer.signMessage(params.message);
+      setStep("loading");
+      if (signature === "0x") {
+        await waitForGnosisSignature({
+          message: editPledgeMessage(props.pledge.nonce),
```
Atmosfearful, line 67: I think this should be `params.message`

Atmosfearful (changes requested):

site/components/pages/Pledge/lib/firebase.ts
```diff
@@ -34,7 +53,11 @@ export const getPledgeByAddress = async (address: string): Promise<Pledge> => {
   const pledgeRef = pledgeSnapshot.docs[0];
 
   if (!pledgeRef) {
-    throw new Error("Bad request");
+    const e = new Error(
+      `We were unable to find a pledge for address: ${address}`
+    );
+    e.name = "PledgeNotFound";
```
Atmosfearful, line 59: Something I'm realizing right now, we should document these, preferably with typescript. This can be done with:
1. a union type of all the names
```ts
type NamedError: "PledgeNotFound" | "WalletAlreadyPinned" | ...
```
2. use that union type in the expected return value for the API endpoint here https://github.com/KlimaDAO/klimadao/pull/729/files#diff-c9e7c84c661df8d858eda48e157a65da2c0216d9aac18a242face4cb77162965R6

more type safe, always better imo, but this may have waterfall effect of type errors, so if it causes issues we can do it in a follow-up tech debt ticket. In that case just make sure to document all of the known Named errors in a github ticket and link it here in this comment 👇

Atmosfearful (changes requested):

site/components/pages/Pledge/PledgeForm/index.tsx
```diff
@@ -126,217 +250,400 @@ export const PledgeForm: FC<Props> = (props) => {
...
+              </Text>
+              <Tippy
+                content={
+                  <div className={styles.pledge_form_wallet_info_content}>
+                    <Text t="caption">
+                      <Trans id="pledge.form.wallets.tooltip">
+                        Add other wallets to contribute to your pledge. If the
+                        owner of the wallet accepts your invitation, their
+                        wallet will appear on your pledge dashboard. Any
+                        retirments made by a secondary wallet will contribute
```
Atmosfearful, line 336:
```suggestion
                        retirements made by a secondary wallet will contribute
```

site/pages/pledge/[address].tsx
```diff
@@ -83,6 +96,62 @@ export const getStaticProps: GetStaticProps<PageProps, Params> = async (
...
+      );
+      promises.push(
+        getRetirementTotalsAndBalances({
+          address: resolvedAddress,
+        })
+      );
+      const values: RetirementsTotalsAndBalances[] = await Promise.all(
+        promises
+      );
+      values.reduce((prev: any, curr) => {
```
Comment (Atmosfearful) [inline site/pages/pledge/[address].tsx:118]: this should return 1 reduced value at the end. We can also use a Generic type arg <ExpectedType> so it knows what to expect for the `prev` and `return` values

```ts
const sumValues = values.reduce<ExpectedType>((prev, curr) => {
   ...dostuff()
}, SumValuesType)
```

site/pages/pledge/[address].tsx
```diff
@@ -83,6 +96,62 @@ export const getStaticProps: GetStaticProps<PageProps, Params> = async (
...
+        if (prev) {
```
Comment (Atmosfearful) [inline site/pages/pledge/[address].tsx:119]: prev should always be present

site/pages/pledge/[address].tsx
```diff
@@ -83,6 +96,62 @@ export const getStaticProps: GetStaticProps<PageProps, Params> = async (
...
+          prev.mco2 = (Number(prev.mco2) + Number(curr.mco2)).toString();
+          prev.nct = (Number(prev.nct) + Number(curr.nct)).toString();
+          prev.ubo = (Number(prev.ubo) + Number(curr.ubo)).toString();
+          prev.nbo = (Number(prev.nbo) + Number(curr.nbo)).toString();
+
+          return prev;
+        } else {
+          return curr;
+        }
+      });
```
Comment (Atmosfearful) [inline site/pages/pledge/[address].tsx:140]: should give it an initial value of empty object `{}`, this will help typescript understand our intention too.

```ts
const const sumValues = values.reduce<SumValuesType>((prev, curr) => {
   ...dostuff();
}, {} as SumValuesType)
```

note that this changes the behavior of reduce: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce
```
If initialValue is specified, that also causes currentValue to be initialized to the first value in the array. If initialValue is not specified, accumulator is initialized to the first value in the array, and currentValue is initialized to the second value in the array.
```

site/pages/pledge/[address].tsx
```diff
@@ -83,6 +96,62 @@ export const getStaticProps: GetStaticProps<PageProps, Params> = async (
...
+        retirements = values[0];
+      } else {
+        retirements = values as unknown as RetirementsTotalsAndBalances;
+      }
+    } else {
+      retirements = await getRetirementTotalsAndBalances({
+        address: resolvedAddress,
+      });
+    }
+    // const retirements = await getRetirementTotalsAndBalances({
```
Comment (Atmosfearful) [inline site/pages/pledge/[address].tsx:151]: remove ded code plz

Atmosfearful (changes requested):

Atmosfearful (approved):

PR #733, opened 2022-09-29
working example

Atmosfearful:
demo


site/components/pages/Pledge/lib/verifyGnosisSafeMultisig.ts
```diff
@@ -30,3 +31,52 @@ export const verifyGnosisSafeMultisig = async (params: {
...
+  /** Address of the multisig contract */
+  address: string;
+}) => {
+  const provider = getJsonRpcProvider();
+  const Safe = new ethers.Contract(
+    params.address,
+    GnosisSafeSignMessageLib.abi,
+    provider
+  );
+  const safeMessageHash = calculateSafeMessageHash(
```
0xAeterno, line 70: Could this be simplified to calling `Safe.getMessageHash` (since we're already importing that function via the ABI).

Or is there a benefit to the approach here?

PR #728, opened 2022-09-29
Fixes hindi language menu button in production

biwano:
## Description

Fixes hindi language menu button in production

## Related Ticket

Resolves #727

## Changes

| Before  | After  |
|---------|--------|
|img here |img here|

-->

## Checklist

- [X] Building the site with `npm run build-site` works without errors
- [X] Building the app with `npm run build-app` works without errors
- [X] I formatted JS and TS files with running `npm run format-all`

PR #734, opened 2022-09-30
Single source of truth for locale information

biwano:
## Description

Single source of truth for locale information

If we want to have the Opportunity to have differences between app and site we can make the getLocales function parametrized.

## Related Ticket

Resolves #nothing
It is related to @Atmosfearful comment here: https://github.com/KlimaDAO/klimadao/pull/728#pullrequestreview-1125566392

## Changes

None

## Checklist

### DIFF
--- a/lib/constants/index.ts
+++ b/lib/constants/index.ts
@@ -254,3 +254,5 @@ export const verra = {
   projectSearch: `${VERRA_REGISTRY_API}/resource/resource/search?maxResults=2000&$count=true&$skip=0&$top=50`,
   projectDetailPage: `${VERRA_REGISTRY}/app/projectDetail/VCS`, // add ID after VCS like /191
 };
+export const IS_PRODUCTION =
+  process.env.NEXT_PUBLIC_VERCEL_ENV === "production";
--- a/lib/constants/index.ts
+++ b/lib/constants/index.ts
@@ -254,5 +254,3 @@ export const verra = {
   projectSearch: `${VERRA_REGISTRY_API}/resource/resource/search?maxResults=2000&$count=true&$skip=0&$top=50`,
   projectDetailPage: `${VERRA_REGISTRY}/app/projectDetail/VCS`, // add ID after VCS like /191
 };
-export const IS_PRODUCTION =
-  process.env.NEXT_PUBLIC_VERCEL_ENV === "production";
--- a/lib/utils/getLocales/index.ts
+++ b/lib/utils/getLocales/index.ts
@@ -25,7 +25,6 @@ const productionLocales: ILocales = {
   },
   es: { plurals: es, time: "es-ES", label: "Español" },
   hi: { plurals: hi, time: "hi-IN", label: "हिन्दी" },
-  ko: { plurals: ko, time: "ko-KR", label: "한국어" },
 };
 
 const stagingLocales: ILocales = Object.assign({}, productionLocales, {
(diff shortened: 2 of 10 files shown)


lib/utils/getLocales/index.ts
```diff
@@ -0,0 +1,48 @@
...
+      n: number | string,
+      ord?: boolean
+    ) => "zero" | "one" | "two" | "few" | "many" | "other";
+    time: string;
+    label: string;
+    production_ready: boolean;
+  };
+}
+
+const locales: ILocales = {
```
Atmosfearful, line 18: how about `productionLocales` and `stagingLocales` defined as separate objects? just makes it more clear at a glance

lib/utils/getLocales/index.ts
```diff
@@ -0,0 +1,48 @@
+import { en, fr, de, ru, zh, ko, es, hi } from "make-plural/plurals";
```
Atmosfearful, line 1: We need to add this (or lingui) as a dependency for @klimadao/lib

lib/constants/index.ts
```diff
@@ -254,3 +254,5 @@ export const verra = {
   projectSearch: `${VERRA_REGISTRY_API}/resource/resource/search?maxResults=2000&$count=true&$skip=0&$top=50`,
   projectDetailPage: `${VERRA_REGISTRY}/app/projectDetail/VCS`, // add ID after VCS like /191
 };
+export const IS_PRODUCTION =
+  process.env.NEXT_PUBLIC_VERCEL_ENV === "production";
```
Atmosfearful, line 258: I would prefer that `lib` stay completely agnostic about nextjs or vercel, so we will have to pass this flag down at runtime

Atmosfearful (changes requested):

ladytrekker (commented):
One thing is still missing I think:
https://github.com/KlimaDAO/klimadao/blob/staging/site/next.config.js#L92-L107
=> adding Hindi to the list of production languages

Also this should have been added to this fix here?
https://github.com/KlimaDAO/klimadao/pull/728/files

lib/utils/getLocales/index.ts
biwano: Indeed

lib/constants/index.ts
biwano: Okay

biwano:
> One thing is still missing I think: https://github.com/KlimaDAO/klimadao/blob/staging/site/next.config.js#L92-L107 => adding Hindi to the list of production languages
> 
> I assume this should have been added to this PR here too? #728

Yes.  Introduced the async function version of the configuration module to enable dynamic imports: https://nextjs.org/docs/api-reference/next.config.js/introduction

I will make some tests on production builds this week end.

biwano:
My production builds seemed okay.

lib/constants/index.ts
ladytrekker, line 258: I think this needs to be removed from lib still.

ladytrekker (approved):
Code and Preview Links look good to me.

See some minor thing below

lib/utils/getLocales/index.ts
```diff
@@ -0,0 +1,42 @@
...
+  de: { plurals: de, time: "de-DE", label: "Deutsch" },
+  ru: { plurals: ru, time: "ru-RU", label: "Русский" },
+  "zh-CN": {
+    plurals: zh,
+    time: "zh-CN",
+    label: "中文",
+  },
+  es: { plurals: es, time: "es-ES", label: "Español" },
+  hi: { plurals: hi, time: "hi-IN", label: "हिन्दी" },
+  ko: { plurals: ko, time: "ko-KR", label: "한국어" },
```
ladytrekker, line 28: This is not correct IMO, korean is still only for Staging, right?
Double check with https://www.klimadao.finance/
and https://app.klimadao.finance/#/stake

PR #746, opened 2022-10-18
fiat2retirement

Atmosfearful:
draft


Atmosfearful:
@0xAeterno is this goign to create a ton of conflicts with your selective retirement work? Shall I rebase on top of you?

0xAeterno:
@Atmosfearful potentially, but only the parts where we're rearranging the offset form.
The core selective retirement feature is a very self contained changed.

If you reakon this feature is going to be live first (most likely with the current blockers on the other ticket). I'll revert the changes form flow changes and rebase when merged and reapply.

jabby09:
**_Initial feedback...will add more as test through the weekend..._**

- [ ] When enforcing the 1 tonne minimum, once I type in the input box and it default to one I can never clear the input with backspace or delete. To change from 1, I have to use the incriminator control or select one and type from keyboard. Not 100% intuitive, should be able to clear the input box with normal keyboard functions.
- [ ] On the transaction complete page, the "Return to..." cta, maybe that should be https://app.klimadao.finance/#/offset vs just back to the homepage.
- [ ] Should we include a support email to Klima as well or is that just confusing or not necessary?

app/components/views/Offset/index.tsx
```diff
@@ -542,12 +639,9 @@ export const Offset = (props: Props) => {
                   !!beneficiaryAddress && !utils.isAddress(beneficiaryAddress)
                 }
                 onChange={(e) => handleBeneficiaryAddressChange(e.target.value)}
-                placeholder={t({
-                  id: "offset.enter_address",
-                  message: "Enter 0x address",
-                })}
+                placeholder={t`Beneficiary 0x address (optional)`}
```
0xAeterno, line 642: Not a blocker, just starting a conversation (:
Are we dropping translation ids?

0xAeterno (approved):
Nice one @Atmosfearful, nothing really sticks out as a blocker.
QA testing should cover everything else in the feature end to end.

app/components/views/Offset/index.tsx
Atmosfearful: Ah this was me experimenting but I forgot to revert it.
But i've been ruminating on the idea of slowly migrating away from ids, except for maybe common strings. The main benefit of ids are:
1. make it easier to reuse strings (in theory-- in practice we probably have many duplicates)
2. make it easier for translators to find them in the ui, if we are good about writing helpful ids `pledge.edit_modal.loading`

ShimonD-KlimaDAO:
Feedback: #775

app/components/views/Offset/index.tsx
ladytrekker: > Open question: if I remove a string id but keep the source string, does that trigger a re-translation?

I guess this would make a change in translation.io when an existing ID is removed.
And would be a new translation afterwards because internally lingui creates a new ID. 

See your own result in this PR [here](https://github.com/KlimaDAO/klimadao/pull/746/files#diff-c807d6bc34a846ae9029acc99e48d0bedf787ca01fe7e2194628976f5ba567d0R17-R23) 
=> a new ID from the message string was created

I would keep IDs for now.
Especially because detecting doubled translations is easier.

Issue #720, opened 2022-09-24
[tech] Reduce bundle size for _app file


Atmosfearful:
For the `site` repo (I haven't looked at `app` yet) the Nextjs _app bundle is growing quite dramatically. I have a few suspects in mind:
- lib/components is not properly tree shaking?
- every `abi` json is being included?
- some libraries we've added?

```
+ First Load JS shared by all                      1.01 MB
  ├ chunks/framework-81942bc1e06a8d22.js           42.1 kB
  ├ chunks/main-c76117333b9674d3.js                26.6 kB
  ├ chunks/pages/_app-17a9b104f534882a.js          937 kB  // 😵💀💀💀
  ├ chunks/webpack-d5ab764da11c7a9c.js             2.29 kB
  └ css/9ac0ea7000e013d4.css                       1.54 kB
```

biwano:
@Atmosfearful @ladytrekker  Hey, do we have some kind of history of builds. This way the culprit(s) could be easily found.

biwano:
@Atmosfearful By bisecting I found two principal culprits.
- 58eafeb7ca6d4184f4acdcf7470b83e98da05c7f (5 May) made the js bundle go from 142Kb to 643Kb. It is a big squashed commit for pledge. It could help if someone still had the original commit history but I think that by investigating the additions to package.json I can find what happened.

- 4e89a5cfe6dacf724521f7f9228faa81726ad68c (21 June) made the js bundle go from 728Kb to 986Kb. This one is also a big squashed commit for the new web3 provider. Maybe one of the providers has too many export... This commit impacted both app and site

I keep diving.

biwano:
![image](https://user-images.githubusercontent.com/11857992/196671629-e786a626-3845-4f83-9acb-34b971d1a570.png)

Clue for crpto-browserify: https://github.com/vercel/next.js/discussions/13631

biwano:
Some research notes:
- Created a hollow version of the app (1 page, no translations) : First load JS => 928Kb
- Removed dependency on constants => 834kB (-90Kb)
- Removed dependency on Torus => 547kB (-287Kb) ~ Removes crypto browserify bundle
- Removed dependency on WalletConnect => 314Kb (-233Kb) 
- Removed dependency on Web3Modal => 104Kb (-210Kb)

biwano:
Output of duplicate-package-checker-webpack-plugin

```@babel/runtime
  Multiple versions of @babel/runtime found:
[...]

PR #748, opened 2022-10-19
Adds a bundle analyzer for site

biwano:
## Description

Adds a bundle analyzer for app

to use do `ANALYZE=true npm run build-site`

The installation instructions https://www.npmjs.com/package/@next/bundle-analyzer says to check if we are in dev mode since we install @next/bundle-analyzer as a dev dependency. But it was not done for app. So I did the same for site. I suppose all environnements download the dev dependencies. I can add this extra layer of security if it is important.

## Related Ticket

Related to #720 (but does not fix)

## Changes

image

## Checklist


0xAeterno (approved):

site/package.json
```diff
@@ -37,6 +37,7 @@
     "@lingui/cli": "^3.13.0",
     "@lingui/loader": "^3.13.0",
     "@lingui/macro": "^3.13.0",
+    "@next/bundle-analyzer": "^12.3.1",
```
Atmosfearful, line 40: Can you also upgrade the one in `app` to match 12.3.1, looks like its 12.0.1 right now
```suggestion
    "@next/bundle-analyzer": "^12.3.1",
```

Atmosfearful (changes requested):

Atmosfearful (approved):

Issue #660, opened 2022-08-29
Connect Modal Redesign


jabby09:
This issue is the next iteration of the Connect modal on the app. We need to enhance and change this to help make the bridge between web2 users and web3.

The current version assumes a the user is web3 and has a wallet created.
![Image](https://user-images.githubusercontent.com/103535195/187316859-3c7f0c0e-2bbd-4bdb-923c-7f13cf0d7236.png)

In the next iteration, we need to redesign the modal to assume be inclusive of web2 users.

## Requirements

- Change journey entry point button CTA to:
    - “Login / Connect” or
    - “Login / Connect Wallet”
- Once in the modal, two sections:
    - “Login / Signup”
    - “Connect Wallet”
- “Login / Signup”
    - Social login powered by Torus
    - Show primary social login icon options on the card, i.e. google, facebook, twitter, discord
- “Connect Wallet”
    - Same options as we have currently
        - Wallet Connect
        - Coinbase Wallet
- Include some CTA for a user to get help/support if not sure how to proceed

## Design Sketch
![Image](https://user-images.githubusercontent.com/103535195/187316735-6b555ede-6713-42ac-972f-af9185f766cd.png)

## Other Notes
- For the social login with Torus, note cannot change or customize the current Torus implementation

jabby09:
Shipped design: https://www.figma.com/file/yhjZJKulthgEtixt6L6XPW/Login-%2F-Connect-Modal?node-id=0%3A1#258979350

jabby09:
@Atmosfearful ready for dev assignment

Atmosfearful:
@jabby09 Kicked off with @angusfretwell 
I hope this also fixes #659 as I don't have any idea how to debug that with our current modal 0_0

jabby09:
hey @angusfretwell how are things going on this task; any updates you can share?

jabby09:
Hi @angusfretwell have not heard from you in a bit, hope things are going well. Can you please share an update on status of development for this feature? When do you anticipate being ready for QA?

cc @Atmosfearful

jabby09:
@goatsOnDrugs design is above in one of the comments and this is the Notion roadmap card: https://www.notion.so/klima-dao/Connect-Modal-Redesign-49fbbac5d59647e1b475896aa727d3d0

Atmosfearful:
@goatsOnDrugs Documenting some behavior and expectations:
[...]

PR #789, opened 2022-11-09
Reduce _app bundle size -800kb

Atmosfearful:
## Description

1. Lazyload `web3modal` and all wallet provider dependencies (`coinbase`, `torus-embed`, `wallet-connect`) only after page mount, and only on pages where the web3 hook is used. We should maintain this pattern for #660 
2. upgrade @torus-embed which does reduce the size of the lazy-loaded bundle by ~200kb
3. don't import entire ethers object (this did not end up reducing any kb)
4. remove @ethersproject import (this also had no impact)

Follow my commit history for easier review.

## Before
First Load JS shared by all 1.01 MB

## After
First Load JS shared by all 221 kB

## Related Ticket

#720 

This will conflict with work on #660 

* [ ] Need to test and make sure I didn't break the login/connect experience
* [ ] Need to re-run bundle analyzer and see if further improvements can be made

### DIFF
--- a/app/actions/app.ts
+++ b/app/actions/app.ts
@@ -1,4 +1,4 @@
-import { ethers } from "ethers";
+import { utils } from "ethers";
 
 import { Thunk } from "state";
 import { setAppState } from "state/app";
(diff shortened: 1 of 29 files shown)


biwano (approved):
Nothing to say except that I learned stuff reading this code. ty.

PR #799, opened 2022-11-18
New Sign In / Connect Modal

### LINKED ISSUE #660: Connect Modal Redesign

goatsOnDrugs:
## Description

@ShimonD-KlimaDAO 
besides making sure the connect/disconnect functionality works you will want to check that if you are using brave browser and metamask is disabled you should see the brave button. If you have metamask you should see the metamask button but no brave button. If the brave setting (brave://settings/wallet) is set to brave wallet only then the brave button should appear even if metamask is enabled.

## Related Ticket

Resolves #660 

## Checklist


lib/utils/useWeb3Modal/index.tsx
```diff
@@ -29,52 +28,112 @@ export const useWeb3Modal = (strings: Web3ModalStrings): Web3ModalState => {
...
-      if (!modalRef.current) {
-        modalRef.current = await downloadModal();
-      }
-      if (modalRef.current.cachedProvider) {
-        connect();
-      }
-    };
-    init();
+    if (window.ethereum) {
+      connect();
```
Atmosfearful, line 127: We need to maintain the behavior that if a user connects and refreshes the page, it should auto reconnect to the same wallet.

If they disconnect, and refresh, it should not try to connect again.

So we can't rely on `window.ethereum` as 1. that only works for `metamask`/`brave`, and 2. it's always there even after they disconnect.
We probably need to maintain provider info in `LocalStorage`. I wonder if we can just write the entire ethers `provider` object to localstorage... might be worth looking how web3modal does this under the hood?

lib/utils/useWeb3Modal/index.tsx
```diff
@@ -29,52 +28,112 @@ export const useWeb3Modal = (strings: Web3ModalStrings): Web3ModalState => {
...
-    return newModal;
-  };
+  const connect = async (wallet?: string): Promise<void> => {
+    try {
+      if (wallet === "metamask" || wallet === "brave" || window.ethereum) {
+        const provider = new ethers.providers.Web3Provider(
+          window.ethereum as any,
+          137
+        ) as any;
+        await provider.send("eth_requestAccounts", []);
```
Atmosfearful, line 38: what does this do?

lib/utils/useWeb3Modal/index.tsx
```diff
@@ -29,52 +28,112 @@ export const useWeb3Modal = (strings: Web3ModalStrings): Web3ModalState => {
     }
   };
 
-  const downloadModal = async () => {
-    const newModal = (await import("../getWeb3Modal")).getWeb3Modal(strings);
```
Atmosfearful, line 33: This was from my recent PR #789 and it removes ~800kb of dependencies from the initial page load, so we need to keep this `download` behavior. Or do a quick follow up PR to enhance w/ lazyloading.

Now that we wrote the modal component ourselves instead of relying on web3modal, we can probably use [Next Dynamic Imports w/ suspense](https://nextjs.org/docs/advanced-features/dynamic-import#example)

requirements:
* [ ] All references to web3 provider libraries (coinbase, torus, walletconnect) are contained in the `<ConnectModal />` component.
* [ ] when the page mounts it should lazyload in the background, even if the modal is not triggered/visible
* [ ] by the time the user clicks "connect" to open the modal, it should be downloaded
* [ ] maybe support a loading placeholder modal via react suspense, e.g. on slow connections where they click Connect before it can finish downloading

Atmosfearful (changes requested):
not sure why package-lock has changes but package.json does not-- can you try deleting the lock and running npm install again?

lib/utils/useWeb3Modal/index.tsx
goatsOnDrugs: added comment `        // if user is not already connected this request will prompt the wallet modal to open and the user to connect
`

lib/utils/useWeb3Modal/index.tsx
goatsOnDrugs: Fixed. We are now storing a web3-wallet value in localStorage that is used to reconnect a user on refresh. web3 modal also does this.

lib/utils/useWeb3Modal/index.tsx
goatsOnDrugs: added dynamic imports of specific libraries in useWeb3Modal as discussed on call.

Atmosfearful (changes requested):
* [ ] button text is not visible on dapp, but is visible on site 🤔 probably a lightmode/darkmode thing
* [ ] we need to wire the button on /pledge marketing page as well

goatsOnDrugs:
@Atmosfearful where should the connect button go on /pledge page?

jabby09:
@goatsOnDrugs Looking good man! I've successfully connected using MM, google, discord, email, and twitter. 

- [ ] The only issue I have noticed (which is not necessarily a bug) is if I connect using one of the social login methods, and then disconnect my wallet, the connect modal comes up again, however, no matter what option I choose it just logs me in based the last successful login I had. This all happens in the same browser session. If I exit the browser and reopen then the social logins opens the torus modal as expected.

jabby09:
Comment from @Atmosfearful:
@goatsOnDrugs if you look at the current code you can see where I force torus to fully logout when the user clicks disconnect. I'd say that is a showstopper tbh

ShimonD-KlimaDAO:
@goatsOnDrugs @jabby09 
This is great. It passed all of my QA on Brave, Chrome, Firefox - on both desktop and mobile.
The connections work as expected, the design is exactly as expected, and the behavior with Brave is as expected.
Approved.

Atmosfearful (changes requested):
Some show stoppers:
* [x] Disconnect is not working. When I click the disconnect button in `app` it refreshes the page but then reconnects the same address instantly. Tried with both Metamask and WalletConnect
* [x] If I disconnect via the wallet itself, and refresh the `app`, it successfully disconnects me. However, the previous wallet keeps trying to reconnect itself whenever I refresh the page. It also seems to block me from switching to a different provider. I suspect there is some cached provider that needs to be cleared, and we need to be smarter about when we invoke the "request wallets" or whatever. There should be a disconnect event we can listen for.

ShimonD-KlimaDAO:
@goatsOnDrugs 
The buttons are not showing any text within the modal on the pledge dashboard. Below are Brave and Chrome on desktop.
They are working, but no text.
The same is true for mobile.

image
image

goatsOnDrugs:
@ShimonD-KlimaDAO fixed the button text issue. I also fixed the network switching and "disconnect manually from wallet" support as well as clearing wallet connect and coinbase cached provider on disconnect

jabby09:
@goatsOnDrugs @ShimonD-KlimaDAO thanks, I smoke tested both of these fixes and confirm they are working now. 

I'll let Shimon have the final say if this is good to merge and move to production.

Atmosfearful (approved):

PR #808, opened 2022-12-01
Source code checks github workflow

0xemc:
## Description

Automates the required checks as part of creating a PR. Will block PRs if the following fails:

* Build
* Lint
* Prettier
image

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -20,11 +20,3 @@ Resolves #<issue-number>
 |img here |img here|
 
 -->
-
-## Checklist
-
-<!-- Check completed item: [X] -->
-
-- [ ] Building the site with `npm run build-site` works without errors
-- [ ] Building the app with `npm run build-app` works without errors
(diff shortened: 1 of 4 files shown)


Atmosfearful (commented):
We already have the vercel preview builds for `site` and `app`
Both of those do a lint (via `next build`) and a prettier check (via `prettier --check .` defined in `package.json.scripts.build`)  and this covers the `lib` as well.

So do these checks accomplish something that isn't already covered by vercel?

0xemc:
> We already have the vercel preview builds for `site` and `app` Both of those do a lint (via `next build`) and a prettier check (via `prettier --check .` defined in `package.json.scripts.build`) and this covers the `lib` as well.
> 
> So do these checks accomplish something that isn't already covered by vercel?

No you're totally right, I wasn't aware that there was linting as part of the Vercel build. Also, where is `package.json.scripts.build` defined? 

The only advantage these workflows might give us is programmatically blocking breaking builds (not sure if the Vercel action can do that).

PR #823, opened 2022-12-13
check for infura id and use that provider

### LINKED ISSUE #739: Migrate away from ethers getDefaultProvider to avoid unecessary network calls
This method is creating network requests to cloudflare and a bunch of other services we don't need.
We need to just pick an ethereum provider URL (infura?) and run with it so we don't have to invoke this method.

goatsOnDrugs:
## Related Ticket

Resolves #739

## Checklist


lib/utils/getInfuraUrl/index.ts
```diff
@@ -1,4 +1,4 @@
-import { urls } from "@klimadao/lib/constants";
+import { urls } from "../../constants";
```
Atmosfearful, line 1: `lib` package should not depend on any node env variables. Goal is to keep the `lib` package totally agnostic of the environment (someday, published as a standalone component library to be used by the community)

So we can instead export a util that accepts the infuraId which is passed down at run time from `site` or `app`

The infuraId should be a required param. The two functions can be combined into one

```ts
export const getInfuraUrl = (params: { chain?: "polygon" | "eth", infuraId: string }) => {
  const { chain = "polygon" } = params; // fallback to polygon as default
  const baseUrl = chain === "eth" ? urls.infuraEth : urls.infuraRpc;
  return `${baseUrl}/${params.infuraId}`;
};
```

Atmosfearful (changes requested):
Your PRs always have a package-lock change for some reason. plz don't commit these. Make sure you are using `npm` v 8.11 or higher. I will go a head and rebase this on staging and remove the package-lock changes from this PR to save you some time. Just pull latest changes down, nuke node_modules and run npm install and it should show zero changes.

lib/utils/getInfuraUrl/index.ts
goatsOnDrugs: Ok, makes sense. Updated to be one function with infura id required. Everyone will need to update their env files (create one on app) for app/site to work properly. Is this intended or do we still want to fall back to defaultProvider somewhere? @Atmosfearful

site/components/pages/Pledge/PledgeDashboard/Profile/index.tsx
```diff
@@ -32,11 +31,21 @@ export const Profile: FC<Props> = (props) => {
     const setProfile = async () => {
       const kns = await getKNSProfile({
         address: props.pledge.ownerAddress,
-        providerUrl: getInfuraUrlPolygon(),
+        providerUrl: getInfuraUrl({
+          chain: "polygon",
+          infuraId: process.env.INFURA_ID!,
```
ladytrekker, line 36: With the "!" you are sure that the env exists, right? (non-null assertion)

But how should the env var locally be defined?
Do all developers now need to have their own `env.local` file in "site" and "app"?

I would remove the "!" if possible 
and extend the readme with something like "you need an env.local file with an infura ID" or alike
and ensure that the function fails without the env var.

What do you think?

site/locale/de/messages.po
```diff
@@ -1021,7 +1021,7 @@ msgstr ""
 
 #: components/pages/Pledge/InvitationModals/AcceptModal/index.tsx:153
 msgid "pledge.modal.confirm"
-msgstr ""
+msgstr "Bestätigen"
```
ladytrekker, line 1024: Is it possible to remove the changes on the locale files?
You did not change any strings in this PR, correct?

site/components/pages/Pledge/PledgeDashboard/Profile/index.tsx
goatsOnDrugs: Yes, good point. We would need to all have an infura id in app and site. I created an env.example file in app and updated the one in site. Do you think I should add this to the contribution / style guide or the main readme?

I replaced the `!` with `as string`. Is that what you were thinking of instead of the `!`?

Thanks!

site/locale/de/messages.po
goatsOnDrugs: I think the rebase took care of all the non english files.

goatsOnDrugs:
Hey, @ShimonD-KlimaDAO would you mind smoke testing this? We are relying on infura for RPC now. Please check kns, klima supply, block rate,

goatsOnDrugs:
@Atmosfearful can you please add NEXT_PUBLIC_INFURA_ID to site and app env files for prod/staging?

site/components/pages/Pledge/PledgeDashboard/Profile/index.tsx
ladytrekker: Yes, I would add it to the readme.
Otherwise running site and app locally fails, correct?

In this PR it was added to the README already:
https://github.com/KlimaDAO/klimadao/pull/834/files

Just check if this is already enough to mention it.

Atmosfearful (approved):

PR #837, opened 2022-12-19
reset modal to connect state when opening

goatsOnDrugs:
## Description
The modal gets stuck in an error state if the user closes the modal in error state and re opens it. 

## Related Ticket

Resolves #<issue-number>

## Checklist

### DIFF
--- a/lib/components/ConnectModal/index.tsx
+++ b/lib/components/ConnectModal/index.tsx
@@ -26,6 +26,7 @@ export const ConnectModal = (props: Props) => {
 
   useEffect(() => {
     if (showModal) {
+      setStep("connect");
       document.body.style.overflow = "hidden";
     } else {
       document.body.style.overflow = "";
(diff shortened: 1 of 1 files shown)


lib/components/ConnectModal/index.tsx
```diff
@@ -24,6 +24,11 @@ export const ConnectModal = (props: {
       document.body.style.overflow = "";
     }
   }, [showModal]);
+  useEffect(() => {
+    if (showModal) {
+      setStep("connect");
```
Atmosfearful, line 29: just add this to L22 instead of making a second effect

Atmosfearful (changes requested):

lib/components/ConnectModal/index.tsx
goatsOnDrugs: updated. sorry it took so long.

Atmosfearful (approved):

PR #840, opened 2022-12-21
[selective-retirements] Select from projects with more than 1 tonne remaining

### LINKED ISSUE #819: [offset] Don't show projects with less than 1 tonne in remaining quantity
AKA only query for projects where quantity > 1 tonne. We have too many un-useable projects filling the selective retirement search

0xAeterno:
## Description

Only allow projects with more than 1 tonnage remaining to be selected from the advanced selective retirement flow.

## Related Ticket

Resolves #819 

### DIFF
--- a/app/components/views/Offset/ProjectSearch/ProjectSearchForm/index.tsx
+++ b/app/components/views/Offset/ProjectSearch/ProjectSearchForm/index.tsx
@@ -54,7 +54,7 @@ export const ProjectSearchForm: FC<Props> = (props) => {
       (project) =>
         Number(
           project[`balance${selectedRetirementToken}` as BalanceAttribute]
-        ) > 0
+        ) > 1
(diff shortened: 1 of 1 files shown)


app/components/views/Offset/ProjectSearch/ProjectSearchForm/index.tsx
```diff
@@ -54,7 +54,7 @@ export const ProjectSearchForm: FC<Props> = (props) => {
       (project) =>
         Number(
           project[`balance${selectedRetirementToken}` as BalanceAttribute]
-        ) > 0
+        ) > 1
```
Atmosfearful, line 57: would it make more sense to add this in the subgraph query, so it returns less results?

Atmosfearful (approved):

PR #844, opened 2022-12-23
[selective-retirements] Limit filter options by selected pool

### LINKED ISSUE #817: [offset] Limit Filter Options by Pool
Currently in the project section of selective retirement the filter options are based on all available data across all the carbon pools. 

A UX enhancement would be to limit the filter options based on what is available in the carbon pool selected. 

Note from testing:

> (medium impact) We shouldn't have the option to select a single box that doesn't have any matching records. For example, there is 1970 vintage in UBO and NBO, but not in BCT and NCT, but it's available as an option in NCT, returning nothing with a single box selection. That seems to me like poor UX.

[Discord note here](https://discord.com/channels/841390338324824096/964698508751949824/1049835396051247134).

0xAeterno:
## Description

Reduce the number of filter options based on the selected pool token.
Currently manually generating the list found in `filterOption.ts`.
While not ideal, this manual process is okay for now as we're not expecting frequent updates to the subgraph data atm. We can look at automating the list generation when it's required.

## Related Ticket

Resolves #817 

## Checklist


app/components/views/Offset/ProjectSearch/filterOptions.ts
```diff
@@ -1,61 +1,99 @@
...
+      { label: "2019", value: "2019" },
+    ],
+  },
+  nbo: {
+    types: [
+      { label: "Forestry", value: "Forestry" },
+      { label: "Other Nature Based", value: "Other Nature Based" },
+    ],
+    regions: [
+      { label: "Asia", value: "Asia" },
```
Atmosfearful, line 82: we should translate these. Can go in a follow up

Atmosfearful (approved):
I guess the one benefit of the bridges being turned off is we won't have to update this config for a while :) We definitely will need to automate this in the future though!

PR #867, opened 2023-01-10
Goats/828 connect buttons

### LINKED ISSUE #828: [connect modal] Tech debt - separate modal from button
After merging #799 

Rather than rending a new modal wherever the button is used, we can render one modal and place it higher in the react component tree in Site and in App. It can be controlled to show/hide based on react context

The react context which controls the modal can be triggered using a hook, ideally `useWeb3`. For example
```jsx
const { toggleModal }  = useWeb3();

return <button onClick={toggleModal}>Click Me To Show The Connect Modal</button>
```

Why this pattern?

- We can use any button or any component to trigger the Connect modal
- We won't need style overrides like we do in all the app views (`100% width`)
- Can remove a lot of lines of code repeating those modal props over and over

goatsOnDrugs:
## Description
Separate buttons from connect modal. This allows us to pass less props and render the modal less. We can now have custom components that open the modal.  Aw yea.

## Related Ticket

Resolves #828 

## Checklist


lib/components/Web3Context/index.tsx
```diff
@@ -10,10 +11,16 @@ interface Props {
 }
 /** Init the web3Modal and expose via react context  */
 export const Web3ContextProvider: FC<Props> = (props) => {
-  const web3ProviderState = useWeb3Modal();
-
+  const providerState = useProvider();
+  const [showModal, setShowModal] = useState(false);
+  const toggleModal = () => setShowModal((s: boolean) => !s);
```
Atmosfearful, line 16: isnt the `boolean` type auto-inferred?
```suggestion
  const toggleModal = () => setShowModal((s) => !s);
```

lib/components/Web3Context/index.tsx
```diff
@@ -10,10 +11,16 @@ interface Props {
...
+  const renderModal = (props: any) => (
```
Atmosfearful, line 17: Provide a type for `props`
You can export the ConnectModalProps or slightly more ugly you can do `React.ComponentProps<typeof ConnectModal>`

lib/components/Web3Context/types.ts
```diff
@@ -71,6 +72,9 @@ export type Web3State = ConnectedWeb3State | DisconnectedWeb3State;
 export type Web3ModalState = Web3State & {
   connect?: (wallet?: string) => Promise<void>;
   disconnect?: () => Promise<void>;
+  showModal?: boolean;
+  renderModal?: FC;
+  toggleModal?: () => void;
```
Atmosfearful, line 77: set more strict types so you don't have to keep checking for existence everywhere you use it
```suggestion
  showModal: boolean;
  renderModal: () => ReactNode;
  toggleModal: () => void;
```

And then on L80 web3InitialState provide dummy defaults
```
{
  showModal: false,
  renderModal: () => null,
  toggleModal: () => {},
}
```

site/components/pages/Pledge/index.tsx
```diff
@@ -108,36 +133,16 @@ export const Pledge: NextPage = () => {
...
+                    variant="blue"
+                    label={t({
                       message: "Create a pledge",
                       id: "pledges.home.hero.create",
                     })}
-                    buttonVariant="blue"
-                    onClose={handleCreatePledge}
+                    onClick={() => {
+                      handleCreatePledge();
+                      toggleModal && toggleModal();
```
Atmosfearful, line 144: this is preferred syntax for future reference. However when you update the initial state and the type as I suggested above, then you can just invoke directly and dont need to check for existence.
```suggestion
                      toggleModal?.();
```

* [ ] after you update the types/defaults please ctrl+shift+F for all invocations of toggleModal and renderModal and remove the null checks, invoke them all directly

lib/components/ConnectModal/index.tsx
```diff
@@ -50,44 +76,108 @@ export const ConnectModal = (props: Props) => {
...
-      <ConnectContent
-        showModal={showModal}
-        handleConnect={handleConnect}
-        setShowModal={setShowModal}
-        step={step}
-        setStep={setStep}
-        errorMessage={props.errorMessage}
-        torusText={props.torusText}
-        titles={props.titles}
+  return props.showModal ? (
```
Atmosfearful, line 87: I would just early exist one line above. keeps it more flat and readable
```
if (!props.showModal) return null
```

lib/components/ConnectModal/index.tsx
```diff
@@ -13,26 +30,35 @@ interface Props {
...
   const [step, setStep] = useState<"connect" | "error" | "loading">("connect");
-  const { address, connect, disconnect, isConnected } = useWeb3();
+  const { connect, toggleModal } = useWeb3();
+  const focusTrapRef = useFocusTrap();
+  const [showMetamask, setShowMetamask] = useState(false);
+  const [showBrave, setShowBrave] = useState(false);
 
-  const buttonVariant = props.buttonVariant ?? null;
+  useEffect(() => {
+    if (window.ethereum && (window.ethereum as any).isBraveWallet) {
```
Atmosfearful, line 45: I just noticed these don't need to live in state because `window.ethereum` doesn't change at runtime.
they can just be variables.

* [ ] Assume window can also be undefined (server render).
* [ ] Also help typescript out a bit:

```ts
import { providers } from "ethers";
type WindowEthereum = providers.ExternalProvider && { isBraveBrowser: boolean }

const eth = window?.ethereum as WindowEthereum | undefined;
const showBrave = eth?.isBraveWallet;
const showMetamask = eth?.isMetamask;
const showCoinbaseWallet = eth?. isCoinbaseWallet;
```

Also coincidentally yapper reported a coinbase wallet issue a few hours ago and the bug is right here in this codeblock. See ticket (high priority) https://github.com/KlimaDAO/klimadao/issues/881

app/components/views/PKlima/index.tsx
```diff
@@ -36,6 +35,7 @@ interface Props {
   provider?: providers.JsonRpcProvider;
   address?: string;
   isConnected?: boolean;
+  toggleModal?: () => void;
```
Atmosfearful, line 38: should always exist, see related suggestions above plz update all the other `views` component props to be non-optional 
```suggestion
  toggleModal: () => void;
```

Atmosfearful (changes requested):
close, some easy cleanup things

app/components/views/PKlima/index.tsx
goatsOnDrugs: updated

lib/components/Web3Context/index.tsx
goatsOnDrugs: true, fixed

lib/components/ConnectModal/index.tsx
goatsOnDrugs: updated

site/components/pages/Pledge/index.tsx
goatsOnDrugs: updated. noted on the syntax.

lib/components/ConnectModal/index.tsx
goatsOnDrugs: updated. eth needs to be in state because we need to access window in a useEffect but the wallet checks are out of state now.

lib/components/Web3Context/types.ts
goatsOnDrugs: updated

lib/components/Web3Context/index.tsx
```diff
@@ -1,19 +1,27 @@
...
 /** Init the web3Modal and expose via react context  */
 export const Web3ContextProvider: FC<Props> = (props) => {
-  const web3ProviderState = useWeb3Modal();
-
+  const providerState = useProvider();
+  const [showModal, setShowModal] = useState(false);
+  const toggleModal = () => setShowModal((s) => !s);
+  const renderModal = (props: RenderModalProps) => (
+    <ConnectModal {...props} showModal={showModal} />
+  );
```
Atmosfearful, line 20: ah i just realized we should use useCallback here, so that this component only reinstantiates itself when `showModal` changes. Otherwise changes to `providerState` may cause it to fully unmount and remount which can cause scary hard-to-debug issues. ill drop that in and merge

Atmosfearful (approved):

Issue #773, opened 2022-10-31
[pledge] multi wallet improvements


0xAeterno:
- [ ] How do we improve `PledgeForm` to remove the need of `isDeleteMode`
  - Required to persist form state when editing secondary wallets
  
- [ ] Refactor accept and reject secondary wallet invitations to seperate endpoints
  - Actions are independent from putPledge => single responsibility 
  - Consider what validation is required

- [ ] move isOwner check to server
   -  formSchema should be independent from dependancies like the window object.
   - add translated error on pledge form

goatsOnDrugs:
@0xAeterno moved the schema refactor to here

jabby09:
moving to backlog, now a low priority

jabby09:
no longer relevant

PR #874, opened 2023-01-11
pledge enhancements

### LINKED ISSUE #791: [pledge] cleanup - validate incoming pledges
Use yup to validate pledge attributes on the server

goatsOnDrugs:
## Description
Use yup for pledges

## Related Ticket

Resolves #791 

## Checklist

### DIFF
--- a/site/pages/api/pledge/index.ts
+++ b/site/pages/api/pledge/index.ts
@@ -1,10 +1,39 @@
 import { findOrCreatePledge } from "components/pages/Pledge/lib/firebase";
 import { NextApiRequest, NextApiResponse } from "next/types";
+import { array, boolean, number, object, string } from "yup";
 
 export default async function handler(
   req: NextApiRequest,
   res: NextApiResponse
 ) {
+  const pledgeSchema = object({
+    id: string().nullable(),
+    ownerAddress: string().required(),
+    nonce: string().required(),
+    name: string().required(),
+    profileImageUrl: string().url(),
+    profileWebsiteUrl: string().url(),
+    description: string().required().max(500),
+    wallets: array().of(
+      object({
+        address: string().required(),
+        status: string().required(),
+        saved: boolean().required(),
+      })
+    ),
+    methodology: string().required().max(1000),
+    footprint: number().required().min(0),
+    categories: array()
+      .of(
+        object({
+          name: string().required().max(32),
+          quantity: number().required().min(0),
+        })
+      )
+      .required()
+      .min(1),
+  });
+
   switch (req.method) {
     case "PUT":
       try {
@@ -13,10 +42,22 @@ export default async function handler(
         if (!signature) {
           return res.status(400).json({ message: "Bad request" });
         }
-
+        await pledgeSchema.validate(req.body.pledge);
         const pledge = await findOrCreatePledge({
           pageAddress: req.body.pageAddress,
-          pledge: req.body.pledge,
+          pledge: {
+            id: req.body.pledge.id,
+            ownerAddress: req.body.pledge.ownerAddress,
+            nonce: req.body.pledge.nonce,
+            name: req.body.pledge.name,
+            profileImageUrl: req.body.pledge.profileImageUrl,
+            profileWebsiteUrl: req.body.pledge.profileWebsiteUrl,
+            description: req.body.pledge.description,
+            wallets: req.body.pledge.wallets,
+            methodology: req.body.pledge.methodology,
+            footprint: req.body.pledge.footprint,
+            categories: req.body.pledge.categories,
+          },
           signature,
           secondaryWalletAddress: req.body.secondaryWalletAddress,
           action: req.body.action,
--- a/site/pages/api/pledge/index.ts
+++ b/site/pages/api/pledge/index.ts
@@ -32,7 +32,9 @@ export default async function handler(
       )
       .required()
       .min(1),
-  });
+    createdAt: number(),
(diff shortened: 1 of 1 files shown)


0xAeterno (changes requested):
You can reuse the same validation schema we've already defined for pledge and the types we infer from the schema :)

site/pages/api/pledge/index.ts
```diff
@@ -13,10 +42,22 @@ export default async function handler(
         if (!signature) {
           return res.status(400).json({ message: "Bad request" });
         }
-
+        await pledgeSchema.validate(req.body.pledge);
```
0xAeterno, line 45: When thsi throws, what does the sent errors object look like client side?

site/pages/api/pledge/index.ts
```diff
@@ -13,10 +42,22 @@ export default async function handler(
...
+            nonce: req.body.pledge.nonce,
+            name: req.body.pledge.name,
+            profileImageUrl: req.body.pledge.profileImageUrl,
+            profileWebsiteUrl: req.body.pledge.profileWebsiteUrl,
+            description: req.body.pledge.description,
+            wallets: req.body.pledge.wallets,
+            methodology: req.body.pledge.methodology,
+            footprint: req.body.pledge.footprint,
+            categories: req.body.pledge.categories,
+          },
```
0xAeterno, line 60: Don't have full context here. Are we constructing the pledge with more attributes here or could this be simplified to `req.body.pledge`

0xAeterno: Just a thought, if you were concerned about other attributes being slipped in to the pledge.
The validation schema has [`.noUnknown()`](https://github.com/KlimaDAO/klimadao/blob/b6209d7af20e71fe5ff2cfadeaeaedbf665a1eb4/site/components/pages/Pledge/lib/formSchema.ts#L160) which will filter out any attributes that aren't specified in the schema.
https://github.com/KlimaDAO/klimadao/blob/b6209d7af20e71fe5ff2cfadeaeaedbf665a1eb4/site/components/pages/Pledge/lib/formSchema.ts#L160

site/pages/api/pledge/index.ts
goatsOnDrugs: It will be the default error I think. In theory it will only throw if someone is trying to use the api without our front end bc the form values are validated before they reach this point.

site/pages/api/pledge/index.ts
goatsOnDrugs: @0xAeterno noUnknown added. Feels cleaner now, thanks!

goatsOnDrugs:
@Atmosfearful I think this one is good to go? @0xAeterno how do you feel?

site/pages/api/pledge/index.ts
```diff
@@ -1,10 +1,41 @@
 import { findOrCreatePledge } from "components/pages/Pledge/lib/firebase";
 import { NextApiRequest, NextApiResponse } from "next/types";
+import { array, boolean, number, object, string } from "yup";
 
 export default async function handler(
   req: NextApiRequest,
   res: NextApiResponse
 ) {
+  const pledgeSchema = object({
```
jonathanly, line 9: Ah so this was missed in the overall comment.
Let's use the schema we've already defined for pledge form. I expect both shapes to be exactly the same so better not to duplicate this.

Only caveat I can think of is translations. We return a lingui translation id string so that probably needs to be handled on the FE.

site/pages/api/pledge/index.ts
jonathanly: Agreed!

jonathanly (changes requested):

site/pages/api/pledge/index.ts
0xAeterno, line 9: Ah so this was missed in the overall comment.
Let's use the schema we've already defined for pledge form. I expect both shapes to be exactly the same so better not to duplicate this.

Only caveat I can think of is translations. We return a lingui translation id string so that probably needs to be handled on the FE.

0xAeterno (changes requested):

site/pages/api/pledge/index.ts
goatsOnDrugs: @0xAeterno I hear ya but the schema doesnt work in pages/api because of the `isOwner` check using window. We get the error "window is not defined" bc its on the server. Also not sure if this will delete the updatedAt with the noUnknown check or if thats added by firebase (havent checked bc of window issue)

0xAeterno: Gotcha @goatsOnDrugs.
What's the purpose of `isOwner`?, IMO the formSchema should be independent from dependancies like the `window` object.
`isOwner` feels like a validation that should belong on the server, but also there's probably another way to approach how we validate here.

Can you please create a ticket with some context for us to look at this later. I'm okay with coming back to this later

goatsOnDrugs: isOwner checks that the user isnt adding the pledge owner address as a secondary wallet. I can create a ticket for this right meow.

goatsOnDrugs: TY for the review!

goatsOnDrugs:
@Atmosfearful I think we can merge this. I moved the form schema refactor to #773

0xAeterno (approved):

Atmosfearful (approved):

PR #876, opened 2023-01-11
Refactor RPC calls to remove NEXT_PUBLIC_INFURA_URL

Atmosfearful:
### History / background
- We added Infura as a server-side RPC gateway instead of polygon-rpc because we had problems with reliability leading to flaky builds in the early days #231
- Later on to support ENS, we needed an Ethereum RPC gateway, so we used `ethers.getDefaultProvider`. This works, but creates a buttload of network requests and console errors #739 
- So we decided to add an infura id to be used on the client for these ENS requests, but this has the downside of breaking builds for devs who don't have their own INFURA_ID

## Changes
- Use `https://cloudflare-eth`, a free public eth gateway, for our ENS requests instead of Infura
- Remove the `NEXT_PUBLIC_INFURA_ID` variable
- update types to assume that INFURA_ID can be undefined
- Create a utility called `getStaticProvider` that replaces `getJsonRpcProvider`
  - it constructs the right URL for you based on the chain and accepts an optional `infuraId`
  - it uses sane fallbacks when no infuraId is provided
  - it uses a single runtime provider instance rather than instantiating a new one on every invocation.
- refactor some params to pass down an optional `infuraId` instead of the whole url or provider.
- update the .env.example files

## Related Ticket
#823

## Checklist


lib/utils/getStaticProvider/index.ts
```diff
@@ -0,0 +1,54 @@
...
+  },
+};
+/**
+ * Returns a provider for making RPC calls to the given chain.
+ * 'static' because the RPC is chosen by us, can't be changed by the user, and can't be used for signing txns.
+ * For server environments, we recommend using infura for (historically) improved stability & performance.
+ * If no infura id is provided, a default public rpc is used.
+ */
+export const getStaticProvider = (params?: {
+  chain?: "polygon" | "eth";
```
ladytrekker, line 35: What about "mumbai"?

Atmosfearful: good call, done. See recent commits.

ladytrekker: 👍

ladytrekker (approved):
Thanks, looks good to me

goatsOnDrugs (approved):
Looks good!

PR #877, opened 2023-01-12
[redemptions] add feature flagging mechanism

0xAeterno:
## Description

Feature flagging!!! 🏴‍☠️

one-piece-anime

### Smol guide
Add a feature flag key to `getFeatureFlags` in `lib/featureFlags/index.ts` and update the types.
There are two constants for flagging configuration:
```ts
LIVE_FEATURE // toggled on for all environments (e.g. live in production)
IN_DEVELOPMENT // toggled on for preview and local dev
```

#### Usage
As per @Atmosfearful comment, we don't want to maintain and sync two sets of feature flagging config.
Config will live in the shared lib and be re-exported to where it's initialised in `app` & `site`.
So how do we use it, it's simple :)
```ts
// example.tsx in app/site
import { features } from "lib/featureFlags"

export const Example = () => {
  if (features.newFeatureFlagKey) {
    return <NewFeature />
  }

  return <ExistingCode />
}
```

## Checklist


app/lib/hooks/useFeatureFlags.ts
```diff
@@ -0,0 +1,29 @@
...
+const PRODUCTION: FeatureFlags = { testFeatureFlag: false, redemptions: true };
+
+const getFeatures = (): FeatureFlags => {
+  if (IS_PREVIEW_BUILD) return PREVIEW;
+  if (IS_PRODUCTION) return PRODUCTION;
+
+  return DEVELOPMENT;
+};
+
+export const useFeatureFlags = () => {
```
Atmosfearful, line 21: A hook would be useful if we wanted to update these on-the-fly, like based on some state that can change at runtime.
But our prod builds only take 3 mins or less, and we have preview links for everything, so static is fine

Also the downside of the hook is we can't use it outside of a react context. We may need to flag things inside of actions, utils and getStaticProps (for redirects)

So I'd just go with a simple config object. The only 3 states we need are
1. Enable everywhere
2. Only enable on preview + local (hide on prod)
3. Only enable on local

```ts
const featureFlags = {
  selectiveRetirement: true, // enable everywhere
  aggregatorV2: !IS_PRODUCTION, // enable on preview and local
  redemptionUI: IS_LOCAL_DEVELOPMENT, // enable on local
}
```

hold up: we have the complexity that env vars don't exist in `lib`. But we don't want two separate configs.

So I think the best way to do this is to reexport at runtime
```ts
// in `lib`
const getFeatureFlags = (params: { isProduction: boolean, isLocal: boolean }) => {
  const featureFlags = {
      selectiveRetirement: true, // enable everywhere
      aggregatorV2: !params.isProduction, // enable on preview and local
      redemptionUI: isLocal, // enable on local (hide on preview and prod)
  };
  return featureFlags
}

// in `site/lib/featureFlags` and `app/lib/featureFlags`
export const featureFlags = getFeatureFlags({ isProduction: IS_PRODUCTION, isLocal: IS_LOCAL_DEVELOPMENT });

// in react component
import { featureFlags } from "lib/featureFlags"
if (featureFlags.redemptionUi) return <RedemptionUI />
```

app/lib/hooks/useFeatureFlags.ts
```diff
@@ -0,0 +1,29 @@
+import { IS_PREVIEW_BUILD, IS_PRODUCTION } from "lib/constants";
+import { useMemo } from "react";
+
+type FeatureFlags = {
```
Atmosfearful, line 4: probably a good idea to maintain a strict type like this 👍 as these flags will be imported all over the app, ts can warn us when we forget to clean something up.

app/lib/hooks/useFeatureFlags.ts
0xAeterno: Love this, it's way more simple than the original implementation.

To improve on this suggestion further, we can drop the `isLocal` param altogether.
I reakon its redundant to enable it on local dev only. I consider preview builds a "development" environment too.

app/lib/hooks/useFeatureFlags.ts
0xAeterno: 💯!! Left a note to remind us to clean up feature flags too when features are live and stable for a while.

lib/featureFlags/index.ts
```diff
@@ -0,0 +1,23 @@
+/* 
+  Strict types for current feature flag set
+  Note: Be proactive in cleaning up feature flags once feature is live and stable
```
Atmosfearful, line 3: thx for the comments

lib/featureFlags/index.ts
```diff
@@ -0,0 +1,23 @@
...
+  redemptionUI: boolean;
+};
+
+export const getFeatureFlags = (params: {
+  isProduction: boolean;
+}): FeatureFlags => {
+  /** toggled on for all environments */
+  const LIVE_FEATURE = true;
+  /** toggled on for preview and local dev */
+  const IN_DEVELOPMENT = !params.isProduction;
```
Atmosfearful, line 17: smart, easier to see everything at a glance.

Atmosfearful (approved):
image

Issue #880, opened 2023-01-12
Never use white-text-over-Klima-Green


Atmosfearful:
we must pay the price of rushed design decisions pre-launch (choosing this as our primary color 🤮 )

Typically we have "primary font" color which is white in light mode, black in dark mode. But because of color-contrast concerns, we must now use a different color when klima-green is our background. 

Related to @ShimonD-KlimaDAO 's note here: https://github.com/KlimaDAO/klimadao/pull/834#issuecomment-1377781060

And a discussion in discord here:
https://discord.com/channels/841390338324824096/964698337230086174/1059953750820790354

# Task
As far as I know this only applies to buttons, so lets start there
* [ ] the top right button in the site
* [ ] the confirmation / connect buttons in the app (top right, but also seen in the retirement ui and elsewhere)
* [ ] buttons on the /retirement pages
![image](https://user-images.githubusercontent.com/88635679/212184482-c25d3a59-d8bb-473b-ace7-aa1a43b01893.png)

# Pass
![gud](https://cdn.discordapp.com/attachments/964698337230086174/1060275413856833617/image.png)

# Fail
![bad](https://cdn.discordapp.com/attachments/964698337230086174/1060275118485545090/image.png)

PR #882, opened 2023-01-13
buttons to be explicitly circular via circle prop (861)

### LINKED ISSUE #861: regression: various buttons have a circle style applied
Various buttons which previously did not have a circle style applied are now styled as circles.

| What | Before | After |
| ----- | ----- | ----- |
| Resource bar search and filter option button | image|image|
| Pledge form |  image | image |

Link to previous build before regression introduced: 
- [Resource center](https://klimadao-site-git-goats-icon-bug-klimadao.vercel.app/resources)
- [Pledge](https://klimadao-site-git-goats-icon-bug-klimadao.vercel.app/pledge)

I've tracked the style regressions to this line that applies the circle styles:
https://github.com/KlimaDAO/klimadao/blob/038e1a4666a8ab9984b0089afa0653aa7584c4b7/lib/components/Buttons/ButtonBase.tsx#L33-L37

0xemc:
## Description

* Reverts changes made to automatically create circular buttons when only an `icon` prop was provided. This must now be explicitly set via the `circle` prop
* Font size and color are also no longer automatically inherited and must be explicitly set. 

## Related Ticket

Resolves #861

## Changes

| Before  | After  |
|---------|--------|
|image|image|

## Checklist


0xAeterno (commented):
Looking good @0xemc. Thanks for jumping on this one.
I did a quick scan and it's looking a lot better. One thing I did note was that the color of the icons in light mode don't look quite right. I'm not confident in saying this was a regression introduced at some point or it was always the case.
Would you be able to have a poke around?  I'm pretty sure the color should be the same as the placeholder text.

image

0xAeterno (changes requested):
As per comment above :)

lib/components/Buttons/ButtonBase.tsx
```diff
@@ -22,6 +22,7 @@ export type ButtonBaseProps = {
   disabled?: boolean;
   /** Apply rounded corners */
   rounded?: boolean;
+  circle?: boolean;
```
Atmosfearful, line 25: I assume circle and rounded are never both true. So how about:
```suggestion
  shape?: "default" | "rounded" | "circle";
```

Where 'default' is the fallback if no prop is provided.
This will require you to update ~19 instances where the `rounded` prop is used.

lib/components/Buttons/styles.ts
```diff
@@ -53,6 +51,7 @@ const buttonBase = css`
 export const buttonPrimary = css`
   ${buttonBase};
   background-color: var(--klima-green);
+  color: white; /* same in darkmode */
```
Atmosfearful, line 54: So a new contrib @0xMakka was gonna tackle #880 as a first-ticket. Perhaps it's as easy as updating this line ahaha. Anyways ill leave that to him!

Atmosfearful (changes requested):

lib/components/Buttons/ButtonBase.tsx
0xemc: Nice, that update with this 👍

lib/components/Buttons/styles.ts
0xemc: To be fair there is a bit of selector wars going on so I doubt if this would be enough. See how he goes though.

0xemc:
> Looking good @0xemc. Thanks for jumping on this one. I did a quick scan and it's looking a lot better. One thing I did note was that the color of the icons in light mode don't look quite right. I'm not confident in saying this was a regression introduced at some point or it was always the case. Would you be able to have a poke around? I'm pretty sure the color should be the same as the placeholder text.
> 
> image
> 
> Edit: Looked into it a bit more and confirmed my suspicions it's also a regression at some point. Here's a screenshot of it way back in the day. Link to ancient preview build: `https://klimadao-site-git-release-pledge-klimadao.vercel.app/pledge/<address>` (@Atmosfearful surprised these builds are still up, does it cost us anything to keep these deployments?) image

Ok should definitely be good to go now as I removed all automatic styling and made everything explicit and opt-in.  Unfortunately in a lot of places the background and font colors are being set explicitly for svgs, backgrounds and font colors so it was going to be a nightmare to find all of those and make them work with the automatic style 🤷

I had a quick look around and everything seemed ok. If you find anything else let me know. Cheers!

0xAeterno:
> Ok should definitely be good to go now as I removed all automatic styling and made everything explicit and opt-in. Unfortunately in a lot of places the background and font colors are being set explicitly for svgs, backgrounds and font colors so it was going to be a nightmare to find all of those and make them work with the automatic style 🤷
> 
> I had a quick look around and everything seemed ok. If you find anything else let me know. Cheers!

Appreciate you having a scan through the codebase for other instances where this might be an issue @0xemc!
What you've addressed is perfectly fine :) We can raise and address future instances as they come.
One of the challenges having contributors come and go at the beginning, so it feels like a broader style review is required at some point to get alignment on how we do things, now that there seems to be more stability.

Atmosfearful (approved):

0xAeterno (approved):

0xemc:
@0xAeterno @Atmosfearful looks like this is ready to go, just need it merged by someone with the power ⚡

PR #887, opened 2023-01-13
fix(app): don't allow fractional quantities when fiat

### LINKED ISSUE #818: [offset] Don't allow user to enter fractional quantity when credit card is selected
Stripe does not allow fractional quantities. If this becomes a problem in the future we will have to refactor the backend so that quantities are handled in kilograms instead of tonnes. For now, let's fix it in the UI and enforce whole-tonne retirements when Credit Card is selected.

 In PR #743 I added a line to round the quantity that is send to the checkout API, but this is poor UX.

This is not a high priority ticket.

0xMakka:
## Description

This PR fixes an issue where a user can enter a fractional quantity when inputting the tonnes of carbon to offset (when the "Credit Card" option is selected as the payment method.

## Related Ticket

Resolves #818

## Changes

| Before  | After  |
|---------|--------|
|Screenshot 2023-01-13 at 23 11 21 | Screenshot 2023-01-13 at 23 11 56|

## Checklist

### DIFF
--- a/app/components/views/Offset/index.tsx
+++ b/app/components/views/Offset/index.tsx
@@ -94,7 +94,7 @@ export const Offset = (props: Props) => {
   const [isRetireTokenModalOpen, setRetireTokenModalOpen] = useState(false);
   const [isInputTokenModalOpen, setInputTokenModalOpen] = useState(false);
   const [isInvalidFractionalQtyFiat, setIsInvalidFractionalQtyFiat] =
-    useState<boolean>(false);
+    useState(false);
   const [paymentMethod, setPaymentMethod] =
     useState<OffsetPaymentMethod>("fiat");
   const [selectedRetirementToken, setSelectedRetirementToken] =
--- a/app/components/views/Offset/index.tsx
+++ b/app/components/views/Offset/index.tsx
@@ -93,6 +93,8 @@ export const Offset = (props: Props) => {
   // local state
   const [isRetireTokenModalOpen, setRetireTokenModalOpen] = useState(false);
   const [isInputTokenModalOpen, setInputTokenModalOpen] = useState(false);
+  const [isInvalidFractionalQtyFiat, setIsInvalidFractionalQtyFiat] =
+    useState<boolean>(false);
   const [paymentMethod, setPaymentMethod] =
     useState<OffsetPaymentMethod>("fiat");
   const [selectedRetirementToken, setSelectedRetirementToken] =
@@ -170,6 +172,11 @@ export const Offset = (props: Props) => {
 
   // effects
   useEffect(() => {
+    // round the value for the edge case where a user inputs a fractional value
+    // when fiat isn't selected to pay. Example (100.5 -> 101)
+    if (paymentMethod === "fiat" && !Number.isInteger(parseFloat(quantity))) {
+      setQuantity(Math.round(Number(quantity)).toString());
+    }
     // if input token changes, force a compatible retirement token
     if (
       !offsetCompatibility[paymentMethod]?.includes(selectedRetirementToken)
@@ -464,6 +471,9 @@ export const Offset = (props: Props) => {
   };
 
   const handleChangeQuantity = (e: ChangeEvent<HTMLInputElement>) => {
+    // reset the state on each change event
+    setIsInvalidFractionalQtyFiat(false);
+
     if (debounceTimerRef.current) {
       clearTimeout(debounceTimerRef.current);
     }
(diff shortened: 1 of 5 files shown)


app/components/views/Offset/index.tsx
```diff
@@ -610,19 +620,41 @@ export const Offset = (props: Props) => {
...
             </div>
+
+            {isInvalidFractionalQtyFiat && paymentMethod === "fiat" && (
+              <Text
+                t="caption"
+                color="lightest"
+                className="invalid_project_tonnage"
+              >
+                <Trans id="offset.invalid_fractional_quantity_fiat">
+                  Cannot use a fractional quantity when paying with credit card!
```
0xAeterno, line 654: I wonder if we can come up with a more user friendly error messaging here.
"Fractional" isn't immediately clear to me.

@ShimonD-KlimaDAO You're good with this kind of stuff. Could you please give us a hand?

Using the error message format for the [pledge form](https://github.com/KlimaDAO/klimadao/blob/staging/site/components/pages/Pledge/lib/formSchema.ts) as a reference.
```
Enter a whole number when paying with a credit card
```

app/components/views/Offset/index.tsx
```diff
@@ -93,6 +93,8 @@ export const Offset = (props: Props) => {
   // local state
   const [isRetireTokenModalOpen, setRetireTokenModalOpen] = useState(false);
   const [isInputTokenModalOpen, setInputTokenModalOpen] = useState(false);
+  const [isInvalidFractionalQtyFiat, setIsInvalidFractionalQtyFiat] =
+    useState<boolean>(false);
```
0xAeterno, line 97:
```suggestion
    useState(false);
```
nit: typescript can infer the type here :)

app/components/views/Offset/index.tsx
```diff
@@ -170,6 +172,11 @@ export const Offset = (props: Props) => {
 
   // effects
   useEffect(() => {
+    // round the value for the edge case where a user inputs a fractional value
+    // when fiat isn't selected to pay. Example (100.5 -> 101)
+    if (paymentMethod === "fiat" && !Number.isInteger(parseFloat(quantity))) {
+      setQuantity(Math.round(Number(quantity)).toString());
+    }
```
0xAeterno, line 179: What's the thinking behind this bit here?

Thinking out loud, is this to cover cases where a fractional input value could be:
- copy pasted into the input
- when a user switches payment methods
- guard/validate against url params

0xAeterno (commented):
Awesome first PR @0xMakka :) Great to have you with us!
I haven't tested it in the preview build yet but everything looks great code wise.

Going to wait for @ShimonD-KlimaDAO for advice around the error messaging before I give the ✅✅✅

app/components/views/Offset/index.tsx
0xMakka: @0xAeterno Firstly, thank you for the feedback and warm welcome. It's great to be here and I'm excited to be able to contribute  :)

Agreed on this, I wasn't sure what the error message should be here - but your suggestion is much better than what I have currently there. Will make the change once we get confirmation from @ShimonD-KlimaDAO. :)

app/components/views/Offset/index.tsx
0xMakka: > when a user switches payment methods

Yep as above:

-> A user wishes to pay in USDC so they select USDC from the payment method dropdown 
-> They enter 100.55 in the quantity input, they change their mind and want to pay by credit card.
-> Because by default we don't reset the input field if the payment method is changed, the value is now 100.55 in the input field. Because nothing changed in the input field, an onChange/onKeyDown event is never triggered so the validation doesn't kick in to prevent decimal values.
-> There is a number of ways I guess this could be handled, for example -> clear the input field completely, round the value (which is what I am doing here) or we could attempt to somehow trigger an onChange event on the input field to trigger the validation.

Atmosfearful, line 179: we can be more concise since setting state is cheap
```suggestion
    if (paymentMethod === "fiat") {
      setQuantity(Math.ceil(quantity).toString());
    }
```

also I'd go with `ceil` because, with carbon offsetting, you are generally wanting to cover **all** of a given emissions quantity.

edit: I think doing it this way will actually solve the query params issue. If so, please document that in a comment here.

Atmosfearful (changes requested):
@0xAeterno made a good callout that I would have totally missed: This doesn't seem to catch copy pasting or our query parameters.

example query param:
`https://klimadao-app-git-bugfix-818-no-fractional-quant-783305-klimadao.vercel.app/#/offset?quantity=1.123`

So I think the best way to do this is:
1. handle it all in the `handleChangeQuantity` method, similar to how we handle the minimum quantities. Just round it up via `Math.ceil` before setting to state.
2. The error message is helpful and clever but it might get messy now dealing with copy paste and query params. So to reduce code complexity lets remove that logic/state and instead just update the helper text below the input (it is only visible for credit cards). I would update the string to say:
`Minimum 1-tonne purchase. Whole integers only.`

Another thing:
During review I saw an old outdated line that can now be updated:
https://github.com/KlimaDAO/klimadao/blob/f192a8f4259c56a190cae22ebae9bd760af160b5/app/components/views/Offset/index.tsx#L323

app/components/views/Offset/index.tsx
0xMakka: Thanks for all the suggestions, updated the PR and addressed all comments. This check covers the case for when we switch payment provider. However, it didn't cover the case for query params as there is another update that occurs after this update in an effect that handles query params specifically, so I've added a check around this also.

app/components/views/Offset/index.tsx
```diff
@@ -144,8 +144,14 @@ export const Offset = (props: Props) => {
       setProjectAddress(params.projectTokens);
     }
     if (params.quantity) {
-      setQuantity(params.quantity);
-      setDebouncedQuantity(params.quantity);
+      // handles the case where a decimal value for quantity is passed
```
Atmosfearful, line 147: beautiful

Atmosfearful (approved):

PR #893, opened 2023-01-15
Use JsonRpcBatchProvider by default

Atmosfearful:
## Description
This PR enhances `getStaticProvider` to support a `batchRequests` parameter which is `true` by default.
When `batchRequests` is `true`, we use `JsonRpcBatchProvider` instead of the normal `JsonRpcProvider`.

The BatchProvider does a single https request for multiple calls. Magic!

This pr also changes how default providers are instantiated to avoid unnecessary calls to mumbai and ethereum rpcs.

## Experiments
Based on these experiments I decided it is worth it to use `batchRequests: true` as the default setting, simply to reduce the number of requests (cost savings) with minimal change to actual page performance.

### batching disabled
/#/offset page, while connected, full page refresh with cache disabled:
```
143 requests
522 kB transferred
1.5 MB resources
Finish: 2.04 s
DOMContentLoaded: 364 ms
Load: 592 ms
```

### batch specific requests
As an experiment I tried only batching a few locations where we use Promise.all
```
77 requests
514 kB transferred
1.5 MB resources
Finish: 1.87 s
DOMContentLoaded: 350 ms
Load: 631 ms
```

### batching enabled for all requests
when batchRequests is true by default:
```
59 requests
513 kB transferred
1.5 MB resources
Finish: 2.05 s
DOMContentLoaded: 361 ms
Load: 684 ms
```

## Related Ticket

#892

## Checklist

### DIFF
--- a/app/actions/bonds.ts
+++ b/app/actions/bonds.ts
@@ -180,7 +180,7 @@ export const calcBondDetails = (params: {
   value?: string;
 }): Thunk => {
   return async (dispatch) => {
-    const provider = getStaticProvider({ batchRequests: true });
+    const provider = getStaticProvider();
     let amountInWei;
     if (!params.value || params.value === "") {
       amountInWei = utils.parseEther("0");
@@ -318,7 +318,7 @@ export const calculateUserBondDetails = (params: {
 }): Thunk => {
   return async (dispatch) => {
     if (!params.address) return;
-    const provider = getStaticProvider({ batchRequests: true });
+    const provider = getStaticProvider();
     // inverse bonds dont have user details
     if (getIsInverse(params.bond)) {
       const klimaContract = getContract({
--- a/app/actions/bonds.ts
+++ b/app/actions/bonds.ts
@@ -180,7 +180,7 @@ export const calcBondDetails = (params: {
   value?: string;
 }): Thunk => {
   return async (dispatch) => {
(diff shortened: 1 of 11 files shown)


ladytrekker (approved):
Thanks

app/actions/bonds.ts
```diff
@@ -315,15 +315,15 @@ export const calcBondDetails = (params: {
 export const calculateUserBondDetails = (params: {
   address: string;
   bond: Bond;
-  provider: providers.JsonRpcProvider;
 }): Thunk => {
   return async (dispatch) => {
     if (!params.address) return;
+    const provider = getStaticProvider();
```
ladytrekker, line 321: Can you say again, why you replaced the provider from the useWeb3 hook with the staticProvider again?

Because the contract just needs the "simple" provider?
And to ensure less requests here too?

ladytrekker (approved):
Thanks for the tests results.

Haven't checked the preview link, but code looks good 🍺

app/actions/bonds.ts
Atmosfearful: Just because the user RPC endpoint can be unreliable. So we only really want to use that for outgoing txns (as some users might prefer to set up their own RPC for privacy), but even then I am considering *always* using the static provider even for outgoing...

PR #896, opened 2023-01-16
bugfix(app,site): use dark text on green buttons based on figma files

### LINKED ISSUE #880: Never use white-text-over-Klima-Green

0xMakka:
## Description

Adds dark text to klima green buttons when in dark mode. I've made the changes to buttons based on the figma [file](https://www.figma.com/file/HXwNOLfM9CQHgwmouz0rpk/CARBON-OFFSETTING?node-id=1609%3A1771&t=aGsna9o9vmDyMLay-0). Let me know if anyone thinks I've missed anything.

## Related Ticket

Resolves #880 

## Changes

| Before  | After  |
|---------|--------|
|Screenshot 2023-01-17 at 22 23 20|Screenshot 2023-01-17 at 22 22 12|

| Before  | After  |
|---------|--------|
|Screenshot 2023-01-17 at 22 26 01|Screenshot 2023-01-17 at 22 25 40|

| Before  | After  |
|---------|--------|
|Screenshot 2023-01-17 at 22 36 09|Screenshot 2023-01-17 at 22 36 00|

## Checklist


Atmosfearful (changes requested):
Sorry if the requirements were unclear, instead of overriding each instance, these should be set as the default styles in ButtonPrimary.

https://github.com/KlimaDAO/klimadao/blob/cf932aaeb5bcf128d92b4216ba024e1d008f0dba/lib/components/Buttons/styles.ts#L53

I also realized the white-on-blue button fails a11y guidelines so we need to update the blue variant as well

0xMakka:
> Sorry if the requirements were unclear, instead of overriding each instance, these should be set as the default styles in ButtonPrimary.
> 
> https://github.com/KlimaDAO/klimadao/blob/cf932aaeb5bcf128d92b4216ba024e1d008f0dba/lib/components/Buttons/styles.ts#L53
> 
> I also realized the white-on-blue button fails a11y guidelines so we need to update the blue variant as well

> Ah my bad apologies, I originally had done this but then went through all the figma file and seen there is only some cases where the text color is light and other cases where it is dark and this is why I went with this approach. Just so I have it clear, all klima green and blue variants of the primary button must have the dark text color when in dark mode.

Atmosfearful:
@0xMakka 
>Just so I have it clear, all klima green and blue variants of the primary button must have the dark text color when in "dark mode" only or also when in light mode?
If the background is klima-green or blue then always use the black text color. The goal is to assure that we never violate the color contrast guidelines anywhere in the app. A design debt from early days.

And I apologize that the requirements weren't laid out clearly for you. I should warn you that our figma is frequently out of sync right now so requirements are often unclear 😓 Always reach out when in doubt, and I appreciate your patience if I end up wasting your time!

0xMakka:
> And I apologize that the requirements weren't laid out clearly for you. I should warn you that our figma is frequently out of sync right now so requirements are often unclear 😓 Always reach out when in doubt, and I appreciate your patience if I end up wasting your time!

@Atmosfearful  No problem at all, you aren't wasting my time. Noted :), I made the assumption the figma file was up to date and followed. I've made the suggested updates to the PR now. Blue variant also has black text and the primary button by default, and any buttons that were overwritten with black text when they shouldn't be were addressed also.

Atmosfearful (approved):

PR #898, opened 2023-01-17
Better pledge redirect UX

Atmosfearful:
## Description

This was bothering me for months so I had to jump on it.

- add "connecting" label and disabled state (when connection is still pending)
- add "view your pledge" instead of "create pledge" when the user is already connected
- only show "create pledge" when not connected
- use modal onClose callback to detect when the user cancels the connect modal (reset button state)
- Fix a css issue where disabled button styles were not working for blue buttons,
- Fix a css issue where disabled buttons had an incorrect hover state (text should stay gray and not turn white on hover)

## Changes

Screenshot_20230116_034637
Screenshot_20230116_034608
Screenshot_20230116_034556

## Checklist

### DIFF
--- a/lib/components/Buttons/styles.ts
+++ b/lib/components/Buttons/styles.ts
@@ -65,17 +65,10 @@ export const buttonPrimary = css`
     color: white; /* same in darkmode */
   }
 
-  &:disabled {
-    background-color: var(--surface-03);
-    color: var(--font-03);
-    cursor: not-allowed;
-  }
-
   &.gray {
     background-color: var(--surface-01);
 
     &,
-    &:hover,
     &:hover:not(:disabled),
     &:visited {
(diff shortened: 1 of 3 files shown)


0xemc (approved):
Nice 👍

I'm not sure why but it's many seconds (like > 10) for the view pledge button to successfully navigate.  Long enough that I wasn't sure if it was going to work. Are you experiencing the same thing?

0xAeterno:
> Nice 👍
> 
> I'm not sure why but it's many seconds (like > 10) for the view pledge button to successfully navigate. Long enough that I wasn't sure if it was going to work. Are you experiencing the same thing?

@0xemc I presume it's because the pledge hasn't been cached yet on a preview build, the subgraph queries are expensive af

0xAeterno (approved):
Very naice

PR #899, opened 2023-01-18
Upgrade next, typescript, prettier and eslint

Atmosfearful:
## Description

Latest version of nextjs / typescript and related dev dependencies to prepare for the new package. This creates some prettier formatting changes which is why I need to do it in advance.


Atmosfearful:
preemptively merging as I am running out of time, but I still need @0xemc to review j.i.c.

PR #900, opened 2023-01-18
Goats/modal error handling

### LINKED ISSUE #841: [connect modal] Enhancement: better error handling
description: Currently, when connecting to a wallet fails we only show one message. We should show the user the error from the wallet/provider. Example: when user exits the wallet without connecting it fails. We should show: "User refused connection" or whatever the error says.

AC:
Display error message from wallet and fallback to default error if one does not exist.

goatsOnDrugs:
## Description
The modal now expects an errors object with "default" and "refused" keys. We can add more errors but the message from the provider will now be logged to the console correctly so people can send us helpful console logs if they have errors connecting

## Related Ticket

Resolves #841 

## Checklist

### DIFF
--- a/app/lib/constants.ts
+++ b/app/lib/constants.ts
@@ -1,4 +1,5 @@
 import { urls } from "@klimadao/lib/constants";
+import { t } from "@lingui/macro";
 
 /** True if actually deployed on the production domain (not a preview/staging domain, not local dev) */
 export const IS_PRODUCTION =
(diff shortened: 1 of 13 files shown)


app/lib/constants.ts
```diff
@@ -23,3 +24,14 @@ export const IS_STATIC_EXPORT = process.env.IS_STATIC_EXPORT;
...
   ? "https://checkout.offsetra.com/api/checkout"
   : "https://staging-checkout.offsetra.com/api/checkout";
+
+export const connectErrorStrings = {
+  default: t({
+    message: "We had some trouble connecting. Please try again.",
+    id: "connect_modal.error_message_default",
+  }),
+  rejected: t({
+    message: "User refused connection. Dont do that. Bad user.",
```
Atmosfearful, line 34: lol lets be boring
```suggestion
    message: "User refused connection.",
```

site/lib/constants.ts
```diff
@@ -12,3 +13,14 @@ export const MONTH_IN_SECONDS = 2592000;
...
   ? "http://localhost:3000"
   : urls.home;
+
+export const connectErrorStrings = {
+  default: t({
+    message: "We had some trouble connecting. Please try again.",
+    id: "connect_modal.error_message_default",
+  }),
+  rejected: t({
+    message: "User refused connection. Dont do that. Bad user.",
```
Atmosfearful, line 23:
```suggestion
    message: "User refused connection.",
```

Atmosfearful (changes requested):

app/lib/constants.ts
goatsOnDrugs: updated :)

Atmosfearful (approved):

PR #906, opened 2023-01-20
fix focus trap for connect and enter key listeners

### LINKED ISSUE #830: [connect modal] add a focus trap
We have a hook for this, should be easy

goatsOnDrugs:
## Description
I added tab indexes to the modal content because the focus trap wasn't working. The tabbing is off on the error step and I cant figure out why the tab indexes are getting messed up. I also added a handler for enter button press when buttons are focused.

## Related Ticket

Resolves #830 

## Checklist

### DIFF
--- a/lib/components/ConnectModal/index.tsx
+++ b/lib/components/ConnectModal/index.tsx
@@ -105,15 +105,14 @@ export const ConnectModal = (props: ConnectModalProps) => {
       <div className={styles.modalBackground} onClick={handleClose} />
       <div className={styles.modalContainer}>
         <div className={styles.modalContent} ref={focusTrapRef}>
-          <div className="title">
+          <span className="title">
             <div>
-              <Text t="h4" tabIndex={1}>
-                {getTitle(step)}
-              </Text>
+              <Text t="h4">{getTitle(step)}</Text>
             </div>
-            <button onClick={handleClose} tabIndex={2}>
+            <button onClick={handleClose}>
               <Close fontSize="large" />
             </button>
+<<<<<<< HEAD
           </div>
           {step === "connect" && (
             <div>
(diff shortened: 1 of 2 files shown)


lib/components/ConnectModal/index.tsx
```diff
@@ -100,28 +100,42 @@ export const ConnectModal = (props: ConnectModalProps) => {
       <div className={styles.modalBackground} onClick={handleClose} />
       <div className={styles.modalContainer}>
         <div className={styles.modalContent} ref={focusTrapRef}>
-          <span className="title">
-            <Text t="h4">{getTitle(step)}</Text>
-            <button onClick={handleClose}>
+          <div className="title">
+            <div>
+              <Text t="h4" tabIndex={1}>
+                {getTitle(step)}
```
Atmosfearful, line 106: title doesn't need focus

lib/components/ConnectModal/index.tsx
```diff
@@ -134,13 +148,25 @@ export const ConnectModal = (props: ConnectModalProps) => {
...
+                    handleConnect({
+                      wallet: showCoinbaseWallet ? "injected" : "coinbase",
+                    })
+                  }
+                  tabIndex={5}
                 >
                   <CoinbaseWalletIcon />
                   <Text t="button">Coinbase</Text>
                 </span>
                 <span
```
Atmosfearful, line 162: this is why focusTrap doesn't work-- only anchor tags, buttons and inputs are focusable. The focustrap hook grabs all buttons and injects the listeners for us.

So you can rip out all the keydown and tabindex stuff and just make sure every focusable element is a button

Atmosfearful (changes requested):

lib/components/ConnectModal/index.tsx
goatsOnDrugs: Ok! Yea that makes much more sense than adding tab index. TY!

goatsOnDrugs: updated to use proper html tags and now using css to show/hide elements so that the focus trap works properly when in error step and bc less code. Fun fact: MM doesn't support tabbing.

lib/components/ConnectModal/index.tsx
goatsOnDrugs: I was tired lol. Friday brain fry. Deep fry day.

lib/components/ConnectModal/index.tsx
```diff
@@ -106,100 +107,99 @@ export const ConnectModal = (props: ConnectModalProps) => {
...
+                <button
+                  className={styles.walletButton}
+                  onClick={() => handleConnect({ wallet: "injected" })}
+                >
+                  <BraveIcon />
+                  <Text t="button">Brave</Text>
+                </button>
+              )}
+              {showBrowserWallet && (
                 <span
```
Atmosfearful, line 141: missed one

lib/components/ConnectModal/index.tsx
```diff
@@ -55,6 +55,7 @@ export const ConnectModal = (props: ConnectModalProps) => {
 
   const showBrave = eth?.isBraveWallet;
   const showMetamask = eth?.isMetaMask;
+  const shouldUseCoinbaseSDK = eth?.isCoinbaseWallet;
```
Atmosfearful, line 58: naming is misleading, if isCoinbaseWallet is true it means its injected so we don't need the SDK.

Id go with this:
```suggestion
  const isInjectedCoinbase = eth?.isCoinbaseWallet;
```

lib/components/ConnectModal/index.tsx
```diff
@@ -106,100 +107,99 @@ export const ConnectModal = (props: ConnectModalProps) => {
...
-              </span>
-              <div
-                className={styles.torusButtons}
-                onClick={() => handleConnect({ wallet: "torus" })}
+              )}
+              <button
+                className={styles.walletButton}
+                onClick={() =>
+                  handleConnect({
+                    wallet: shouldUseCoinbaseSDK ? "injected" : "coinbase",
```
Atmosfearful, line 153:
```suggestion
                    wallet: isInjectedCoinbase ? "injected" : "coinbase",
```

Atmosfearful (changes requested):

lib/components/ConnectModal/index.tsx
goatsOnDrugs: updated

lib/components/ConnectModal/index.tsx
goatsOnDrugs: oh good point. updated.

lib/components/ConnectModal/index.tsx
goatsOnDrugs: updated

Atmosfearful (approved):

PR #907, opened 2023-01-23
Fix coinbase

### LINKED ISSUE #702: [App] Coinbase Wallet Connection Bug
Related to #698 
I connected once using CoinbaseWallet, and then disconnected. After that, every time I try to reconnect, I get immediately disconnected automatically. This behavior is reproducible at least across Chrome, Firefox, and Brave.
Screen-capture video is attached.
[Uploading Screencast from 09-14-2022 07:37:14 PM.webm…]()

Atmosfearful:
we broke the class constructer by destructuring like `{ makeWeb3Provider }` rather than saying `coinbaseWallet.makeWeb3Provider`

Resolves #702 and an undocumented coinbase bug

### DIFF
--- a/lib/package.json
+++ b/lib/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "type": "module",
   "dependencies": {
-    "@coinbase/wallet-sdk": "^3.5.4",
+    "@coinbase/wallet-sdk": "^3.6.3",
     "@emotion/css": "^11.10.5",
     "@emotion/react": "^11.10.5",
     "@emotion/styled": "^11.10.5",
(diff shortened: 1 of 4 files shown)


goatsOnDrugs:
good catch! How did you figure that out? Trial and error bc you knew the class was broken?

goatsOnDrugs (approved):
good catch! How did you figure that out? Trial and error bc you knew the class was broken?

Atmosfearful:
> 

blind trial and error! I first messed with the async import, then tried using default vs property, then somehow ended up refactoring that line and narrowed it down.

PR #912, opened 2023-01-25
add rss id to podcast schema

### LINKED ISSUE #908: [resources] Link to rss.com rather than embedding spotify

# Why
- fix #890 safari bug

# Requirements

# Implementation

0xMakka:
## Description

This PR adds a new field to the podcast schema called `rssId`. This is the first PR of 3 as suggested in #908 and will need to be deployed to sanity studio.

## Related Ticket

Resolves #908 

## Checklist

### DIFF
--- a/cms/schemas/podcast.js
+++ b/cms/schemas/podcast.js
@@ -22,6 +22,8 @@ export default {
     {
       name: "rssId",
       title: "RSS Id",
+      description:
+        "6-digit ID used to construct the URL to this podcast episode on rss.com",
       type: "string",
     },
     {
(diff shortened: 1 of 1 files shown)


cms/schemas/podcast.js
```diff
@@ -19,6 +19,11 @@ export default {
       },
       validation: (Rule) => Rule.required(),
     },
+    {
+      name: "rssId",
```
Atmosfearful, line 23:
```suggestion
      name: "rssId",
      description:
        "6-digit ID used to construct the URL to this podcast episode on rss.com",
```

Atmosfearful (changes requested):
figure we can throw a helpful description in there for the content people! Thanks!

cms/schemas/podcast.js
0xMakka: Thanks @Atmosfearful, yeah for sure the description will be helpful - updated.

PR #913, opened 2023-01-25
Makka/908 add rss to card

### LINKED ISSUE #908: [resources] Link to rss.com rather than embedding spotify

0xMakka:
## Description

This is a follow up PR from #912. In this PR, the rssId field is added, all queries are updated and the podcast card is updated to match the figma designs.

## Related Ticket

Resolves #908 

## Changes

| Before  | After  |
|---------|--------|
|Screenshot 2023-01-25 at 23 09 57|Screenshot 2023-01-25 at 23 09 32|

Note: if rssId is null - the play button is hidden (hence why the button on the TOKEN2049 card in the above screenshot is missing) - not sure what the expected behaviour should be for this - disabled or hidden?.

## Checklist


site/components/PodcastCard/index.tsx
```diff
@@ -19,14 +23,19 @@ export function PodcastCard(props: CardProps) {
         <Text t="body2" className="summary">
           {props.podcast.summary}
         </Text>
+        {props.podcast.rssId && (
```
Atmosfearful, line 26: ah good point, after we add all the ids, we should assume that the rss id always exists and then enforce it on the content people so they cant publish without it. I guess thats another cms update follow-up!
```suggestion
        {props.podcast.rssId && (
```

site/components/PodcastCard/index.tsx
```diff
@@ -1,5 +1,9 @@
-import { Text } from "@klimadao/lib/components";
+import Image from "next/legacy/image";
```
Atmosfearful, line 1: we can use the new Next image from now on. legacy ones will need to be migrated so might as well start using the new one!

site/components/PodcastCard/index.tsx
```diff
@@ -19,14 +23,19 @@ export function PodcastCard(props: CardProps) {
         <Text t="body2" className="summary">
           {props.podcast.summary}
         </Text>
+        {props.podcast.rssId && (
+          <A href={`${urls.podcast}/${props.podcast.rssId}/?listen-on=true`}>
```
Atmosfearful, line 27: can you make the whole card clickable actually? Same as how the blog post cards work

Atmosfearful (changes requested):

site/components/PodcastCard/styles.ts
```diff
@@ -25,6 +25,17 @@ export const content = css`
   display: flex;
   flex-direction: column;
   gap: 1.2rem;
+
+  svg {
+    width: 40px;
+    height: 40px;
+    cursor: pointer;
```
Atmosfearful, line 32: always use rem so that all elements scale along with the users browser font size
```suggestion
    width: 4rem;
    height: 4rem;
    cursor: pointer;
```

Atmosfearful (changes requested):

site/components/PodcastCard/index.tsx
0xMakka: Done, updated to the latest on both the Card & PodcastCard.

site/components/PodcastCard/index.tsx
0xMakka: Done

site/components/PodcastCard/styles.ts
0xMakka: Apologies my bad, I took the pixel value from the figma file and forgot to convert - done.

site/components/PodcastCard/index.tsx
0xMakka: Yep, have updated all rssId's inside sanity and have now removed this check. In the next PR which will be removing the embed fields and code, I will enforce this to be a required field.

Atmosfearful (approved):

PR #919, opened 2023-01-31
Goats/784 merge holdings

### LINKED ISSUE #784: [pledge] cleanup - merge holdings data
Sortof related to #783 
follow-up cleanup ticket for #658  per discussion with @goatsOnDrugs after #729 gets released

I think the most efficient and cleanest way to do this is pass over each array, sort, then pass over again to calculate the final values.
```ts
// starting data, two arrays from two addresses sorted by timestamp
const holdings = [
  { timestamp: 1, value: 1 },
  { timestamp: 3, value: 3 },
  { timestamp: 5, value: 0 },
];
const holdings2 = [
  { timestamp: 2, value: 2 },
  { timestamp: 4, value: 4 },
];

// first pass -> convert values to "value-change" events by subtracting each entry value from the previous entry value
const holdings1ch = [
  { timestamp: 1, value: 1, change: 1 }, // no prev, use initial value
  { timestamp: 3, value: 3, change: 2 },
  { timestamp: 5, value: 0, change: -3 }, // 0 - 3 = -3
];
const holdings2ch = [
  { timestamp: 2, value: 2, change: 2 },
  { timestamp: 4, value: 4, change: 2 },
];

// then merge arrays and sort by timestamp, and finally calculate the running total
const sort = [
  { timestamp: 1, change: 1, value: 1 },
  { timestamp: 2, change: 2, value: 3 },
  { timestamp: 3, change: 2, value: 5 },
  { timestamp: 4, change: 2, value: 7 },
  { timestamp: 5, change: -3, value: 4 }, // -3 + 7 = 4
];
```

### LINKED ISSUE #916: [tech] Unit Test Setup
Set up Jest and Testing Library for unit testing site, app, and lib

goatsOnDrugs:
## Description
merge holdings for chart. merging balances in a separate PR. This PR also adds unit tests with Jest and testing library. We decided to co-locate tests.

## Related Ticket

Resolves #784 
Resolves #916  

## Checklist

### DIFF
--- a/site/lib/mergeHoldings.test.ts
+++ b/site/lib/mergeHoldings.test.ts
@@ -1,406 +1,94 @@
 import { Holding } from "components/pages/Pledge/types";
+import { Dictionary } from "lodash";
 import {
   calculateChange,
   calculateTokenValues,
   groupHoldings,
   mergeHoldings,
 } from "./mergeHoldings";
+import {
+  calculatedGroupedHoldingsDecreasing,
+  calculatedGroupedHoldingsIncreasing,
+  expectedOutputDecreasing,
+  expectedOutputIncreasing,
+  expectedOutputLargeNumber,
+  groupedHoldingsDecreasing,
+  groupedHoldingsIncreasing,
+  holdingsInputDecreasing,
+  holdingsInputIncreasing,
+  holdingsInputLargeNumber,
+} from "./testFixtures";
 
-const holdingsInput: Holding[][] = [
-  [
-    {
-      id: "0xdeb8c24ad9640d62334da6f30c191e9cfd180e5b-BCT-1674777600",
-      timestamp: "1",
-      token: "BCT",
-      tokenAmount: "1",
-      carbonValue: "1",
-    },
-    {
-      id: "0xdeb8c24ad9640d62334da6f30c191e9cfd180e5b-BCT-1674777600",
-      timestamp: "3",
-      token: "BCT",
-      tokenAmount: "2",
-      carbonValue: "2",
-    },
-    {
-      id: "0xdeb8c24ad9640d62334da6f30c191e9cfd180e5b-BCT-1674777600",
-      timestamp: "5",
-      token: "BCT",
-      tokenAmount: "3",
-      carbonValue: "3",
-    },
-    {
-      id: "0xdeb8c24ad9640d62334da6f30c191e9cfd180e5b-NCT-1674777600",
-      timestamp: "7",
-      token: "NCT",
-      tokenAmount: "1",
-      carbonValue: "1",
-    },
-    {
-      id: "0xdeb8c24ad9640d62334da6f30c191e9cfd180e5b-NCT-1674777600",
-      timestamp: "9",
-      token: "NCT",
-      tokenAmount: "2",
-      carbonValue: "2",
-    },
-    {
-      id: "0xdeb8c24ad9640d62334da6f30c191e9cfd180e5b-NCT-1674777600",
-      timestamp: "11",
-      token: "NCT",
-      tokenAmount: "3",
-      carbonValue: "3",
-    },
-  ],
-  [
-    {
-      id: "0x838cd36c38e64c166b2c5130d15a8acff1e09d44-BCT-1674777600",
-      timestamp: "4",
-      token: "BCT",
-      tokenAmount: "2",
-      carbonValue: "2",
-    },
-    {
-      id: "0x838cd36c38e64c166b2c5130d15a8acff1e09d44-BCT-1674777600",
-      timestamp: "6",
-      token: "BCT",
-      tokenAmount: "3",
-      carbonValue: "3",
-    },
-    {
-      id: "0x838cd36c38e64c166b2c5130d15a8acff1e09d44-NCT-1674777600",
-      timestamp: "8",
-      token: "NCT",
-      tokenAmount: "1",
-      carbonValue: "1",
-    },
-    {
-      id: "0x838cd36c38e64c166b2c5130d15a8acff1e09d44-NCT-1674777600",
-      timestamp: "10",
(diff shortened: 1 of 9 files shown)


site/lib/mergeHoldings.test.ts
```diff
@@ -0,0 +1,406 @@
...
+      carbonValue: "7",
+      id: "0x838cd36c38e64c166b2c5130d15a8acff1e09d44-NCT-1674777600",
+      timestamp: "14",
+      token: "NCT",
+      tokenAmount: "7",
+    },
+  ],
+};
+
+export const holdingsTest = test("setup", () => {
```
Atmosfearful, line 397: I've always done it with the `describe` and `test` pattern like this:

1. top level `describe` for each utility
2. inside there, multiple `tests`, one test for each of the expected behaviors/outcomes/edge cases
3. extra descriptive test names and comments so that newcomers can look at a test and understand all the requirements

```js
describe('groupHoldings', () => {

  test('Groups 2 holdings arrays of BCT and NCT', () => {
    expect(groupHoldings(holdingsInput)).toStrictEqual(groupedHoldings);
  });
  
  // test('other edge case', ...
});

describe('calculateChange', () => {

  test('Determine the balance change value for each entry', () => {
     expect(calculateChange(groupedHoldings)).toStrictEqual(
      calculatedGroupedHoldings
    );
  });
});

// etc
```

site/pages/pledge/[address].tsx
```diff
@@ -94,61 +103,83 @@ export const getStaticProps: GetStaticProps<PageProps, Params> = async (
...
+
+      // Holdings
+      const holdingsPromises = verifiedWallets.map((wallet) =>
+        queryHoldingsByAddress(wallet.address)
       );
-      retirements = reduced;
+      holdingsPromises.push(queryHoldingsByAddress(resolvedAddress));
+      // starting with array of arrays
+      const holdingsValues = await Promise.all(holdingsPromises);
+      mergeHoldings(holdingsValues);
```
Comment (Atmosfearful) [inline site/pages/pledge/[address].tsx:175]: extra
```suggestion

```

site/lib/mergeHoldings.ts
```diff
@@ -0,0 +1,77 @@
...
+import { Dictionary, groupBy } from "lodash";
+
+export const mergeHoldings = (holdings: Holding[][]) => {
+  const groupedHoldings = groupHoldings(holdings);
+  const calculatedChangeHoldings = calculateChange(groupedHoldings);
+  const mergedHoldings = calculateTokenValues(calculatedChangeHoldings);
+  return mergedHoldings;
+};
+
+export const groupHoldings = (holdings: Holding[][]) =>
```
Atmosfearful, line 11: can we get some comments on each util for newcomers and dumdums to outline the "why" of all this code

site/lib/mergeHoldings.test.ts
```diff
@@ -0,0 +1,406 @@
...
+  [
+    {
+      id: "0xdeb8c24ad9640d62334da6f30c191e9cfd180e5b-BCT-1674777600",
+      timestamp: "1",
+      token: "BCT",
+      tokenAmount: "1",
+      carbonValue: "1",
+    },
+    {
+      id: "0xdeb8c24ad9640d62334da6f30c191e9cfd180e5b-BCT-1674777600",
```
Atmosfearful, line 19: we actually don't care about `id` or `carbonValue` so I would remove it from the fixtures for readability. You can do typescript asserts if needed (avoid `any` though b.c. our linter hates it)
```ts
mergeHoldings(myFixture as Holding[])
```

site/lib/mergeHoldings.test.ts
```diff
@@ -0,0 +1,406 @@
...
+    },
+  ],
+};
+
+export const holdingsTest = test("setup", () => {
+  expect(groupHoldings(holdingsInput)).toStrictEqual(groupedHoldings);
+  expect(calculateChange(groupedHoldings)).toStrictEqual(
+    calculatedGroupedHoldings
+  );
+  expect(calculateTokenValues(calculatedGroupedHoldings)).toStrictEqual(
```
Atmosfearful, line 402: * [ ] we need a test for this util for balances that go down

* [ ] also I'd like to make sure it doesn't explode with big decimals. BCT has 18 decimals which can cause overflows and weirdness. So try doing another test with a fixture that has values like:
`0.111111111111111111`
plus `0.111111111111111111`
equals `0.222222222222222222`
if it causes an overflow diget thats ok, because its only used in the chart, we can at least document the quirky math with the test. or if you want to get fancy use `ethers.BigNumber.from(value).sub(othervalue).format()` to do the math...

Atmosfearful (changes requested):

site/lib/mergeHoldings.test.ts
goatsOnDrugs: Oh, right! Its been too long since I wrote a test lol

site/lib/mergeHoldings.ts
goatsOnDrugs: good idea, comments added

site/lib/mergeHoldings.test.ts
goatsOnDrugs: @Atmosfearful I actually found the ids useful when looking through the test features because they are the wallet address. This makes it easier to look at the giant arrays and see how the data is being merged. I can trim off the token and timestamp at the end to make it shorter or change it to something like wallet1 and wallet2. carbonValue removed for inputs but `calculateTokenValues` returns carbonValue so test fails if its not included in expected output.

site/lib/mergeHoldings.test.ts
goatsOnDrugs: @Atmosfearful the output of those numbers is 16 digits so I created a test tase to document.

PR #920, opened 2023-01-31
cleanup embed codes

### LINKED ISSUE #908: [resources] Link to rss.com rather than embedding spotify

0xMakka:
## Description

Clean up of embed codes on podcasts which were replaced by rss id's in #908.

## Related Ticket

Resolves #908 

## Checklist

### DIFF
--- a/cms/schemas/podcast.js
+++ b/cms/schemas/podcast.js
@@ -27,11 +27,6 @@ export default {
       type: "string",
     },
     {
-      name: "embedCode",
-      title: "Embed code",
-      type: "string",
-    },
-    {
       name: "publishedAt",
       title: "Published at",
       type: "datetime",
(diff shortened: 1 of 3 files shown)


Atmosfearful (changes requested):
Thanks! We need to remove `embed` property from the Document and Post types in `lib/queries` as well.

Note to self: I need to deploy this after merge

0xMakka:
> Thanks! We need to remove `embed` property from the Document and Post types in `lib/queries` as well.
> 
> Note to self: I need to deploy this after merge

Ah man, done! My find/replace let me down here - searched for "embedCode" and thought I had all covered. Thanks for the feedback and noticing I missed this.

Atmosfearful (approved):

PR #932, opened 2023-02-18
Reduce Retirement Page load times

Atmosfearful:
- Never do reverse domain lookups (saves **~0.5 seconds**)
- Only invoke `getRetirementDetails` (an RPC call) if the subgraph does not have a record yet (saves **~0.6 seconds**)
  - Migrate to v2 contract `getRetirementDetails` and remove the old v1 method
  - Add a fallback UI for when the subgraph does not have a record. This includes an auto-poll and refresh mechanism as soon as the subgraph catches up. This should be a temporary "fix" for #933 which was showing a broken UI because the subgraph failed to index that retirement. Now it should show the fallback UI
- Remove the call to pull Verra project details. This was adding up to **9 seconds** (!!) in the worst case I saw.
  - We only needed verra for the project name. Soon we will have a performant Sanity CMS to give this to us. For now, just fall back to project ID.
  
After this PR the load time is down to about 300ms, the only bottleneck is now the subgraph.

# Additional changes
While in there I snuck in some low hanging fruit cleanups for project details. Small padding and language fixes, and added currentSupply.

# Screenshots
## Default view for verra retirement w/ missing project name
Screenshot_20230218_092133
## Default view for MCO2
Screenshot_20230218_092113
## Fallback ui if the subgraph is slow, desktop
Screenshot_20230217_071244
## Fallback ui if the subgraph is slow, mobile
Screenshot_20230217_071126


site/components/pages/Retirements/SingleRetirement/ProjectDetails/List/index.tsx
```diff
@@ -15,48 +15,63 @@ type Props = {
...
-        <Text>
-          {trimTotalRetired}{" "}
-          <Trans id="retirement.single.project_details.tonnes">Tonnes</Trans>
+        <Text t="caption">
+          <Trans>{trimTotalRetired} Tonnes retired in total</Trans>
+        </Text>
+      )}
+      {trimCurrentSupply && (
+        <Text t="caption">
+          <Trans>{trimCurrentSupply} Tonnes remaining</Trans>
```
ladytrekker, line 62: All Single retirements do not show the values from the retirement  🦐
I think you need to extract the strings again.

And remove all dynamic values out of the `<Trans>` for an easier translator life.

For example [here](https://klimadao-site-git-atmos-page-performance-klimadao.vercel.app/retirements/atmosfearful.klima/1)

Bildschirm­foto 2023-02-20 um 13 10 06

site/components/pages/Retirements/SingleRetirement/index.tsx
```diff
@@ -126,79 +137,96 @@ export const SingleRetirementPage: NextPage<SingleRetirementPageProps> = (
...
-                    href={`/retirements/${
-                      nameserviceDomain || beneficiaryAddress
-                    }`}
-                    className="address"
-                  >
-                    {nameserviceDomain || concatAddress(beneficiaryAddress)}
-                  </Link>
-                }
-              />
+          <RetirementMessage message={retirement.retirementMessage} />
```
ladytrekker, line 140: Message can be empty.
You removed the fallback "No retirement message provided"

See result [here](https://klimadao-site-git-atmos-page-performance-klimadao.vercel.app/en/retirements/0x2d2f9d78a0ed19bceae13bb81daee2825a1c53fc/4)

Bildschirm­foto 2023-02-20 um 13 23 18

ladytrekker (approved):
Thanks, just nitpicks below

PR #937, opened 2023-02-21
Retire your own TCO2 & C3T tokens

### LINKED ISSUE #879: [retirement] Selective retirement option for redeemed tokens (tco2, c3t) user owns
Need to support selective retirement of redeemed tokens (like tco2 or c3t) that a user already has in their wallet.

This is in support of the ability to Redeem specific project tokens #735 

This feature has been covered off in bezos but needs to also be added to the retirement aggregator with timing aligned with rollout of redeem capability.

- [x] TASK#1 need a design review and updated designs to support this feature

Atmosfearful:
## Description
No point in reviewing the code it will all get nuked when we migrate to the new ui.

I had to remove the "breakdown by asset" for retirements on the /retirements page and on the /offset view of the dapp. It's too expensive to query every TCO2 and pool token to check balances. Let's see if anyone notices...

Pledges do not support TCO2 / C3T yet in the breakdowns. But the total tonnes should still reflect tco2 retirements.

Screenshot_20230222_064114
Screenshot_20230222_033643
image

Resolves #879


ShimonD-KlimaDAO:
The retirement certificate, `/retirements` page, and Pledge are all looking good for both c3t and tco2. Just a few minor issue and one minor suggestion below. 

1. It would be nice if the assets were sorted by pool type, grouping all tco2s and c3ts separately.
image

2. "TONNES AVAILABLE" is incorrectly showing 0 for every asset I have for both tco2s and c3ts.
image

3. I am able to input an amount greater than the amount I own (2 tonnes) into the amount field, and I am able to click "Approve" and submit an Approve [transaction](https://polygonscan.com/tx/0x14c8e57c5eb703d9df932d4c3d64e35d673bb0742d5258dd144867592203d1b2). This should be captured and the field should turn red. I then get a submit transaction which fails since I don't have enough tonnage.
image
image

 This is what it looks like when we try to do the same using the normal offset process:
image

PR #947, opened 2023-03-02
Selective redemption

### LINKED ISSUE #735: Retirement Aggregator Redeem/Retire Support
[Related roadmap card](https://www.notion.so/klima-dao/Better-Redemption-Experience-ffb1a8162bda46adb4d26cebf569f079)

This workstream is related to the idea of supporting in our applications the ability to redeem carbon reference pool tokens (i.e. BCT, NCT, UBO) for specific project fractionalized tokens in their respective pools (for example, TCO2) and then retire those project tokens.

_Confirmed that V2 of the aggregator already supports redemption. @cujowolf  is just writing tests, so we should be ready to use that pretty soon to implement a simple redemption frontend view._

- [x] TASK#1 - Retirement aggregator **design changes** to support (1) Show the default project being retired for a basic offset (i.e. not advanced) (2) Redeem pool token & retire specific project tokens

For **(1) Show the default project being retired for a basic offset** (i.e. not advanced)...further context that came from a [discord post](https://discord.com/channels/841390338324824096/967377329171099648/988542905792741427) from @Atmosfearful :
> - In the offset ui, we completely abstract away the concept of a "carbon pool" and instead show the *default project* for each. For example, we know that the next 470k tonnes of BCT retired will be VCS-191.
> - Instead of choosing from a completely meaningless list of icons and letters BCT|MCO2|UBO|NBO they can see a list of projects and USDC denominated prices

Atmosfearful:
Acquire carbon using USDC, KLIMA, or a pool token!

Closes #735

### DIFF
--- a/app/actions/redeem.ts
+++ b/app/actions/redeem.ts
@@ -97,7 +97,6 @@ export const redeemCarbonTransaction = async (params: {
 
     params.onStatus("networkConfirmation");
     await txn.wait(1);
-
     params.onStatus("done");
   } catch (e: any) {
     if (e.code === 4001) {
--- a/app/actions/redeem.ts
+++ b/app/actions/redeem.ts
@@ -1,0 +1,134 @@
+import { addresses } from "@klimadao/lib/constants";
+import { AllowancesFormatted } from "@klimadao/lib/types/allowances";
+import {
+  getAllowance,
+  getContract,
+  getStaticProvider,
+  getTokenDecimals,
+  safeAdd,
+} from "@klimadao/lib/utils";
+import { providers, utils } from "ethers";
+import { formatUnits } from "ethers/lib/utils";
+import {
+  RedeemablePoolToken,
+  RedeemPaymentMethod,
+  redeemPaymentMethods,
+} from "lib/hooks/useRedeemParams";
+import { Thunk } from "state";
+import { updateAllowances } from "state/user";
+import { OnStatusHandler } from "./utils";
+
+export const getRedeemAllowances = (params: {
+  address: string;
+  onRPCError: () => void;
+}): Thunk => {
+  return async (dispatch) => {
+    try {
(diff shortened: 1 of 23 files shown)


ShimonD-KlimaDAO:
1. Paying for an NBO project is possible with BCT, but then the NBO field shows that the token is incompatible.
image
The transaction goes through regardless, since every token has a valid path with the RA v2. In this example, the path is BCT->USDC->KLIMA->NBO.
image

2. When "Payment Method" is set to USDC, I can select any pool. I can then change USDC to several options that allow me to do a transaction that's supposed to be incompatible according to the UI.
 a. This should not be possible to do with "incompatible" pairs.
 b. The pairs are not actually incompatible using v2, as described in (1) above.
 c. KLIMA is not an option I can pay with for any token, which is  not intuitive given that the KLIMA token is actually the main liquidity pair with all the carbon pool tokens.
 
 3. I am able to input a tco2 address which isn't part of the NCT pool (it is part of the BCT pool). This is just an example, and points to the need to disallow this at the frontend.
image
image
When I then try to execute the transaction, it can't estimate gas because the pool passed to the contract function as an argument is incompatible with the tco2, so gas estimation fails.
image

4. We recently got rid of balances on the right side in the Offset tab. Do we need them here?
image

Atmosfearful:
@ShimonD-KlimaDAO 
1. I made all tokens compatible with eachother. I didn't know that V2 supported this! Nice!
2. No longer an issue now that all are compatible. I tested with KLIMA to NBO and it worked.
3. This one is too complicated to fix at the moment, requires a separate query & state to manage. We can address this when we migrate these journeys to the new product.
4. Indeed I added balances back because it's important to see your new redeemed tokens appear!

Will pass back to you @ShimonD-KlimaDAO and @jabby09 for a final run through?

PR #951, opened 2023-03-13
Fix USDC decimals in Offset view

cujowolf:
We were doing unsafe math in getApproval in the Offset view.


cujowolf:
@Atmosfearful is sitting next to me and approved this PR and is actually typing this message right now.

Issue #681, opened 2022-09-05
Migrate to offset aggregator v2


Atmosfearful:
@cujowolf please document the main changes to the methods or behavior that we should be aware of
- changes to input parameters?
- changes to outputs?
- changes to approval flow or anything else?

* [x] give contract to atmos via dm or post here
* [x] give ABI json to atmos via dm or post here or open a PR to add it to our /abi folder :)

Atmosfearful:
Cujo will first route v1 aggregator to the v2 when it is running. After that, we can migrate the ABI at our leisure

ShimonD-KlimaDAO:
Left comments in Cujo's private repo. Waiting to test deployed contracts on mainnet after my comments have been addressed.

ShimonD-KlimaDAO:
@Atmosfearful The following, along with [this discord message](https://discord.com/channels/841390338324824096/964697951542837258/1052804653223522344) should be everything you need to get you going in parallel while I carefully document in gitbook.

- changes to input parameters?
**Yes, the function names and input parameters have changed. linked discord message for now to work in parallel.**

- changes to outputs?
**Yes, as requested, every retirement function now returns the retirement index.**

- changes to approval flow or anything else?
**No. The Diamond contract (main contract address) needs approval to spend source token, same as before.**


This link is very helpful for a general overview of the function names and inputs as well, but you shouldn't need it:
https://louper.dev/diamond/0x8ce54d9625371fb2a068986d32c85de8e6e995f8?network=polygon


Contract address: `0x8cE54d9625371fb2a068986d32C85De8E6e995f8`
Contract ABI:
```
[...]

PR #960, opened 2023-03-29
RetirementAggregatorV2 upgrade

psparacino:
## Description

All instances of retirementAggregator have been removed and replaced with retirementAggregatorV2.

The frontend can successfully call specific and default retirements.

## Related Ticket

[Issue #681 ](https://github.com/KlimaDAO/klimadao/issues/681)

## Reviewers Requested

@Atmosfearful @ShimonD-KlimaDAO 

## Checklist


app/actions/offset.ts
```diff
@@ -115,24 +113,23 @@ export const getOffsetConsumptionCost = async (params: {
...
+        addresses["mainnet"][params.retirementToken],
+        parsed
+      );
   }
-
   return [
-    formatUnits(sourceAmount[0], getTokenDecimals(params.inputToken)),
-    formatUnits(sourceAmount[1], getTokenDecimals(params.retirementToken)),
+    formatUnits(sourceAmount, getTokenDecimals(params.inputToken)),
+    // formatUnits(sourceAmount[1], getTokenDecimals(params.retirementToken)),
```
Atmosfearful, line 132: stray comment here

lib/constants/index.ts
```diff
@@ -187,18 +185,17 @@ export type Bond = (typeof bonds)[number];
...
     "nct",
-    "mco2",
     "usdc",
     "klima",
+    "mco2",
     "sklima",
     "wsklima",
-  ],
-  retirementAggregatorV2: ["ubo", "nbo", "bct", "nct", "usdc", "klima"], // redeemables
+  ], // redeemables
```
Atmosfearful, line 198:
```suggestion
  ],
```

Atmosfearful (changes requested):

app/actions/offset.ts
```diff
@@ -169,39 +172,62 @@ export const retireCarbonTransaction = async (params: {
...
+      );
+    }
     let txn;
     if (!!params.projectAddress) {
-      txn = await retireContract.retireCarbonSpecific(
+      txn = await retireContract.retireExactCarbonSpecific(
         addresses["mainnet"][params.inputToken],
         addresses["mainnet"][params.retirementToken],
+        params.projectAddress,
+        sourceAmount,
```
Atmosfearful, line 206: Assuming this parameter, `maxAmountIn` is meant to prevent slippage, i think we should pass in the value that we already approved for spend, which should happen before we invoke this method. I asked cujo/shimon if my reasoning is correct here

Atmosfearful (changes requested):

app/actions/offset.ts
psparacino: so just retrieve the allowance for that particular token and use that for `maxAmountIn`? I was working off the docs, which said to use the `getSourceAmount... `functions for the maxAmountIn param:

> Maximum amount of the source token to spend. To find the exact amount needed, please refer to the view functions getSourceAmountSpecificRetirement and getSourceAmountDefaultRetirement

Atmosfearful (approved):

Atmosfearful:
Hold off on merging until we fix #963

ShimonD-KlimaDAO:
I was able to execute default/specific retirements using Klima  for UBO and NBO. I then tried to retire BCT using Klima, and was not prompted to approve the new Klima allowance amount to the contract. This results in gas estimation failure, as expected:
image

The question is - why was I not prompted to approve this amount before the retirement transaction? When I try to switch the "pay with" to usdc, I get the Approve prompt. When I then try to switch back "Pay with" to Klima, I get "Submit" right away. It's as if the frontend remembers that I already approved the contract to spend Klima, but doesn't re-check allowances.

psparacino:
> I was able to execute default/specific retirements using Klima for UBO and NBO. I then tried to retire BCT using Klima, and was not prompted to approve the new Klima allowance amount to the contract. This results in gas estimation failure, as expected: image
> 
> The question is - why was I not prompted to approve this amount before the retirement transaction? When I try to switch the "pay with" to usdc, I get the Approve prompt. When I then try to switch back "Pay with" to Klima, I get "Submit" right away. It's as if the frontend remembers that I already approved the contract to spend Klima, but doesn't re-check allowances.

Got it; I believe I found where the error was happening. At the risk of too many words, outlining the flow as it currently stands as I think this should be correct and work:

1) Get the retirement cost from`getOffsetConsumptionCost` (which uses either `getSourceAmountSpecificRetirement `or `getSourceAmountDefaultRetirement`)
**app/actions/offset line 100**

2) The approval flow adds 1% to the cost for slippage and approves that amount if the current allowance is insufficient.
`handleApprove` and` getApprovalValue` in **app/components/views/Offset/index.tsx**

3) Get the allowance of the payment method (i.e. inputToken) to pass in for maxAmountIn
https://github.com/KlimaDAO/klimadao/pull/960/files#diff-b68c42fa04a583195c0f40a15e663dcebdd59c469a31bac00a34e1848e2a6c06R355

4) format amount to be able to pass to contract
https://github.com/KlimaDAO/klimadao/pull/960/files#diff-6ad5ddf9bd7983af4e1ccb760b14f9e695320de4784621fa1288e84170087aaeR182-R186

app/components/views/Offset/index.tsx
```diff
@@ -352,11 +352,12 @@ export const Offset = (props: Props) => {
       } else if (paymentMethod === "fiat") {
         return; // type guard
       } else {
+        const maxAmountIn = allowances?.[paymentMethod];
```
Atmosfearful, line 355: Unfortunately we can't rely on this because some people might have infinite allowances from early days :(

Need to rely on the quoted cost. I'm having analysis paralysis about adding slippage buffer or not, let's go with No and see if that causes problems in the wild (It didn't during my testing)

Atmosfearful (changes requested):

ShimonD-KlimaDAO:
Works well!

Atmosfearful (approved):

PR #1028, opened 2023-04-14
Resolve type discrepancies

### LINKED ISSUE #1007: [tech-debt] Update API / json types to string instead of BigNumber
BigNumber is not a seriealizable json, in reality everything we receive from the backend is a `string`. We need to use jsdoc comments to make it clear which values have already been converted to float strings on the backend, and which ones are still sent as bignumber integer strings.

Atmosfearful:
## Description

I looked at the real JSON data on prod and aligned the types.
I separated the activity types into UserActivity and ProjectActivity because they differ between these endpoints
I should have separated the Listing types as well, but it was faster to just use `Omit<Listing, "project">` where needed.
Then I addressed all the downstream issues that were exposed.

Some TODOs on the backend still, which I marked in the carbonmark types file, and will direct najadas attention to.

## Related Ticket
Closes #1007

### DIFF
--- a/carbonmark/components/pages/Project/Purchase/index.tsx
+++ b/carbonmark/components/pages/Project/Purchase/index.tsx
@@ -27,12 +27,12 @@ import { useState } from "react";
 import { FormValues, PurchaseForm } from "./PurchaseForm";
 import * as styles from "./styles";
 
-type Props = {
+export interface ProjectPurchasePageProps {
   project: Project;
-  listing: Listing;
-};
+  listing: Omit<Listing, "project">;
+}
 
-export const ProjectPurchase: NextPage<Props> = (props) => {
+export const ProjectPurchase: NextPage<ProjectPurchasePageProps> = (props) => {
   const { locale, push } = useRouter();
   const { address, provider } = useWeb3();
   const [isLoadingAllowance, setIsLoadingAllowance] = useState(false);
--- a/carbonmark/components/pages/Project/Purchase/index.tsx
+++ b/carbonmark/components/pages/Project/Purchase/index.tsx
@@ -29,7 +29,7 @@ import * as styles from "./styles";
 
 export interface ProjectPurchasePageProps {
   project: Project;
-  listing: Omit<Listing, "project">;
+  listing: Listing;
 }
 
 export const ProjectPurchase: NextPage<ProjectPurchasePageProps> = (props) => {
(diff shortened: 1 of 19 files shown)


carbonmark/lib/actions.ts
```diff
@@ -294,9 +294,9 @@ export const getProjectInfoFromApi = async (
       key: project.key,
       projectID: project.projectID,
       name: project.name,
-      methodology: project.methodology,
+      methodology: project.methodologies[0].id,
       vintage: project.vintage,
-      category: project.category?.id || "Other",
+      category: project.methodologies[0].category || "Other",
```
ladytrekker, line 299: To reduce boilerplate and to have deep object constructions in one place
you can use the getter `getCategoryFromProject`  everywhere where you use 
`!!project.methodologies[0]?.category && project.methodologies[0].category` 

And create a second getter `getCategoryIdFromProject` to retrieve the category ID too.
`project.methodologies[0].id`

carbonmark/components/pages/Project/Purchase/index.tsx
```diff
@@ -27,12 +27,12 @@ import { useState } from "react";
 import { FormValues, PurchaseForm } from "./PurchaseForm";
 import * as styles from "./styles";
 
-type Props = {
+export interface ProjectPurchasePageProps {
   project: Project;
-  listing: Listing;
-};
+  listing: Omit<Listing, "project">;
```
ladytrekker, line 32: I know you mentioned it, but still as you are using the `Omit` in multiple places, why not create a dedicated type after all?

carbonmark/components/pages/Project/index.tsx
```diff
@@ -50,9 +51,11 @@ const Page: NextPage<PageProps> = (props) => {
   }
 
   const allListings =
-    Array.isArray(project.listings) && getAllListings(project.listings);
+    Array.isArray(project.listings) &&
+    getAllListings(project.listings as Listing[]);
```
ladytrekker, line 55: When you give the `listingsGetters`like
- getActiveListings
- getAllListings
- getSortByUpdateListings
- ....
the correct type `Listing without Project` then you can remove the assertions here, right?

ladytrekker (approved):
Thanks, some remarks below.

I think when you create two separated types for the "Listings with or without Project" you can get rid of some assertion.

carbonmark/components/pages/Project/Purchase/index.tsx
Atmosfearful: ok done

carbonmark/components/pages/Project/index.tsx
Atmosfearful: cleaned it all up thx!

PR #1064, opened 2023-04-20
#988 Show handle names

### LINKED ISSUE #988: [activities] Show Handle name
## Description

Where should the hyperlink from the seller/buyer addresses go to?
=> Always go to carbonmark.

It may be the case that they don't have a profile created yet, but at least then you can explore their other past activity and see their balances.

~No kns/ens lookup, this is not super relevant for carbonmark users now that we have handles (only klima / defi nerds)~

But showing the handle would be nice, if the API gives it to us. Not a must-have for v1.

## TODO
- [x] check with API dev if handle can be added to the data
- [x] or look it up on the client
- [ ] if buyer/seller is a carbonmark user => show handle instead of truncated address

psparacino:
## Description

Shows user handle name where possible instead of the wallet address.

The only place this is currently not possible is where the type is UserActivity, which looks to be on the user page. There is a note from two days ago that this is being added to the api response, so sounds like this will be a quick fix to add once that is done. 

Also in this PR:

1) small chaining issue in Project Getter

2) fixed some error types so the linter is satisfied.

Reviewers:
@Atmosfearful 
@ShimonD-KlimaDAO 

## Related Ticket

Resolves #988 

## Changes

| Before  | After  |
|---------|--------|
|before1 |after1|

| Before  | After  |
|---------|--------|
| before |after|

## Checklist

### DIFF
--- a/carbonmark/components/Activities/Activity.tsx
+++ b/carbonmark/components/Activities/Activity.tsx
@@ -119,7 +119,7 @@ export const Activity = (props: Props) => {
       <Text t="body1">
         {!!addressA && (
           <Link className="account" href={`/users/${addressA}`}>
-            {formatWalletAddress(addressA, connectedAddress)}{" "}
+            {addressA.length >= 11 ? formatWalletAddress(addressA, connectedAddress) : addressA}{" "}
           </Link>
         )}
 
@@ -128,7 +128,7 @@ export const Activity = (props: Props) => {
         {addressB && (
           <Link className="account" href={`/users/${addressB}`}>
             {" "}
-            {formatWalletAddress(addressB, connectedAddress)}
+            {addressB.length >= 11 ? formatWalletAddress(addressB, connectedAddress) : addressB}
           </Link>
         )}
       </Text>
--- a/carbonmark/components/Activities/Activity.tsx
+++ b/carbonmark/components/Activities/Activity.tsx
@@ -119,7 +119,9 @@ export const Activity = (props: Props) => {
       <Text t="body1">
         {!!addressA && (
           <Link className="account" href={`/users/${addressA}`}>
-            {addressA.length >= 11 ? formatWalletAddress(addressA, connectedAddress) : addressA}{" "}
+            {addressA.length >= 11
+              ? formatWalletAddress(addressA, connectedAddress)
+              : addressA}{" "}
           </Link>
         )}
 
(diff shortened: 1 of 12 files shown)


jabby09:
@psparacino @Atmosfearful @ShimonD-KlimaDAO 

Now that we are seeing this with user names, the ... styling I don't think is working for user names shorter than the space allowed. See `jabby...bby` and `najad...ada` as example.

It makes sense to support this format for 0x addresses (which are the fall back) but looks like we need a condition for checking if the handle name will will as is in the space allowed.

Screenshot 2023-04-21 at 1 01 06 PM

jabby09:
> The only place this is currently not possible is where the type is UserActivity, which looks to be on the user page. There is a note from two days ago that this is being added to the api response, so sounds like this will be a quick fix to add once that is done.

@psparacino @Atmosfearful @Najada I can confirm the user handle is not showing in the activity feed in profile or portfolio (only works on project pages) - so is that an additional task for the backend?

psparacino:
> 
> Now that we are seeing this with user names, the ... styling I don't think is working for user names shorter than the space allowed. See `jabby...bby` and `najad...ada` as example.
> 
> It makes sense to support this format for 0x addresses (which are the fall back) but looks like we need a condition for checking if the handle name will will as is in the space allowed.
> 
> Screenshot 2023-04-21 at 1 01 06 PM

We can just show the whole handle if that's preferable? Some of them were long so I went the brevity route as a default.

psparacino:
> > The only place this is currently not possible is where the type is UserActivity, which looks to be on the user page. There is a note from two days ago that this is being added to the api response, so sounds like this will be a quick fix to add once that is done.
> 
> @psparacino @Atmosfearful @Najada I can confirm the user handle is not showing in the activity feed in profile or portfolio (only works on project pages) - so is that an additional task for the backend?

I tried adding buyer and seller handles to the UserActivity type in case the type just hadn't been updated yet, but no luck.

Yes, as far as I can tell atm the handles needs to be added to the api.

jabby09:
> > Now that we are seeing this with user names, the ... styling I don't think is working for user names shorter than the space allowed. See `jabby...bby` and `najad...ada` as example.
> > It makes sense to support this format for 0x addresses (which are the fall back) but looks like we need a condition for checking if the handle name will will as is in the space allowed.
> 
> We can just show the whole handle if that's preferable? Some of them were long so I went the brevity route as a default.

@psparacino yea it looks to me like we have space for 11 characters. In my mind, if handle <= 11 characters show the whole handle with no ... break. If handle >11 characters then do the ... break but I would not repeat the characters

Najada:
@jabby09 on /users/:userhandle I need to add the handles. Working on it now and will push an update on the api side.

psparacino:
> @jabby09 on /users/:userhandle I need to add the handles. Working on it now and will push an update on the api side.

ok, great. I'll add it in once it's added to the endpoint

jabby09:
> 

@Najada can you please let us know when this is complete? thanks

jabby09:
@psparacino @Atmosfearful handle name formatting looks now to be correct

Atmosfearful (approved):

PR #1070, opened 2023-04-22
Minor Dependency Updates

0xemc:
## Description

Update all packages to their safest minor versions 

https://docs.npmjs.com/cli/v8/commands/npm-update


ladytrekker (changes requested):
I would expect that the main `package.json` contains the updated versions for each package too?

Which command did you run? Not `npm update --save` correct?
https://docs.npmjs.com/cli/v8/commands/npm-update

0xemc:
> I would expect that the main `package.json` contains the updated versions for each package too?
> 
> Which command did you run? Not `npm update --save` correct? https://docs.npmjs.com/cli/v8/commands/npm-update

You're right.. I was missing the save flag but ended up using `ncu -u --workspaces --target minor` this time around.

Should be updated now.

ladytrekker (approved):

PR #1072, opened 2023-04-22
feat: import carbonmark api (#1032)

### LINKED ISSUE #1032: Import Carbonmark API to klimadao
**What**:  Add the existing repository at https://github.com/KlimaDAO/carbonmark-api to the klimadao monorepo

**Why**: Unblock developers and allow them to work across the stack

**How**:  Include a new folder titled carbonmark-api with the contents on https://github.com/KlimaDAO/carbonmark-api

**Expected Outcome**:

[ ] - Carbonmark API project is a project within the klimadao repo
[ ] - Build and development scripts are included in package.json
[ ] - ? Update deployment?

0xemc:
## Description

Import the project at https://github.com/KlimaDAO/carbonmark-api in to our monorepo

## Related Ticket

Resolves #1032

## Remaining Work
- [x] Update env.example with correct envs (and add correct values to .env or via vercel cli)
- [x] Test all endpoints manually
- [x] Update Readme to be more informative
- [ ] Update the repo link in the [vercel project settings ](https://vercel.com/klimadao/carbonmark-api/settings/git) (will complete after closing this PR)


package.json
```diff
@@ -27,6 +28,7 @@
     "dev-app": "npm run build-lib && npm run dev --workspace=app",
     "dev-cms": "cd ./cms && npm run start",
     "dev-carbonmark": "concurrently --kill-others-on-fail \"npm:dev-lib\" \"npm run dev --workspace=carbonmark\"",
+    "dev-carbonmark-api": "npm run dev --workspace=carbonmark-api",
```
ladytrekker, line 31: Do we need a build script too?
For local development and for Vercel?

ladytrekker (commented):
Thanks.

I guess you simply copy pasted?

The build scripts are missing in my opinion.
Or will this happen in a follow up?

package.json
0xemc: My understanding is that the build process is handled by the [vercel github integration](https://vercel.com/klimadao/carbonmark-api/settings/git). If you check the scripts on the original repo there was no `build` script

https://github.com/KlimaDAO/carbonmark-api/blob/42e9df24f31563cacd0d5d9bc65fd34fc3f0d266/package.json#L9-L13

You're right though we might need to update this to be the same as the carbonmark deploy process since it's now in a monorepo.. what do you think?

0xemc: Ok looking in to this the process seems to be:

1. update the git integration (to point to our mono repo)
2. update the [root directory](https://vercel.com/docs/concepts/monorepos#using-monorepos-with-vercel-dashboard) setting
3. the serverless.js file will handle the rest

Najada: hi @0xemc 
point 1,2,3 all correct.

but i wonder which settings will you update on vercel?
To link the new repo to the vercel deployment project, we should only run on the cli `vercel` and it ask us to link to the correct project.

Atmosfearful (approved):
Regarding the build discussion in the comments above:
We should be able to follow the same pattern as our other deployments, i.e. a vercel-specific npm install command. This allows us to deploy changes across the backend and frontend in the same PR.

Let's do this ASAP so that we can continue development in this repo, and not have to deal with the commit history diverging.
- start by creating a new deployment in vercel, separate from the existing carbon-api deployment that is already running.
- Reassign the domain to the new one, and it should 'just work' without changes to the frontend (perhaps, only reassign staging first?)
- If it fails we can always rollback to the existing deployment.
- @0xemc you should have sufficient permissions in vercel to do this, lmk if I need to elevate you to Admin, or if you need my help in any other way

We will have to increase the number of parallel deployments in vercel on our Pro plan, to keep build times from slowing down too much. I will do that soon

Issue #1019, opened 2023-03-19
[performance] General caching issues


jabby09:
I recently reported an issue in discord after creating a listing the project title was missing.

![Image](https://user-images.githubusercontent.com/103535195/226177713-47494115-175a-4ea6-81a7-c8347dba1bae.png)

Turns out it was a caching issue.

Overall seeing some issues when creating/updating listings etc of the information not displayed correctly or very quickly on the frontend. This can lead to confusion since the user may not know they need to hard refresh multiple times to see changes. 

As one example, when I create a new listing, sometimes it takes numerous refresh for the listing to show and if I change screens, like go to marketplace and then back to my profile my listing is not there until I refresh.

The other thing I often see (not a caching issue) is every time I click on my profile page the create listing button spinner shows for 8,10 or more seconds - what is causing that?

There are other issues already created around caching, i.e. KlimaDAO/klimadao#1008 and KlimaDAO/klimadao#1015 so clearly caching is a known issue. I'm not sure we can icebox this before launch but also not sure what the solution or level of effort would be to improve the UX.

ShimonD-KlimaDAO:
Single project page caching updates passed QA.
#1025

ShimonD-KlimaDAO:
Portfolio data page looks good, except for two minor notes. See my comment in #1047 
https://github.com/KlimaDAO/klimadao/pull/1047#issuecomment-1515618024

ShimonD-KlimaDAO:
#1099 passes QA. This concludes all caching PRs.

PR #1099, opened 2023-04-27
Use SWR for updating user profile data

ladytrekker:
## Description

This is the final PR for improving the page cache.

### How to test:

**Edit Profile**
- Change your profile data
- The changes should be visible immediately 
-  Note: When you refresh the page now, it might be that the data is not updated! This is because the backend might be slow. But in normal cases a user would not refresh now.

**Create, Change a Listing**
- Create or edit a Listing
- Loadings spinner should appear with "Updating your data" (see screenshot)
- Meanwhile, the backend is polled to wait for an updated user's activity
- Until then all listings are disabled to prevent the user from further editing
- When the backend needs to long, an error message is shown "please refresh the page"
- On Success, the listing is updated ✔️ 
- On Success, the activity is updated ✔️ 
- On Success, the corresponding project is updated  ✔️ (different page URL) 
- On Success, the portfolio activity is updated ✔️ (different page URL) 

**Additional Changes:**

- When a user has not yet been registered on Carbonmark => the Profile button is "Create Profile"

## Related Ticket

Resolves https://github.com/KlimaDAO/klimadao/issues/1019

### PENDING
Bildschirmfoto 2023-04-27 um 16 01 11

### New USER with "create profile" button label
Bildschirmfoto 2023-04-26 um 12 40 58

## Checklist

### DIFF
--- a/carbonmark/components/pages/Users/ProfileHeader/index.tsx
+++ b/carbonmark/components/pages/Users/ProfileHeader/index.tsx
@@ -1,34 +1,34 @@
 import { Trans } from "@lingui/macro";
 import { ProfileLogo } from "components/pages/Users/ProfileLogo";
 import { Text } from "components/Text";
+import { User } from "lib/types/carbonmark";
 import { FC } from "react";
 import * as styles from "./styles";
 
 type Props = {
+  carbonmarkUser: User | null;
   userName: string;
-  description?: string;
-  isCarbonmarkUser: boolean;
-  profileImgUrl?: string;
-  handle?: string;
 };
 
 export const ProfileHeader: FC<Props> = (props) => {
+  const isCarbonmarkUser = !!props.carbonmarkUser;
+
   return (
     <div className={styles.profileHeader}>
       <ProfileLogo
-        isCarbonmarkUser={props.isCarbonmarkUser}
-        profileImgUrl={props.profileImgUrl}
+        isCarbonmarkUser={isCarbonmarkUser}
+        profileImgUrl={props.carbonmarkUser?.profileImgUrl}
       />
       <div className={styles.profileText}>
         <div className={styles.titles}>
-          <Text t="h4">{props.userName}</Text>
-          {props.handle && (
+          <Text t="h4">{props.carbonmarkUser?.username || props.userName}</Text>
+          {props.carbonmarkUser?.handle && (
             <Text t="h4" className="handle">
-              @{props.handle}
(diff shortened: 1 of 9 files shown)


carbonmark/components/pages/Users/ProfileHeader/index.tsx
```diff
@@ -1,51 +1,51 @@
...
-        isCarbonmarkUser={props.isCarbonmarkUser}
-        profileImgUrl={props.profileImgUrl}
+        isCarbonmarkUser={isCarbonmarkUser}
+        profileImgUrl={props.carbonmarkUser?.profileImgUrl}
       />
       <div className={styles.profileText}>
         <div className={styles.titles}>
-          <Text t="h4">{props.userName}</Text>
-          {props.handle && (
+          <Text t="h4">{props.carbonmarkUser?.username || props.userName}</Text>
```
Atmosfearful, line 24: I'm confused because it looks like here props.userName can actually be the handle?
https://github.com/KlimaDAO/klimadao/pull/1099/files#diff-4ffaaf4f61284f7106b9ea09e882f4c38d70b88eab545268a412b9f50f923c95R25

So its possible that we could render `Handle @Handle` ?

Atmosfearful (approved):
perfection 🌟 
 just 1 question

ShimonD-KlimaDAO:
This passes all QA flawlessly.

carbonmark/components/pages/Users/SellerConnected/index.tsx
```diff
@@ -97,81 +114,58 @@ export const SellerConnected: FC<Props> = (props) => {
...
+      const userFromApi = await getUser({
+        user: props.userAddress,
+        type: "wallet",
+      });
+
+      // Merge with data from Updated Profile as backend might be slow!
+      const newUser = { ...userFromApi, ...profileData };
+
+      // Update the cache only, do not revalidate
+      await mutate(newUser, false);
```
0xemc, line 131: This is cool, I didn't know you could do this 👍

0xemc (approved):

carbonmark/components/pages/Users/ProfileHeader/index.tsx
ladytrekker: This should never happen. 
If `props.userName` falls back to the handle, then the carbonmark user data exists.
In this case the display name (username) is shown first.

PR #1103, opened 2023-04-28
Update all packages to latest major version (except ethers)

0xemc:
## Description

Updates all package version to their latest major version except for Ethers.js due to migration work required.

Some minor work needed for Lingui macro


app/components/views/Offset/index.tsx
```diff
@@ -599,9 +599,8 @@ export const Offset = (props: Props) => {
     .map((tkn) => ({
       ...tokenInfo[tkn],
       description: (function () {
-        if (isLoading) return <Trans id="shared.loading" />;
+        if (isLoading) return <Trans id="shared.loading">Loading...</Trans>;
         if (!props.isConnected || !Number(balances?.[tkn])) return "0";
-        return Number(balances?.[tkn]).toFixed(2);
```
ladytrekker, line 604: Why is this line removed here?

app/components/views/Stake/index.tsx
```diff
@@ -349,21 +349,16 @@ export const Stake = (props: Props) => {
               </button>
             </div>
             <div className={styles.stakeInput}>
-              <Trans
-                id={inputPlaceholderMessage[view].id}
-                render={({ translation }) => (
```
ladytrekker, line 354: Did you check if the language switch still works here?
I remember that this render function was needed back then as the component did not re-render when the user switches the language and the placeholder remains in the former language

app/components/views/Wrap/index.tsx
```diff
@@ -326,18 +326,13 @@ export const Wrap: FC<Props> = (props) => {
               </button>
             </div>
             <div className={styles.stakeInput}>
-              <Trans
-                id={inputPlaceholderMessage[view].id}
-                render={({ translation }) => (
```
ladytrekker, line 331: Same here.

app/pages/_app.tsx
```diff
@@ -14,7 +14,7 @@ function MyApp({ Component, pageProps }: AppProps) {
   return (
     <>
       <StyledEngineProvider injectFirst>
-        <I18nProvider i18n={i18n} forceRenderOnLocaleChange={false}>
```
ladytrekker, line 17: Aha, maybe this check here is enough? see comments above

ladytrekker (commented):
Some comments below.

I will test this PR myself again too next week.

app/components/views/Offset/index.tsx
0xemc: Great pickup, I think it was from auto formatting 🤦‍♂️

Atmosfearful (changes requested):
Pls document any intentional code changes, sounds like they were unintentional?

I just merged the API pr so you might need to take a second pass at this 😅

PR #1111, opened 2023-05-02
[carbon-projects] Add Project Content document type

Atmosfearful:
## Description

Adds a top-level document type called Project Content which is where any "unofficial" content should go, whether human or ai-generated.

Moves Image gallery and banner image fields to this new type.
Adds short description, long description, and meta fields


0xAurelius (approved):
Looks fine but not sure why we're reverting changes to the translation files?

PR #1112, opened 2023-05-02
feat: show popup before redirecting

### LINKED ISSUE #1055: [redirect] Pop up message when redirecting user to klimadao site
Currently in Carbonmark, we are leveraging the retirement aggregator when a user chooses to buy/retire from a pool or retire an asset from their portfolio page. The UX is jarring in that there is not a lot of information provided to the user about why this is happening or where they are being redirected.

There is a feature request enhancement requested to create pop up a msg before the user is redirected to the klimadao app.

The pop up message should be enabled for the following:
- BUY button on pool listing on project details page
- RETIRE NOW button on pool listing on project details page
- RETIRE button on portfolio page listing

The message in the pop up should read:

> You will be taken to KlimaDAO.finance to complete this transaction. After you click the link below to continue, you will be asked to connect your wallet and the project you selected here will already be selected for you on KlimaDAO.finance. From there, follow the directions on that page to complete your transaction.
>     
> Carbon credits you purchase or any retirements you make on KlimaDAO.finance will also be accessible in your Carbonmark portfolio.

0xMakka:
## Description

Show popup before redirecting to klimadao dApp. 

## Related Ticket

Resolves #1055 

## Checklist


jabby09:
@0xMakka @Atmosfearful based on my review, the only thing I see is wrong font style for the cancel button. Looking at the design should be black text, not red and also not sure if the font weight is correct.

Screenshot 2023-05-03 at 9 37 14 AM

carbonmark/components/ExitModal/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+    <div className={styles.content}>
+      <div>
+        <LaunchOutlinedIcon />
+        <Trans>
+          You will be taken to KlimaDAO.finance to complete this transaction.
+        </Trans>
+      </div>
+      <div>
+        <AccountBalanceWalletIcon />
+        <Trans>
```
Atmosfearful, line 22: Custom message IDs are useful for these really long strings to avoid duplicate data and inflating our string file sizes over time, as you can see in the output:

(short strings I prefer with no ids just to avoid boilerplate and duplicates)

image

carbonmark/components/ExitModal/styles.ts
```diff
@@ -0,0 +1,45 @@
...
+  width: 100%;
+  margin-top: 1.2rem;
+  border-color: var(--warn);
+  color: var(--warn) !important;
+`;
+
+export const continueButton = css`
+  width: 100%;
+  margin-top: 2.4rem;
+  color: white !important;
```
Atmosfearful, line 42: coare you able to explain why !important is neccessary, is there something we can do in how we order classnames for the underlying button components?

Atmosfearful (changes requested):

carbonmark/components/ExitModal/styles.ts
0xMakka: Let me look into this again and see can if I can resolve without using !important. As far as I remember the underlying colour was black and I couldn't seem to get around it any other way without using !important.

carbonmark/components/ExitModal/index.tsx
0xMakka: Noted & sounds good! Will update this to include custom message IDs for these longer strings.

PR #1120, opened 2023-05-04
Fix: handle should fallback to "You" on connected user

ladytrekker:
## Description

Backend sends down the handle for seller and buyer on the activity data.
These handles are shortened if the length is more than 11 characters.

However, if the current connected user is included in the activity data => "You" should be shown and not the user´s handle.
Same as for the wallet addresses.

## Related Ticket
None
## Changes

| Before  | After  |
|---------|--------|
| Bildschirmfoto 2023-05-04 um 15 11 59 | Bildschirmfoto 2023-05-04 um 15 12 28 |

## Checklist


Atmosfearful (approved):
lgtm thanks for taking the initiative to fix this!

PR #1125, opened 2023-05-08
fix: add required label

### LINKED ISSUE #995: [design] Better UX (tooltip?) for why CREATE NEW LISTING greyed out + Edit Profile required fields
When you login for the first time a basic profile is created for you in Carbonmark.

Screenshot 2023-04-03 at 9 34 34 AM

Note that the CREATE NEW LISTING is greyed out. This is because it is a requirement to edit and complete your seller profile. There is subtle messaging under the temporary display name but it not abundantly clear to a user why they cannot create a listing at this point and what is required to activate that functionality.

**Requirements**
- [ ] Please recommend a better UX for this experience, suggestions made are tool tip on the CREATE NEW LISTING button (but only when greyed out) and / or better + more obvious messaging on what is required to activate the button 
- [ ] As part of this, can we also update the Edit Profile modal to indicate required fields. We currently only show this once you start editing and try and save but it would be helpful to know up front that @username and Display name are required
Screenshot 2023-04-03 at 9 41 19 AM
Screenshot 2023-04-03 at 9 42 31 AM

0xMakka:
## Description

Add required field indicators to edit profile labels.

## Related Ticket

Resolves #995 

## Checklist

### DIFF
--- a/app/locale/en-pseudo/messages.po
+++ b/app/locale/en-pseudo/messages.po
@@ -6,6 +6,12 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "X-Generator: @lingui/cli\n"
 "Language: en-pseudo\n"
+"Project-Id-Version: \n"
+"Report-Msgid-Bugs-To: \n"
+"PO-Revision-Date: \n"
+"Last-Translator: \n"
+"Language-Team: \n"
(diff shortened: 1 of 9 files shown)


Atmosfearful (approved):
Perfect thanks!

Issue #1106, opened 2023-04-29
[retire] Retire from portfolio - dev implementation


jabby09:
Related design task #1087 

See [Retirement Flow (from portfolio)](https://www.figma.com/file/I5hXFeW1DKjV7agHq7T4UZ/Portfolio?type=design&node-id=247-10822&t=ScV4oxm9NsoeqA4z-4) design.

### User journey and requirements:

- [x] Note desktop and mobile versions
- [x] From Portfolio page, click RETIRE button on project card in your portfolio of assets
- [x] Takes user to Retire form page, design [here](https://www.figma.com/file/I5hXFeW1DKjV7agHq7T4UZ/Portfolio?type=design&node-id=247-10835&t=ScV4oxm9NsoeqA4z-4)
    - Ignore for now the translation option in the header
    - Need to check on the need for the "< BACK TO PORTFOLIO" breadcrumb
    - Need to check on the Asset Details icon
- [x] User completes form and clicks RETIRE CARBON
- [x] Confirm Retirement modal, APPROVE screen, design [here](https://www.figma.com/file/I5hXFeW1DKjV7agHq7T4UZ/Portfolio?type=design&node-id=247-10980&t=ScV4oxm9NsoeqA4z-4)
- [x] Change intro text from "Please approve the transaction" to:
> You are about to retire a carbon asset.
> 
> The first step is to grant the approval to transfer your carbon asset from your wallet to Carbonmark, the next step is to approve the actual transfer and complete your retirement.
> 
> Verify all information is correct and click 'approve' to continue.

- [x] User clicks APPROVE
- [x] Confirm Retirement modal, SUBMIT screen
- [x] Change intro text from "Please submit the transaction" to:

> The previous step granted the approval to transfer your carbon asset from your wallet to Carbonmark, your retirement has not been completed yet.
> 
> To finalize your retirement, verify all information is correct and then click 'submit' below.

- [x] Processing Retirement spinner while waiting on transaction to complete, design [here](https://www.figma.com/file/I5hXFeW1DKjV7agHq7T4UZ/Portfolio?type=design&node-id=247-10991&t=ScV4oxm9NsoeqA4z-4)
- [x] Payment Successful screen, design [here](https://www.figma.com/file/I5hXFeW1DKjV7agHq7T4UZ/Portfolio?type=design&node-id=247-11601&t=ScV4oxm9NsoeqA4z-4)
- [x] RETIRE MORE CARBON button takes user back to /portfolio page
[...]

PR #1137, opened 2023-05-12
Tufnel/retire from cm plumbing 1097

### LINKED ISSUE #1097: [retirements] Retire from portfolio - initial plumbing
While waiting for finalized mocks we can start with the plumbing. This is high priority for this epic / milestone: https://github.com/KlimaDAO/klimadao/milestone/9

# Summary
Goal is to support retirement of your own assets, in carbonmark, without bouncing the user to klimadao.

This is the journey:
1. A user has already purchased a carbon asset, which can be seen listed in their /portfolio page.
2. They click RETIRE and are taken to an offset form.
3. They complete the form, with constraints and validation (follow-up PR)
4. They complete an approval txn, and a retirement txn, in the Transaction Modal.
5. On success, they are taken to a 'retirement successful' screen.
6. From that screen (after a few seconds delay for subgraph to catch up) they can click to navigate to their 'retirement page'.

# Engineering Requirements
For this PR we only want to write up the key data fetching, transaction logic, and navigation logic.
We don't have to finalize form validation or designs yet.
* [ ] Do *not* change the URL on the RETIRE button.
  - This allows us to merge this PR without exposing it to users until it's ready
image
* [ ] Set up a page to hold the form under the url `/porfolio/retire/[project_token_0x]`
  * [ ] When this page loads, if the user is not connected, bounce them back to /portfolio
  * [ ] When the page loads, fetch the relevant project info from subgraph. See dapp for an example of how we do that.|
    - project token balance, project name, project id, vintage, project category.
* [ ] Write retirement functions that use aggregator v2, feel free to copy paste as much as you can from the existing logic.
  * [ ] Existing logic can be simplified now, because it's only concerned with retiring an asset you hold (no payment methods, prices etc.)

psparacino:
## Description

Retirement plumbing in place to retire from portfolio. 

Note for testing:
Not all project tokens seem to be available on the [mainnet subgraph](https://api.thegraph.com/subgraphs/name/najada/marketplace-matic) at the moment. 

Ex. this is one that doesn't show up in the subgraph query--> 0xf0b3aed0232b3c51693323e45878c9173b6c43fe
Screen Shot 2023-05-11 at 5 18 57 PM

I'm not sure yet why some are included to the exclusion of others. They all appear to be on the api, which uses the projectKey to search for an individual project (this uses the token address required from the url so I don't see how the current endpoint would work here), so the tokens possibly originate from multiple subgraphs? 

The api only looks to use one graph endpoint though and all projects look to be available. Definitely would like to shed some light on this and investigate further so the fetching solution is solid.

For testing this is a project that shows up on the subgraph:
     
```
        "id": "0x0",
        "projectID": "191",
        "projectAddress": "0xb139c4cc9d20a3618e9a2268d73eff18c496b991",
        "methodology": "ACM0002",
        "name": "4×50 MW Dayingjiang- 3 Hydropower Project Phases 1&2",
        "projectType": "",
        "region": "China",
        "registry": "VCS",
        "updatedAt": "1680637463",
        "vintage": "2008",
        "key": "VCS-191"
```

Going to begin on dialing in the styling and form validation in ticket #1106 
      
## Related Ticket

Resolves #1097  

## Checklist


carbonmark/components/pages/Retire/RetireForm/index.tsx
```diff
@@ -0,0 +1,439 @@
...
+                  onChange={(event) =>
+                    handleRetirementChange("retirementMessage", event)
+                  }
+                  placeholder={t({
+                    id: "offset.retirement_purpose",
+                    message: "Describe the purpose of this retirement",
+                  })}
+                />
+              </div>
+              {isLargeOrBelow ? (
```
Atmosfearful, line 336: why can't we use normal css breakpoints to hide this?

carbonmark/components/pages/Retire/RetirementStatusModal/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+        </Trans>
+      </Text>
+
+      <ButtonPrimary
+        href={props.retirementUrl}
+        className={styles.viewButton}
+        target="_blank"
+        label={
+          <Trans id="offset.successModal.cta_1">
+            VIEW AND SHARE CERTIFICATE
```
Atmosfearful, line 39: As best we can lets keep strings In a consistent case, and then transform using CSS `text-transform`

```suggestion
            View and share certificate
```

carbonmark/pages/portfolio/retire/[project_token_0x]/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+    paths: [],
+    fallback: "blocking",
+  };
+};
+
+interface Params extends ParsedUrlQuery {
+  project_token_0x: string;
+}
+
+export const getStaticProps: GetStaticProps<RetirePageProps, Params> = async (
```
Comment (Atmosfearful) [inline carbonmark/pages/portfolio/retire/[project_token_0x]/index.tsx:18]: For this page I would flip the URL structure to be:
`carbonmark/pages/portfolio/[project_token_0x]/retire/index.tsx`

Because we might some day add 'transfer' or some other actions / views for each asset beyond retire.

Atmosfearful (changes requested):
Looking fantastic, some minor suggestions. I'm open to merging this now if you want to address these in a follow up. Just rebase and let me know

carbonmark/components/Modal/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+import * as styles from "./styles";
+
+interface Props {
+  title: ReactNode;
+  onToggleModal?: () => void;
+  className?: string;
+  children: ReactNode;
+}
+
+export const Modal: FC<Props> = (props) => {
```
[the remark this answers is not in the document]
psparacino: using Modal in shared/components

PR #1141, opened 2023-05-14
0xMakka/1094 add carousel to project view

### LINKED ISSUE #1094: [project details] Create a ProjectMediaCarousel
[Design is here](https://www.figma.com/file/5GewbdVaM1Q06CwHmF7d3F/%2Fprojects?node-id=11-60104&t=26s4JQ0g3hCMiYlx-0).

**Notes:**

- If the project does not have images, only show the map as it appears now
- Fade on either end when there's more images that direction (see designs for details)
- Info icon on the image supports image attribution information
- Auto-scrolling is a long term objective and not required for initial implementation

Screenshot 2023-04-26 at 5 32 25 PM

0xMakka:
## Description

Adds a project carousel to the product page when images are available from the api, otherwise shows the map.

## Related Ticket

Resolves #1094

## Checklist


jabby09:
hey @0xMakka when will this one be out of DRAFT

btw, I checked the sanity document again and it had publishing errors. I fixed those and published the doc and the api (https://api.carbonmark.com/api/projects/VCS-1764-2020) looks like it is working again, at least I can see the image urls

jabby09 (changes requested):
hey @0xMakka feedback below:

Screenshot 2023-05-25 at 9 16 00 AM

- [x] As per the ticket requirements, the map always needs to be present as the first image. Default thumbnail image for the map is in the [figma](https://www.figma.com/file/5GewbdVaM1Q06CwHmF7d3F/%2Fprojects?type=design&node-id=11-60104&t=CaenzpThxh0SEZ7C-4). If you need a better image let me know. 

Once the map is present then I'll be able to see the fade effect and left/right blue advance arrows.

On mobile...

- [x] Probably don't need the blue advance arrows on mobile since on a phone can use your finger to move the roll
- [ ] Info icon on mobile is not working (I can't tap it to show the image description)

IMG_6FEE5039BC3A-1

0xMakka:
> * Info icon on mobile is not working (I can't tap it to show the image description)

@jabby09 I've tried on both mobile & in the devtools and this isn't an issue for me. Each time I click the icon I get the tooltip.

Screenshot 2023-05-25 at 13 44 55

jabby09:
@0xMakka I tried the info icon on my iphone again and still no luck.

What type of phone are you using?

0xMakka:
> @0xMakka I tried the info icon on my iphone again and still no luck.
> 
> What type of phone are you using?

@jabby09 Hmm very strange, it's using the same tooltip component that's used throughout carbonmark (regardless, I will take another look). Is it that nothing happens at all when you click it?

I'm using an Android Xiaomi Mi 9t.

0xMakka:
> hey @0xMakka I noticed the info icon design was not working well with different background images so I asked professor to change it.
> 
> You can see the update in figma [here](https://www.figma.com/file/5GewbdVaM1Q06CwHmF7d3F/%2Fprojects?type=design&node-id=404-21432&t=9lKSQ9Xh9BEEWFk2-4)
> 
> image

Thanks @jabby09 great idea and suggestion on this - I've just made the change and agree it looks much better. Can you retest on iphone again and see if it's working? I've testing on an ipad, android phone and laptop and it seems to be working as expected.

jabby09 (approved):
good to go

0xMakka:
@Atmosfearful when you get a chance where you could take a look at this that would be awesome :)

carbonmark/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,149 @@
...
+  const [maxSlidesToShow] = useState(5);
+  const [selectedIndex, setSelectedIndex] = useState(0);
+  const [images, setImages] = useState<Array<CarouselImage>>(props.images);
+  const [mainRef, mainApi] = useEmblaCarousel(props.options);
+  const [thumbsRef, thumbsApi] = useEmblaCarousel({
+    containScroll: "keepSnaps",
+    dragFree: true,
+  });
+
+  const onThumbClick = useCallback(
```
Atmosfearful, line 30: nitpick, lets go with the onClick -> handleClick convention I outlined here:
https://github.com/KlimaDAO/klimadao/wiki#use-an-oneventhandleevent-naming-convention

carbonmark/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,149 @@
...
+    (index: number) => {
+      if (!mainApi || !thumbsApi) return;
+      mainApi.scrollTo(index);
+    },
+    [mainApi, thumbsApi]
+  );
+
+  const onPreviousClick = useCallback(() => {
```
Atmosfearful, line 38: If these are just being passed down as click handlers they don't need `useCallback`

carbonmark/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,149 @@
...
+    setSelectedIndex(mainApi.selectedScrollSnap());
+    thumbsApi.scrollTo(mainApi.selectedScrollSnap());
+  }, [mainApi, thumbsApi, setSelectedIndex]);
+
+  useEffect(() => {
+    if (!mainApi) return;
+    onSelect();
+    mainApi.on("select", onSelect);
+    mainApi.on("reInit", onSelect);
+  }, [mainApi, onSelect]);
```
Atmosfearful, line 61: We might need to clean up the old listeners-- If `mainAPI` or `onSelect` objects change, this effect will re-run, and new listeners will be created. Unless maybe embla cleans the old ones up automatically?

Question: is it possible to only invoke this on mount? Or does mainApi only get instantiated after the effect fires?
```tsx
  useEffect(() => {
    onSelect();
    mainApi.on("select", onSelect);
    mainApi.on("reInit", onSelect);
    return () => {
      mainApi.off("select", onSelect);
      mainApi.off("reInit", onSelect)
    }
  }, []);
  ```
  
  If so this would allow you to remove the useCallback wrapper from onSelect as well

File: carbonmark/components/Carousel/Carousel.tsx [hunk 4]
```diff
@@ -0,0 +1,149 @@
...
+
+  useEffect(() => {
+    if (props.location) {
+      const images = [...props.images];
+      images.unshift({
+        url: "/map-thumbnail.png",
+        caption: images?.[1]?.caption,
+      });
+      setImages(images);
```
Atmosfearful, line 70: This does not need to happen in an effect, nor does it need state, just use an inline variable

```ts
const imagesWithThumb = [...props.images];
if (props.location) imagesWithThumb.unshift(mapThumbnail);
```

carbonmark/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,149 @@
...
+              <div key={`thumbnail-${index}`} className={styles.thumbsSlide}>
+                <button
+                  type="button"
+                  onClick={() => onThumbClick(index)}
+                  className={styles.thumbsSlideButton}
+                >
+                  {index === selectedIndex && (
+                    <div className={styles.selected} />
+                  )}
+                  <img
```
Atmosfearful, line 127: I believe Next `Image` would work here, which could save quite a bit of bandwidth if we are able to use the `sizes` prop to constrain the max image size intelligently.
We have a util to make that more straightforward, lmk if you need help:
https://github.com/KlimaDAO/klimadao/blob/96db714b4ada84a4e3e3a31fe8301c2db3c68be7/lib/utils/getImageSizes/index.ts#L14

This `sanityImageLoader` should work
https://github.com/KlimaDAO/klimadao/blob/cac833f412c8a9f183f4dbc49e5faf551ce7e920/site/lib/sanityImageLoader.ts#L3

https://nextjs.org/docs/pages/api-reference/components/image#loader

Atmosfearful (approved):
Looks and works fantastic! I have some change requests, but these should not impact the UX so lets do them in follow-up PR(s).

0xMakka:
> Looks and works fantastic! I have some change requests, but these should not impact the UX so lets do them in follow-up PR(s).

@Atmosfearful My bad on a few of these  - I was following along from the embla docs which follow some different conventions. 

https://codesandbox.io/s/3moc9z?file=/src/js/EmblaCarousel.tsx

carbonmark/components/Carousel/Carousel.tsx
0xMakka: Yeah agreed here - I originally had two separate pieces of data for thumbnails and the main slider as I was trying to only prepend the image for the thumbnails and was keeping the thumbnail data in state, but ended up using the same data for both and forgot to remove the effect and state variables.

PR #1152, opened 2023-05-22
Upgrade dependencies to major versions

Atmosfearful:
TODO:
* [ ] test that language switcher in app still works, because lingui removed the forceRerender prop
* [ ] test that placeholder inputs for wrap and stake inputs work

EDIT:
A change in the lingui context provider has broken translations, I set this aside to work on web2 api demo applications. Hoping to return to this next week


0xemc:
@Atmosfearful these conflicts look painful.. Should we pull the lingui changes out in to a seperate PR?

PR #1153, opened 2023-05-23
Add link field to methodology type in carbon-projects CMS

0xAurelius:
## Description

Adding this link field so we can have a click-through on the tooltip for a project's methodology

This allows the user to learn more about the underlying methodology used by a project without having to dig around on the registry website

Subsequent work is required to surface this link field in the API, and integrate it into the user interface as a link on the tooltip for each project page and the project cards on the search page


0xAurelius:
Created from fork by accident

PR #1159, opened 2023-05-25
0xMakka/1129 - Improved filtering & sorting

### LINKED ISSUE #1129: [projects] Improved filter/sort - dev implementation
Here are the [updated designs](https://www.figma.com/file/5GewbdVaM1Q06CwHmF7d3F/%2Fprojects?type=design&node-id=518-35606&t=lrdWGniyV4SsFlG4-4) for enhancing the filter modal plus location and how filter/sort features work (notice we are moving some of the functionality to the marketplace page).

### Notes/Requirements/Acceptance Criteria

- [x] As always, ensure to note differences between desktop and mobile designs

**_/projects Marketplace screen - No Filters applied_**

- [x] Tile/list view feature can be ignored for now, implementing the List View is a separate future epic
- [x] Sort capability is now pulled into the marketplace screen, note the **xxx Results** feature giving the user feedback on number of projects selected. With no filters applied, this should default to the total number of project cards.

Screenshot 2023-05-09 at 9 31 04 AM

**_Filter Results modal_**

- [x] Sort removed
- [x] # selected info reported to user to indicate they have selections made in the various sections
- [ ] Project total results reported back to the user; this total will change as the user makes filter selections
- [ ] Compiling Results... feedback to user since it may take time to calculate the number of projects

Screenshot 2023-05-09 at 9 42 26 AM
Screenshot 2023-05-09 at 9 46 18 AM

**_/projects Marketplace screen - Filters applied_**

- [x] Filters button label changes to indicated the number of filters applied
- [x] Clear Filters button added to allow clearing all the filters without needing to open the modal or clear all the filter pills
- [x] xxx Results feature updated to show the number of projects selected based on the filters applied
- [x] Filters selected in the modal appears as individual pills
- [x] Individual filter pills can be removed, changing the number of projects selected result
- [x] In worse case situations where the number of filters applied do not fit on the screen, a fade is applied on the last pills and a "+X MORE" message is shown; clicking "+X MORE" opens the filter modal

Screenshot 2023-05-09 at 9 47 34 AM

0xMakka:
## Description

This task is now ready for testing and the below issues have been addressed.

- [x] Reset filters from projects view
- [x] Ability to remove individual filters from projects view
- [x] Fix an issue with sorting where the selected sort item resets
- [x] Address issue when single filter is selected & formatted as string
- [x] Address issue with dropdown sort not updating url
- [x] If maximum filters are added on desktop - show fade effect and "+ x more" text.

Note: I've left the text as "+ more" for now so we can get eyes on this PR, but I'm still investigating how feasible it is to show how many filters aren't visible on the screen.

## Related Ticket

Resolves #1129

## Checklist


0xMakka:
@jabby09 / @Atmosfearful this task can start being reviewed. There is one remaining item I need to finish which is showing the "+ 4 more" label when multiple filters are added to overflow the page and one or two styling issues I'm looking at addressing now on mobile.

jabby09:
> > @jabby09 / @Atmosfearful this task can start being reviewed. There is one remaining item I need to finish which is showing the "+ 4 more" label when multiple filters are added to overflow the page and one or two styling issues I'm looking at addressing now on mobile.
> 
> @0xtapi (cc @ShimonD-KlimaDAO) this one is ready to start QA. This is the ticket that has the link to the design plus the dev requirements I outlined: #1129
> 
> I will do a QA review myself in Brave/Chrome desktop. If you could focus on other browsers and mobile that would be helpful.

0xtapi:
Did some QA, results below!

### **Desktop**:

Tested on Brave, Chrome, Edge, Vivaldi (all Chromium), Firefox. Didn't notice any differences.

1. Sort box looks too far left on a 4K screen. FIgma looks like Sort should align vertically with the Search box. It is aligned if I use less screen space, like half of the screen. Seems to break once I stretch the window to about 4/5th of available space.
sortbox

2. Same goes for the filter items, which show identical behavior:
filteritems

3. When blur effect is applied during filter selection, the background seems to be much darker than on Figma (could be an intentional change to make the window pop more, I guess):
blur

4. The small arrow on the right side should point to the right, not down, according to Figma:
arrow

5. "+x MORE" button is missing (already addressed by 0xMakka above). There should be a button here allowing the user to access the Filters menu, but it's not there (second picture is Figma):
more
more2




### **Mobile**:

Tested on Chrome, Vivaldi, Firefox.

1. The blue "BETA" should be aligned to the left, next to the Carbonmark icon:
Screenshot_20230712-150710-057

2. Chrome and Vivaldi browsers (same engine) stretch the project cards to fill the screen, Firefox does not stretch them. Not sure which behavior is correct, see:
Screenshot_20230712-150901-847
Screenshot_20230712-151000-247

3. Globe icon on the language selection button is much smaller than on Figma designs, but I believe that's a known issue?

0xMakka:
> Did some QA, results below!
> 
> ### **Desktop**:
> Tested on Brave, Chrome, Edge, Vivaldi (all Chromium), Firefox. Didn't notice any differences.
> 
> 1. Sort box looks too far left on a 4K screen. FIgma looks like Sort should align vertically with the Search box. It is aligned if I use less screen space, like half of the screen. Seems to break once I stretch the window to about 4/5th of available space.
> 
> sortbox
> 2. Same goes for the filter items, which show identical behavior:
> 
> filteritems
> 3. When blur effect is applied during filter selection, the background seems to be much darker than on Figma (could be an intentional change to make the window pop more, I guess):
> 
> blur
> 4. The small arrow on the right side should point to the right, not down, according to Figma:
> 
> arrow
> 5. "+x MORE" button is missing (already addressed by 0xMakka above). There should be a button here allowing the user to access the Filters menu, but it's not there (second picture is Figma):
> 
> more more2
> ### **Mobile**:
> Tested on Chrome, Vivaldi, Firefox.
> 
> 1. The blue "BETA" should be aligned to the left, next to the Carbonmark icon:
>    Screenshot_20230712-150710-057
> 2. Chrome and Vivaldi browsers (same engine) stretch the project cards to fill the screen, Firefox does not stretch them. Not sure which behavior is correct, see:
> 
> Screenshot_20230712-150901-847 Screenshot_20230712-151000-247
> 3. Globe icon on the language selection button is much smaller than on Figma designs, but I believe that's a known issue?

Awesome @0xtapi, thanks for this - will start fixing the visual issues mentioned above and let you know when they are ready to retest.

0xtapi:
One more for mobile: Navigation section looks different on Firefox vs. Chrome-based browsers. Firefox seems to draw more of its design from the desktop version of Carbonmark. Filters/Clear Filters should not exist on mobile according to Figma. Also, free space between language selector icon and and hamburger menu button is slightly different? 🤔 First picture is Firefox, second is Chrome:

Screenshot_20230712-171615-468
Screenshot_20230712-171721-040

0xMakka:
Hi @0xtapi, I've added some fixes for the following issues:

**Desktop**
1. Sort box too far left 
2. Filter items too far left 
3. Background modal blur effect 
4. The small arrow on the right side should point to the right, not down

**Mobile**
3. Globe icon on the language selection button is much smaller than on Figma designs.

0xtapi:
Hey @0xMakka, I tested your fixes on all previously mentioned browsers. No issues, everything works, nice job!

jabby09 (changes requested):
hey @0xMakka 

Noticed in the Filter Results modal, as I make filter selections the number of results is not updating. I have to click APPLY to run the query which then dismisses the Filter Results modal.

I believe the design intention was that with the Filter Results modal open, as a user I can make filter selections and then the number of results updating. Once I'm satisfied, I can then click APPLY to dismiss the modal and show the results in the marketplace.

This is shown here in this design screen, where a user has applied another filter and we show a Compiling results... message and the APPLY button is greyed out
Screenshot 2023-07-15 at 10 50 56 AM

Is this possible to do?

0xMakka:
> hey @0xMakka
> 
> Noticed in the Filter Results modal, as I make filter selections the number of results is not updating. I have to click APPLY to run the query which then dismisses the Filter Results modal.
> 
> I believe the design intention was that with the Filter Results modal open, as a user I can make filter selections and then the number of results updates. Once I'm satisfied, I can then click APPLY to dismiss the modal and show the results in the marketplace.
> 
> This is shown here in this design screen, where a user has applied another filter and we show a Compiling results... message and the APPLY button is greyed out. Once the updated results is shown, the APPLY button would be active again. Screenshot 2023-07-15 at 10 50 56 AM
> 
> Is this possible to do?

Hey @jabby09, I thought I had this working but might have changed something at some point along the way that broke that functionality. I've spent some time there this evening and added a fix for this. Would appreciate if you could retest this when you have a chance. 🙏

jabby09:
> Would appreciate if you could retest this when you have a chance

thanks @0xMakka i did a quick look at it does now appears to be working as expected. 

will give the entire PR another review when I get a minute either Sunday or Monday morning but this is looking very close to being done.

nice work man!

jabby09 (approved):
looks good!

carbonmark/components/Dropdown/index.tsx
```diff
@@ -48,6 +48,11 @@ export function Dropdown<V, T extends FieldValues = FieldValues>(
     close();
   }, [selected]);
 
+  useEffect(() => {
+    if (initial === selected.id) return;
+    setSelected(defaultOption);
+  }, [initial]);
```
Atmosfearful, line 54: It should already be set to defaultOption where useState is invoked. Why do we need this? It's not clear to me why.

I'm worried it will cause problems as it looks like it could trigger the other useEffect in some cases.

carbonmark/components/pages/Projects/index.tsx
```diff
@@ -21,25 +25,61 @@ import { NextPage } from "next";
...
+  sort: "recently-updated",
+  country: [],
+  category: [],
+  vintage: [],
+};
 
-  const sortFn = get(PROJECT_SORT_FNS, sortKey) ?? identity;
+const Page: NextPage = () => {
+  const router = useRouter();
+  const [modalOpen, setModalOpen] = useState(false);
```
Atmosfearful, line 53:
```suggestion
  const [showFilterModal, setShowFilterModal] = useState(false);
```

carbonmark/components/pages/Projects/index.tsx
```diff
@@ -21,25 +25,61 @@ import { NextPage } from "next";
...
+  const defaultValues = { ...DEFAULTS, ...router.query };
```
Atmosfearful, line 54: Similar to how I consume query params in the old app.klimadao.finance offset UI with a hook called [useOffsetParams](https://github.com/KlimaDAO/klimadao/blob/373c7deb32485bc88d530f6c12bba48e0330a8e2/app/lib/hooks/useOffsetParams.ts#L71), I would try to isolate all the query params logic into a hook or component if possible. (Don't copy that code as the requirements there are very different.)

The idea is the helper should consume, validate, and apply types to the query param values so the component can stay simple. This also avoids us passing unsupported params or typos into the form. Something like:

```ts
const { sort, countries } = useProjectsFilterParams();
const form = useForm({ sort, countries });
```

This hook should contain as much of the useEffect logic as you can as well.

Atmosfearful (changes requested):
* [ ] Opening the filters modal causes an empty sort query to appear in the URL.
  image

* [ ] Strange behavior: Looks like the url query param is not populated when choosing a sort option,  until *after* the filter modal has been opened, then it appears to work correctly.

* [x] When I click `Marketplace` in the nav menu, it looks like the filters are cleared properly, but the sort is not cleared. The expected behavior to me is that all sort and filter should be cleared as if the user had freshly navigated to /projects

carbonmark/components/pages/Projects/index.tsx
0xMakka: Done. 👍

carbonmark/components/Dropdown/index.tsx
0xMakka: Apologies, I should have left a comment around this to give some context. 

> Is it because the query params are parsed after first render?

Yeah, so the issue I was facing was - choose a sort option from the dropdown, reload the page. The sort query param remains as intended but the dropdown never gets updated and always takes the default sort value "recently-updated" - due to the query params not been parsed yet. 

So the idea here was to set the selected value on the dropdown if the sort value changes and exit if the value hadn't changed. Went down the rabbit hole with this one, so would love to hear other suggestions on this one if you have any - I've tried by waiting for the router.isReady to be true before rendering the dropdown (which works in theory - however throws hydration errors, same with not rendering the page until the router.isReady param is true).

0xMakka:
> * [ ]  Opening the filters modal causes an empty sort query to appear in the URL.
>   image
> * [ ]  Strange behavior: Looks like the url query param is not populated when choosing a sort option,  until _after_ the filter modal has been opened, then it appears to work correctly.
> * [ ]  When I click `Marketplace` in the nav menu, it looks like the filters are cleared properly, but the sort is not cleared. The expected behavior to me is that all sort and filter should be cleared as if the user had freshly navigated to /projects

Thanks for this, not sure how I didn't spot the issue with the empty sort query - I think I spent so much time testing it with filters, reloading ensuring everything maintained correctly etc... 

Will take a look at this and resolve - the issue with clicking `Marketplace` and not clearing is likely related.

carbonmark/components/pages/Projects/index.tsx
0xMakka: Sure, will look at tidying this up and moving to a dedicated hook or component.

0xemc (commented):
This is likely not realted to this PR but I noticed that the "Apply" button is now unecessary?

As soon as filters are selected the results are updated. @jabby09 @Atmosfearful thoughts?

image

carbonmark/components/ProjectFilterModal/index.tsx
```diff
@@ -85,69 +92,100 @@ export const ProjectFilterModal: FC<ProjectFilterModalProps> = (props) => {
     value: vintage,
   }));
 
+  const fetchProjects = (values: ModalFieldValues) => {
+    const { search, sort } = router.query;
+    const query = search ? { ...values, search, sort } : { ...values, sort };
```
0xemc, line 97: Isn't this equivalent? (unless the values are possibly null?)
 
```suggestion
    const query = { ...values, search, sort };
```

carbonmark/components/ProjectFilterModal/index.tsx
```diff
@@ -85,69 +92,100 @@ export const ProjectFilterModal: FC<ProjectFilterModalProps> = (props) => {
...
-      undefined,
-      { shallow: true } // don't refetch props nor reload page
-    );
-    //Close the modal on submit
-    props.onToggleModal?.();
+    fetchProjects(values);
+    if (!isValidating) {
+      props.onToggleModal?.();
+    }
+  };
```
0xemc, line 106: Not sure this is necessary since the watch is fetching anyway on change.. Maybe we should remove the Apply button all together..

carbonmark/components/Dropdown/index.tsx
0xMakka: Hey @0xemc - thank you for this tip. I went down the rabbit hole on this one completely and this is a great simple solution.

carbonmark/components/pages/Project/ProjectFilters/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+        router.query[key] as List<string>,
+        (value: string) => value === filter
+      );
+    });
+    router.replace({ query: router.query }, undefined, { shallow: true });
+  };
+
+  return (
+    <>
+      {!!flatMap(omit(props.defaultValues, ["search", "sort"]))?.length && (
```
0xemc, line 42: Can you extract this for readability?

```ts

const filters = flatMap(omit(props.defaultValues, ["search", "sort"]));
const hasFilters = notEmpty(filters)
```

carbonmark/components/pages/Project/ProjectFilters/styles.ts
```diff
@@ -0,0 +1,71 @@
...
+  border-radius: 4rem;
+  color: white;
+  font-weight: 600;
+  font-size: 1.2rem;
+  line-height: 1.6rem;
+
+  & svg {
+    cursor: pointer;
+  }
+`;
```
0xemc, line 71: Is this a common component that we might use elsewhere in the site? If so we should probably extract it

0xemc:
Great PR @0xMakka, Looks smooth on the UI. Some small suggestions from me otherwise gtg

carbonmark/components/pages/Project/ProjectFilters/styles.ts
0xMakka: I haven't seen anywhere else in the figma files where a similar component was used, so didn't extract it for this reason. Will take another look though and double check and can move it out into it's own component.

0xemc (approved):

carbonmark/components/ProjectFilterModal/index.tsx
0xMakka: Hmm, I thought I had a case where the search param was getting added as an empty string to the url hence the check. On second look now it seems this isn't the case, so will remove and clean this up 👍

0xMakka:
> 



> Great PR @0xMakka, Looks smooth on the UI. Some small suggestions from me otherwise gtg

Thanks for the great feedback @0xemc. Getting a little late here, so will respond and address your other feedback tomorrow 👍

jabby09:
> * When I click `Marketplace` in the nav menu, it looks like the filters are cleared properly, but the sort is not cleared. The expected behavior to me is that all sort and filter should be cleared as if the user had freshly navigated to /projects

@0xMakka did this feedback from atmos get addressed? I just re-tested this and the sort is not being cleared if I intentionally re-click 'Marketplace' in the nav

0xMakka:
> > * When I click `Marketplace` in the nav menu, it looks like the filters are cleared properly, but the sort is not cleared. The expected behavior to me is that all sort and filter should be cleared as if the user had freshly navigated to /projects
> 
> @0xMakka did this feedback from atmos get addressed? I just re-tested this and the sort is not being cleared if I intentionally re-click 'Marketplace' in the nav

@jabby09 I was working on it yesterday evening. I put in a fix which addressed the issue but it introduced an issue elsewhere so I reverted the change - will be looking at it again later today.

jabby09:
> I was working on it yesterday evening. I put in a fix which addressed the issue but it introduced an issue elsewhere so I reverted the change - will be looking at it again later today.

okay cool, no worries. not a show stopper - I don't think this should stop us from releasing this. can always fix later in another PR.

jabby09:
> This is likely not related to this PR but I noticed that the "Apply" button is now unnecessary?
> 
> As soon as filters are selected the results are updated. @jabby09 @Atmosfearful thoughts?

@0xemc @Atmosfearful from @theprofessora 

> leave it as is. even if it doesn't do anything, it still provides comfort to the user that their task will be executed. If we're uncomfortable with the word "apply" specifically, we can change it to say "view results"

So perhaps a middle ground is change the button to say VIEW RESULTS

jabby09:
@0xMakka professor noticed one small thing...

When the result in singular, we should say 1 result instead of 1 results

0xemc:
A heads up on rebasing and force pushing @0xMakka. Github fails to understand when the shared history has changed via rebases so a lot of the links and behaviour of PRs breaks :(

Screenshot 2023-07-26 at 10 52 16 am

Likewise for the "View changes since last review feature" which is super helpful as a reviewer. Just something to keep in mind for these long running PRs

0xMakka:
Thanks for the heads up @0xemc, apologies I hadn't realised this was happening to the branch. When pulling the latest from staging are you just merging the latest from staging into the branch rather than rebasing?

0xemc:
> Thanks for the heads up @0xemc, apologies I hadn't realised this was happening to the branch. When pulling the latest from staging I'm guessing you are just merging rather than rebasing and not force pushing?

Yep that's it, I usually go:

> git fetch && git merge origin/staging

Since we're squashing commits when closing PRs the staging branch doesn't get polluted

0xMakka:
@Atmosfearful I've tried to move as much of the router query logic into a hook which carries out the majority of the router logic there. If you get a minute and can review again that would be great. Another thing worth mentioning is - I've now paused work on #1353 for the minute until this branch gets merged, as the sorting logic required for #1353 relies on the dropdown been moved out of the modal in this branch.

Atmosfearful (approved):
Beautiful hook!

The other useEffects that keep router params in sync by invoking the callbacks, these could also be moved in the hook. But this is fine too 💯 

And in my testing everything seems to work great.

PR #1207, opened 2023-06-13
Fix typos

omahs:
Fix typos


Atmosfearful:
Fixed README in https://github.com/KlimaDAO/klimadao/commit/49cd7a6dc2e595bb02934f8b5121243ef1dc846a
Other typo was fixed in our translation platform and should arrive in the next release.
Thanks for catching these!

Issue #1208, opened 2023-06-13
[retirements] Create overview of last five retirements


ladytrekker:
## Retirements overview for address

**URL**
- `carbonmark.com/retire`

**New Design:**
- [FIGMA](https://www.figma.com/file/L8DIYEw8LbYHyBhUIQa2AY/Retire?type=design&node-id=627-5649&t=LotHihQMtZtL3FNw-0)

<img src="https://github.com/KlimaDAO/klimadao/assets/95881624/30d13bb7-7d0c-4d8f-92bc-709b98bc1fd5" width="600" />

## Log In Status
- [ ] Logged in only

## Technical requirements / Questions
- [ ] Status of the current address => What is the Matrix here ?? How de we know if an address is at place "426th" ??
- [ ] => OUT OF SCOPE !

## TODOs
- [x] On client: Check if logged in, show Log-in card otherwise
- [x] On client: Show loading spinner while getting all retirements, show empty state if none
- [x] On client: get the last five retirements 
- [x] On client: for each retirement get the corresponding project info
- [x] Render last five retirement infos
- [x] Link each retirement to retirement receipt in new tab => example [URL](https://www.carbonmark.com/retirements/0xab5b7b5849784279280188b556af3c179f31dc5b/2) 
- [x] Consider empty states !! Missing in designs (NOTED...professor working on this)
- [x] Link "View all retirements" => URL ?? (=> https://carbon.klimadao.finance/retirements/beneficiary)
- [x] Render status of the current address => OUT OF SCOPE

jabby09:
@ladytrekker empty state [design](https://www.figma.com/file/L8DIYEw8LbYHyBhUIQa2AY/Retire?type=design&node-id=673-9125&t=q9MpV43tX7XeJZLA-4)

Note...subject to change with input from creative team

ladytrekker:
@jabby09 @theprofessora 

Is it possible to just use the same components on the same page for all browser sizes?

I do not see the benefit of integrating special checks like (are you small, are you bigger) for rendering the same data in a completely different manner?

Or is there a special idea behind this retirement list being rendered as an activity (from a project pool design) instead?

### CURRENT FIGMA DESIGN

![Bildschirmfoto 2023-06-19 um 16 59 23](https://github.com/KlimaDAO/klimadao/assets/95881624/33b6e200-746d-47f2-8dcb-aab77ad74d3d)

### USE SAME COMPONENT INSTEAD

[...]

Issue #1074, opened 2023-04-24
[sanity] Change date format


jabby09:
[Discord thread feedback](https://discord.com/channels/841390338324824096/1099956633376923699)

DOMAIN = both

> Could we maybe standardise the date format on the Resources pages to one that's more intuitive? The 'MM/DD/YYYY' format is confusing for a lot of people. I would suggest 'Month DD, YYYY' (e.g. April 11, 2023), as that's the format we use in social copy, blogs, etc., but if it's got to be numerals only then 'DD/MM/YYYY' would definitely be more user friendly.

![image](https://user-images.githubusercontent.com/103535195/233991994-8775f08a-c37b-4e36-ba8a-6aaa532583d4.png)

Seems to fine on the article pages themselves, so this change would apply to https://www.carbonmark.com/resources and the dates on the article cards.

jabby09:
@PaulineAnnBar hello and welcome to the project

atmos asked to have you this one assigned to you, let me know if any questions.

PR #1224, opened 2023-06-19
Retire: Show retirement list and totals

ladytrekker:
## Description

URL:
https://carbonmark-git-lady-retire-list-klimadao.vercel.app/retire

### Design Changes

Please read the ticket description first => #1208

- A table has to be responsive in order to work an all possible browser sizes
- Cropping text makes no sense when on smaller screens the texts breaks anyways in multiple rows
- ~I kept the same components for every possible screen size (=> see [comment here](https://github.com/KlimaDAO/klimadao/issues/1208#issuecomment-1597356974))~
- => Updated according to this [comment](https://github.com/KlimaDAO/klimadao/issues/1208#issuecomment-1597493624)
- Dates need to work on all possible languages (longer text, shorted text)
- I used a different date format as discussed [here](https://github.com/KlimaDAO/klimadao/issues/1074)

## Related Ticket

Closes https://github.com/KlimaDAO/klimadao/issues/1208

## Checklist


jabby09 (changes requested):
@ladytrekker looking good. Agree on the date change format, good call out there.

This is what I noticed in reviewing:
- [x] For my wallet, not showing the last 5, it looks like it is showing all the retirements in the table
- [x] `View all Retirements` should be `VIEW ALL RETIREMENTS` according to the design

carbonmark/components/pages/Retire/Activity/Table.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+      {props.retirements &&
+        props.retirements.map((r) => (
+          <Link
+            key={r.id}
+            href={createReceiptLink(props.address, r.index)}
+            target="_blank"
+          >
+            <div className={styles.listItem}>
+              <Text>{r.offset.name || r.offset.projectID}</Text>
```
Atmosfearful, line 53: we are now running a carbon-projects CMS that has normalized names (with more consistent capitalization and punctuation). Would be preferable to use those names. Unfortunately that adds some complexity around merging in the names w/ the retirements array.

We can query the CMS directly from the client for v1, that's no problem, but we can also think about an `/retirements` endpoint which would handle caching and reduce the cms costs. Perhaps you find it easier to just write a rudimentary `pages/api/retirements` endpoint right now to keep the react component simple?

Atmosfearful (changes requested):

carbonmark/components/pages/Retire/Activity/Table.tsx
ladytrekker: Done

Atmosfearful (approved):

PR #1225, opened 2023-06-19
tooltips and link added

### LINKED ISSUE #969: [project details] Add tooltips to project detail page
There are a bunch of concepts that could use tooltips, most notably the far fight chart showing project/vintage supply and retirements: 
image

Draft tooltip text:
Total Retirements - "Amount of credits from this project/vintage combination that have been retired."
Remaining Supply - "Amount of credits that have been bridged from this project/vintage combination but not yet retired."

Would also be good to link out to the Polygonscan project/vintage token page here so people can see the stats themselves

psparacino:
## Description

Added explanatory tooltips and link to project details page.

One style note where there was a discrepancy:

The figma tooltip design:
Screen Shot 2023-06-19 at 1 51 30 PM
didn't match current tooltips already on the project page:
Screen Shot 2023-06-19 at 1 52 12 PM

I matched the tooltip design to what currently exists on the project page for consistency. If that needs to be changed let me know.

## Related Ticket

Resolves #969 

## Changes

| Before  | After  |
|---------|--------|
|Screen Shot 2023-06-19 at 12 44 05 PM
 |Screen Shot 2023-06-19 at 12 43 17 PM|

## Checklist

### DIFF
--- a/carbonmark/components/Stats/StatsBar.tsx
+++ b/carbonmark/components/Stats/StatsBar.tsx
@@ -40,6 +40,7 @@ export const StatsBar: FC<Props> = (props) => {
             </Text>
             <TextInfoTooltip
               className={styles.tooltip}
+              align="start"
               tooltip={t`Amount of credits from this project/vintage combination that have been retired.`}
             >
               <InfoOutlined className={styles.tooltipIcon} />
@@ -59,6 +60,7 @@ export const StatsBar: FC<Props> = (props) => {
             </Text>
             <TextInfoTooltip
               className={styles.tooltip}
+              align="start"
               tooltip={t`Amount of credits that have been bridged from this project/vintage combination but not yet retired.`}
             >
               <InfoOutlined className={styles.tooltipIcon} />
--- a/carbonmark/components/Stats/StatsBar.tsx
+++ b/carbonmark/components/Stats/StatsBar.tsx
@@ -13,7 +13,7 @@ import * as styles from "./styles";
 interface Props {
   currentSupply: Project["currentSupply"];
   totalRetired: Project["totalRetired"];
-  projectAddress: string;
+  projectAddress?: string;
 }
 
 export const StatsBar: FC<Props> = (props) => {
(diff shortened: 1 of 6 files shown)


carbonmark/components/Stats/index.tsx
```diff
@@ -13,6 +13,7 @@ interface Props {
   currentSupply?: Project["currentSupply"];
   totalRetired?: Project["totalRetired"];
   description: string;
+  projectAddress?: string;
```
ladytrekker, line 16: Just wondering why this is optional?
From the API it can not be, I guess.
Also constructing a link to polygonScan would not work if this undefined, right? 🤔

carbonmark/components/Stats/StatsBar.tsx
```diff
@@ -1,14 +1,19 @@
...
 import { Project } from "lib/types/carbonmark";
+import Link from "next/link";
 import { useRouter } from "next/router";
 import { FC } from "react";
 import * as styles from "./styles";
 
 interface Props {
   currentSupply: Project["currentSupply"];
   totalRetired: Project["totalRetired"];
+  projectAddress?: string;
```
ladytrekker, line 16: Same here, why optional.

ladytrekker (approved):
Thanks 🌟  
See my comments below concerning the types.

jabby09 (changes requested):
@psparacino the tool tip styles not matching design on current implementation was an oversight when the methodology tool tip went live - good catch!

the [design](https://www.figma.com/file/5GewbdVaM1Q06CwHmF7d3F/%2Fprojects?type=design&node-id=11-59064&t=TnhYFKfHUqQGcMfN-4) is pretty clear how these should look on the front end. For these info icons in particular, they should be more grey than black.

ideally we could fix this now OR I can create a new ticket?

carbonmark/components/Stats/index.tsx
psparacino: Stats is also used in the portfolio sidebar that didn't require the projectAddress; if it wasn't optional it was breaking the build

carbonmark/components/Stats/StatsBar.tsx
psparacino: ah yes, pushed a fix

psparacino:
@jabby09 Not further ticket necessary, can just fix here. I had been matching the current icon color/style to the methodology tooltip currently on the page, but changed methodology, retirements and supply tooltip style and icon color to all be consistent with each other and design.

Also edited the Tooltip component in this last commit to match the left alignment here (previously only allowed center alignment).

jabby09 (approved):
Looks good!

Atmosfearful (approved):

PR #1232, opened 2023-06-21
Refactor /projects/:id API handler

### LINKED ISSUE #1156: [carbonmark-api] Identify default pool projects
# Summary
API responses should indicate if a pooled project is the default for that pool, and show a lower price.

# Background
Every pool has 1 'default' project.
The default project is cheaper to retire and redeem (because it does not cost a % redemption fee).
Currently, **we do not differentiate the default projects**.

Going forward, we should.
- We need to know if the selected project is a default so we can choose the the correct retirement aggregator method `toucanRedeemPoolDefault` instead of `toucanRedeemPoolSpecific`.

Here are the defaults for each pool. It's the project with the oldest vintage.
- [BCT](https://www.carbonmark.com/projects/VCS-191-2008) - 0xb139c4cc9d20a3618e9a2268d73eff18c496b991
- [NCT](https://www.carbonmark.com/projects/VCS-1529-2012) - 0x6362364a37f34d39a1f4993fb595dab4116daf0d
- [UBO](https://www.carbonmark.com/projects/VCS-1140-2015) - 0xd6ed6fae5b6535cae8d92f40f5ff653db807a4ea
- [NBO](https://www.carbonmark.com/projects/VCS-981-2014) - 0xb6ea7a53fc048d6d3b80b968d696e39482b7e578

# Requirements for `/api/projects/[project_id]`
* [ ] Indicate if a project in the `prices` array is the pool default with a boolean.
* [ ] Show the discounted price, excluding all fees. This should match the price for 1 tonne shown in the klima app ui:
image

# Requirements for `/api/projects`
* [ ] show the discounted price, same as above

# Considerations
This method `getScoredTCO2s` on BCT and NCT tells us the projects held in the pool. The first address returned in the array is the default project.
https://polygonscan.com/address/0x2F800Db0fdb5223b3C3f354886d907A671414A7F#readProxyContract#F21
For UBO and NBO this method is called `getERC20Tokens`
https://polygonscan.com/address/0x2b3ecb0991af0498ece9135bcd04013d7993110c#readProxyContract#F11

I am not aware of a better way to determine which projects are default. Subgraph perhaps?

There are a few ways we can implement this, each has tradeoffs.
The default project should only change every few months, so whichever solution we use, it should be cached, and **not invoked every request**.
Idea 1 - fetch from smart contracts directly - cache the result in memory and reuse across all requests.
Idea 2 - fetch defaults at build time (daily) or first run-time (when the microservice instantiates)
Idea 3 - hardcode. I think this is the easiest option for now as volumes are low. When a default project runs out of pool supply, it will disappear from the listings, and should cause no harm.

Atmosfearful:
## Description

I could not figure out a clean way to update the pricing to reflect default projects in the current code. So I refactored the entire handler.

- reduces the total number of roundtrip network requests significantly:
  - instead of 4 graphql queries (one for each pool), we do 1 query total (`fetchAllPoolPrices`)
  - removes the c3token and tco2token requests, these are not needed
  - parallelizes the firebase requests
- eliminates / consolidates a lot of code into utils
- move to a single-source-of-truth object/map for pool info

## Breaking changes:
- changed some api property names.
  - prices array more clear property names:
    - `name` -> `poolName`
    - `leftToSell` -> `supply`
    - `tokenAddress` -> `poolAddress`
    - `poolTokenAddress` -> `projectTokenAddress`
  - moved `currentSupply` `totalBridged` and `totalRetired` into a field called `stats`
  
## Performance

Performance has only improved slightly. We need to dig deeper into which requests are taking the longest.
Occasionally production would give a 3-6 second load time, and this does not seem to happen as often after this PR. But I can't reliably reproduce this.

### Before PR: Production (api.carbonmark.com)
```
// VCS-191-2008
getCarbonmarkProject: 919.481ms
// VCS-1883-2018
getCarbonmarkProject: 858.302ms
// VCS-1140-2015
getCarbonmarkProject: 1.142s
```

This pr (using preview link).
Time for initial fetch only:
```
// VCS-191-2008
getCarbonmarkProject: 677.088ms
// VCS-1883-2018
getCarbonmarkProject: 783.747ms
// VCS-1140-2015
getCarbonmarkProject: 609.366ms
```
  
## Related Ticket

Resolves #1156


carbonmark-api/helpers/fetchProjectDetails/index.js
```diff
@@ -0,0 +1,21 @@
+import { fetchGSProject } from "./fetchGSProject";
+import { fetchVCSProject } from "./fetchVCSProject";
+
+export const fetchProjectDetails = async ({ key }) => {
+  try {
+    const [registry, registryProjectId] = key.split("-");
```
0xemc, line 6: **nit:** Perhaps the string manipulation should be outside of this function so it's not tied to the format of project keys?

carbonmark-api/helpers/fetchProjectDetails/index.js
```diff
@@ -0,0 +1,21 @@
...
+    if (!registry || !registryProjectId) return null;
+
+    const registryUpper = registry.toUpperCase();
+    if (registryUpper === "VCS") {
```
0xemc, line 10: Without having to go down the refactor road too much it would be nice to have these magic strings as constants somewhere

carbonmark-api/helpers/fetchMarketplaceListings.js
```diff
@@ -0,0 +1,100 @@
...
+      GET_PROJECT_BY_ID,
+      { key, vintageStr: vintage }
+    );
+
+    const project = data?.projects?.[0] || null;
+    const filteredListings =
+      project?.listings?.filter(
+        (listing) =>
+          Number(listing.leftToSell) > 1 && !!listing.active && !listing.deleted
+      ) || [];
```
0xemc, line 43: **nit:** I like to extract these predicates to keep things clearer and isolated

```suggestion
   const isListingActive = (listing) => Number(listing.leftToSell) > 1 && !!listing.active && !listing.deleted
    const filteredListings =
      project?.listings?.filter(isListingActive) ?? [];
```

carbonmark-api/helpers/fetchProjectDetails/fetchVCSProject.js
```diff
@@ -0,0 +1,38 @@
...
+    const registry = "VCS";
+    const project = await sanity.fetch(fetchProjects, {
+      registry,
+      registryProjectId,
+    });
+
+    if (!project) return null;
+
+    return {
+      key: `${registry}-${registryProjectId}`,
```
0xemc, line 22: 😢 magic string formats

carbonmark-api/helpers/formatUSDC.js
```diff
@@ -0,0 +1,6 @@
+import { utils } from "ethers";
+
+export const formatUSDC = (str) => {
+  const decimals = process.env.VERCEL_ENV === "development" ? 18 : 6;
```
0xemc, line 4: Why is it that we would want 18 decimals in development anyway?

0xemc (approved):
Much cleaner,  some small suggestions but otherwise all good.

Though I suggest writing some additional postman or tap tests to confirm we are getting the same results as previously

carbonmark-api/helpers/formatUSDC.js
Atmosfearful: Najada distributed test "USDC" tokens on Mumbai from another project but they are 18 decimal :(

If we want to go all-in on testnet QA then we should create a 6 decimal one and use that instead. but its high effort low reward rn as testnet is always a pain (there are no carbon pools on testnet)

0xemc (approved):

PR #1257, opened 2023-06-29
Fix #1252 Don't add slippage when retiring same token

### LINKED ISSUE #1252: [App] Retire MCO2 with MCO2 reverts due to bad inputs
A user reported a bug which I was able to reproduce as follows.
When attempting to retire MCO2 using MCO2 as the payment method, the retirement transaction reverts (see screenshot below).
image

The approval [transaction](https://polygonscan.com/tx/0x6e2bdc0730148a8a8e309bdf27bc53a91c73dcd3ab8efae4606974424dc37a03) for a 0.1 tonne retirement goes through with the amount `0.10201`. This amount should be `0.101`.
According to the Solidity tests of the RA v2, the function `getSourceAmountDefaultRetirement` should be called to find the correct `maxAmountIn` parameter amount. When calling this function for mco2 with the amount 0.1 tonnes, I get the amount corresponding to 0.101 tco2:
image

However, it looks like the amount passed to the actual retirement transaction is 0.10201, which is not exactly equal to 0.101, causing the contract to revert on line 52:
image

I'm not sure if this is the actual value used, but I found that the value passed to the retirement transaction may just be the incorrect approval amount instead of the value from `getSourceAmountDefaultRetirement`:
image

Atmosfearful:
## Description

Fix #1252

## Checklist

### DIFF
--- a/app/components/views/Offset/index.tsx
+++ b/app/components/views/Offset/index.tsx
@@ -285,6 +285,10 @@ export const Offset = (props: Props) => {
       return quantity;
     }
     if (!cost || cost === "loading") return "0";
+    if (paymentMethod === selectedRetirementToken) {
+      // for example, when using mco2 to retire mco2
+      return cost;
+    }
     const onePercent = utils
       .parseUnits(cost, getTokenDecimals(paymentMethod))
       .div("100");


ShimonD-KlimaDAO:
I verified that this fix works.

PR #1279, opened 2023-07-05
0xMakka/1175 add error message styling

### LINKED ISSUE #1175: [forms] On CM forms, show red ring around invalid fields
Would be helpful if @theprofessora can just do a quick example text input component somewhere in figma for us to reference
It could be a red ring, or a red text label, or red star...

Currently its not very clear which input is invalid
image

0xMakka:
## Description

Adds an error message to the beneficiary wallet address field. 

@Atmosfearful & @jabby09 I've added a quick fix here to address the beneficiary wallet address but I think it's worth investing some time in refactoring this component to use react-hook-form like the Retire & Buy forms. Let me know if you guys want me to tackle it as part of this ticket, or roll out with this quick fix and pick up in another ticket.

## Related Ticket

Resolves #1175

## Checklist

### DIFF
--- a/carbonmark/components/pages/Portfolio/Retire/RetireForm/index.tsx
+++ b/carbonmark/components/pages/Portfolio/Retire/RetireForm/index.tsx
@@ -267,19 +267,15 @@ export const RetireForm = (props: RetireFormProps) => {
                         value: retirement.beneficiaryAddress,
                       }}
                       label={"beneficiaryAddress"}
+                      errorMessage={
+                        retirement.beneficiaryAddress &&
+                        !ethers.utils.isAddress(retirement.beneficiaryAddress)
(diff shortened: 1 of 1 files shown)


jabby09 (approved):

jabby09:
> @Atmosfearful & @jabby09 I've added a quick fix here to address the beneficiary wallet address but I think it's worth investing some time in refactoring this component to use react-hook-form like the Retire & Buy forms. Let me know if you guys want me to tackle it as part of this ticket, or roll out with this quick fix and pick up in another ticket.

thanks @0xMakka I approved this fix, let's push it out.

I'd suggest if that form needs a bit of refactor we do it as a separate ticket.

Atmosfearful (approved):

Issue #1234, opened 2023-06-22
[retire from pool] Pay with Credit card


jabby09:
Updated [figma designs here](https://www.figma.com/file/5GewbdVaM1Q06CwHmF7d3F/%2Fprojects?type=design&node-id=890-29468&mode=design&t=JLl8iTo1R7ThUqRz-4).

See docs (https://checkout.offsetra.com/) for API documentation if required.

`Stretch goal:` during form validation, fetch the balance of the amount of USDC in the offsetra server wallet which then may allow / support transactions greater than 2000. Please reach out to @Atmosfearful to discuss in more detail.

Note that designs are provided for desktop and mobile screens.

**QUESTIONS:**

Why not move the selection of the payment method to the top of the form?
Because depending on which method the user selects, the text changes, the validation changes etc....

## TODOS
- [x] deploy Transaction Payment code to Offsetra production (currently only on Staging)
- [x] Change Logo on Stripe

## Requirements

- [x] Pay with credit card will be supported for `RETIRE NOW` function on **pool listings only**.
![Screenshot 2023-06-26 at 1 37 39 PM](https://github.com/KlimaDAO/klimadao/assets/103535195/d2ca58af-bfff-418f-a1bf-82660364ca7f)

- [x] In the `RETIRE NOW` on pool listings flow, the **Pay with** dropdown needs to be updated to support selecting Credit Card. Note that for all other BUY and RETIRE flows, the Credit Card option in the dropdown will remain greyed out with Coming Soon notation.
![Screenshot 2023-06-26 at 1 38 13 PM](https://github.com/KlimaDAO/klimadao/assets/103535195/7570d17b-214d-4cb0-adc9-f5f2f306ba61)

- [x] When the user selects Pay with Credit Card, the description above the drop down changes from Balance xxxxx to $2,000 maximum for credit cards
![Screenshot 2023-06-26 at 1 38 44 PM](https://github.com/KlimaDAO/klimadao/assets/103535195/386584ba-b8e5-439b-a6db-18949227a94a)

- [x] As the user selects the number of tonnes to retire, the form will need a validation to ensure the total cost does not exceed $2,000. Otherwise, show the following error state:
![Screenshot 2023-06-26 at 1 39 07 PM](https://github.com/KlimaDAO/klimadao/assets/103535195/32fb0703-8923-49b1-91d3-c13886863437)

- [x] CAUTION: Only integers are allowed as tonnes

[...]

PR #1280, opened 2023-07-05
Retire from Pool with Credit Card

ladytrekker:
## Open TODOs
- [x] **_AGAIN:_** deploy Offsetra Staging to Offsetra Production
- [x] Change Logo for Stripe (done and merged to Offsetra Staging) ✅ 
- [x] Again this Question: Where should this link "Learn how to acquire USDC on Polygon." go to ?
- [x] Pass the correct projectTokenAddress to Offsetra => should be correct now, **please double check** !

## QUESTIONS:
- [ ] Are the fees correct? Currently calculating the same as for USDCs payments
- [ ] Or rather take the API response "networkFeeCents". and hide "Show details" all together?

## Design Changes
- Error messages for required fields
- Error Message below Total Costs in Sidebar too (aligned with all other views, also otherwise not visible on mobile screens when the user scrolls down to the submit button)
- Show the real wallet balance of the Server Fiat Wallet
- Show Credit Card Icon (aligned with when choosing tokens)
- Do not format to price for fees (or do you prefer numbers with dollar sign like $0.00045)

## Related Ticket

Resolves https://github.com/KlimaDAO/klimadao/issues/1234

## Checklist


carbonmark/components/pages/Project/Retire/Pool/RetireInputs.tsx
```diff
@@ -9,35 +10,93 @@ import { TextareaField } from "components/shared/Form/TextareaField";
...
+      min: {
+        value: 1,
+        message: t`The minimum amount to retire is 1 Tonnes`,
+      },
+    },
+    totalPrice: {
+      max: {
+        value: Number(fiatBalance || "2000"),
+        message: t`At this time, Carbonmark cannot process credit card payments exceeding ${formatToPrice(
+          fiatBalance || 0
```
Atmosfearful, line 63: I'm having second thoughts about this. I think we should go back to only referencing the MAX_USDC value. We can expose the max value through the new endpoint you created, so Offsetra can change it without having to update the klimadao code.

We can remove the balance checks completely.
- Offsetra and team should always ensure there is at least $2k or more in liquid USDC.
- Clients should not have to care about this balance at all, they just need to know the max allowed. In very rare cases where it gets bought out, checkouts will simply fail until Offsetra can replenish it. They have alerts set up already.
- The rule should actually become `$2k per user per day` or `${MAX_USDC} per user per day` and we should update the error message to say that instead. And some day in the future we could even enforce this by looking up payment records in firebase.

lib/utils/getFiatUrls/index.ts
```diff
@@ -0,0 +1,11 @@
+import { FIAT } from "../../constants";
+
+export const getFiatRetirementUrl = (params: { isProduction: boolean }) =>
+  params.isProduction
+    ? FIAT.retirementCheckout.production
+    : FIAT.retirementCheckout.staging;
+
+export const getFiatBalanceUrl = (params: { isProduction: boolean }) =>
```
Atmosfearful, line 8: we will remove the credit card from KlimaDAO in the near future. So then this wouldn't need to live in `lib` it could actually just go in carbonmark/lib/constants `urls` object like our other ones. this is more clean/consistent IMO.

We can put everything in carbonmark now, even if that results in code duplication in this PR, or we can document it in a ticket and clean it up later

Atmosfearful (changes requested):

jabby09 (changes requested):
hi @ladytrekker here a few things I noticed:

- Can I assume the 4.78 maximum is related to being in stripe test mode?
- Also assume the the stripe logo change will happen when hooked up to real stripe account?
- For this question: Where should this link "Learn how to acquire USDC on Polygon." go to? URL target would be https://docs.carbonmark.com/get-started/how-to-get-usdc-or-matic

Questions:
1. Is it true for credit card the minimum to retire is 1 tonne? (we support < 1 tonne for usdc)
2. Why is retirement message a required field?

Change:
- [ ] I don't think we need the credit card logo here, basically the $ sign replaces the USDC logo.

Screenshot 2023-07-10 at 11 05 52 AM

carbonmark/components/pages/Project/Retire/Pool/RetireInputs.tsx
```diff
@@ -175,21 +275,29 @@ export const RetireInputs: FC<Props> = (props) => {
...
+          </Text>
+        )}
+
         <div className={styles.paymentHelp}>
           <HelpOutline className={styles.helpIcon} />
           <div className={styles.paymentText}>
             <Text t="body3">
-              {t`To retire this project using a different form of payment,`}{" "}
+              {t`Currently, Carbonmark only accepts Polygon USDC or Credit Card Payments. To retire this project using a different form of payment,`}{" "}
               <Anchor href={urls.app}>click here</Anchor>.
```
Atmosfearful, line 301: Complete string should be in the tag for translators. also "click here" is bad a11y practice, lets change the text:
Edit: I've asked for thumbs up from product team to remove this help text altogether. I don't think its needed.
```suggestion
              <Trans>Currently, Carbonmark only accepts Polygon USDC or Credit Card Payments. To retire this project using a different form of payment, <Anchor href={urls.app}>visit app.klimadao.finance</Anchor>.</Trans>
```

carbonmark/components/pages/Project/Retire/Pool/TotalValues.tsx
```diff
@@ -81,13 +89,31 @@ export const TotalValues: FC<TotalValuesProps> = (props) => {
           inputToken: paymentMethod,
           retirementToken: poolName,
           quantity: amount,
-          isDefaultProject: isPoolDefault,
+          isDefaultProject: isPoolDefault, // TODO !! Pass projects address here for pool projects!
```
Atmosfearful, line 92: is this still a TODO?

Atmosfearful (changes requested):

Atmosfearful (approved):
I would say this is mergeable. Some of my changes related to moving files were not addressed but these would fit better in a follow up PR when we remove credit card from klimadao. `checkout` api enhancements are now on production and seem to be working well.

carbonmark/components/pages/Project/Retire/Pool/RetireInputs.tsx
ladytrekker: I replaced the balance endpoint on the Offsetra API to a new endpoint "info".
This endpoint gives back the MAX_USDC value.

When on Production the value of MAX_USDC is hardcoded to 2000.
When on Staging the actual server wallet balance is returned. 
Otherwise the UI does not make sense (Api throws very quickly, but everywhere it says $2.000 ? Rather show here the real max balance of the connected Server Wallet.)

lib/utils/getFiatUrls/index.ts
ladytrekker: Moved to Carbonmark. Deleted on klimadao/lib

carbonmark/components/pages/Project/Retire/Pool/RetireInputs.tsx
ladytrekker: I changed the text now with the correct blog links etc according to the Figma designs.

We can remove everything later again when the feedback from product came back.

jabby09:
> Same as on `app.klimadao.finance`, IMO its a best practice to always include a message, users who dont want a message or beneficiary can add an empty space

@Atmosfearful @ladytrekker I don't think that is good UX to have to know that to skip that required field I need to add an empty space. Also we have moved away from that in the other forms, so RETIRE from portfolio and pool listing do not require a beneficiary or a message

jabby09:
> Correct, and whole-tonne values only. We could enhance the offsetra API to support kilogram instead (medium/low effort change).

@Atmosfearful @ladytrekker part of why I asked that question is that is pay with is USDC we support < 1 tonne. That could be confusing for a user so we might need some helper messaging around this.

jabby09 (changes requested):
@ladytrekker sorry I meant to include this in my feedback yesterday, can we please make Credit Card the default option in the Pay with: drop down on the RETIRE forms?

jabby09 (approved):
I think this is could to merge and release.

Can follow up on anything remaining in a future PR.

PR #1294, opened 2023-07-09
fix: 1289 createData on POST /users

### LINKED ISSUE #1289: Creating a carbonmark user is not working
## Description

1. Go to https://www.carbonmark.com/users/login
2. Connect your wallet with a new address (which has not been registered on Carbonmark yet)
3. Click on "Create profile"
4. Create a Profile
5. Reload page => State is "Not registered"
6. Check the API response for your user
7. => `https://api.carbonmark.com/api/users/<YOUR_ADDRESS>?type=wallet`
8. Result: User is not created correctly: ID is missing, name is missing, handle is missing etc
9. Example link: https://api.carbonmark.com/api/users/0x93aEE3622b0A6F6bAb531F8980Fb811F4E824b1c?type=wallet
10. How the response should look like: See [types](https://github.com/KlimaDAO/klimadao/blob/staging/carbonmark/lib/types/carbonmark.ts#L99-L107)

0xemc:
## Description

The `createData` object was being passed as an attribute rather than as the response.

## Related Ticket

Resolves #1289 

## Todo
Test to come in future PR


jabby09 (approved):
confirmed this is now working

jabby09 (changes requested):
@0xemc I did mark this as approved but now that I look again, not sure the profile form changes are being persistently saved.

I tried creating my profile like this...I get the torus signing modal etc.
Screenshot 2023-07-10 at 3 13 38 PM

This is the result, however, don't see the profile description updated:
Screenshot 2023-07-10 at 3 13 54 PM

If I then click away from profile (like visit the marketplace) and then return to profile my updates do not appear:
Screenshot 2023-07-10 at 3 14 05 PM

Atmosfearful (approved):

PR #1296, opened 2023-07-10
fix: 1293 Import aliases in carbonmark-api

### LINKED ISSUE #1293: [API] Add root import paths to tsconfig
Title: Add Root Import Paths to tsconfig.json

Description:

As our project grows, managing relative paths for module imports has become increasingly complex. To simplify this, we need to add root import paths to our tsconfig.json file. This will allow us to import modules using absolute paths, which will make our code cleaner and easier to manage.

Steps to reproduce:
1. Open any TypeScript file in the project.
2. Try to import a module using an absolute path.

Expected behavior:
The module should be imported successfully without any errors.

Actual behavior:
An error is thrown saying the module cannot be found.

Solution:
Update the "compilerOptions" section in the tsconfig.json file with the "baseUrl" and "paths" keys. The "baseUrl" should be set to "./" and the "paths" should contain the mapping between the alias and the actual location of the module.

Example:
```json
"compilerOptions": {
  "baseUrl": "./",
  "paths": {
    "@/*": ["src/*"]
  }
}
```

This issue requires testing after the update, by trying to import a module using the new alias to ensure everything is working as expected.

0xemc:
## Description

Add the mode-alias package to allow aliased imports

## Related Ticket

Resolves #1293


Atmosfearful (changes requested):
Awesome I agree it's worth ditching relative imports if we can.

The convention established so far (except the Sanity Studio packages) has been to use absolute paths with no alias, which is set in tsconfig via the `baseUrl` property

ie, for importing within a package:
```ts
import { Component } from "components/Component";
import { myUtil } from "utils/myUtil";
```

And for importing across packages, we use the alias defined in package.json for each package (but in practice only share code from `@klimadao/lib`)
```ts
import { myUtil } from "@klimadao/lib/myUtil";
```

So for this pr:
- Any reason we don't stick to this convention for `carbonmark-api`?
- Do we really need `module-alias` package? I'm confident ts gives us what we need out of the box...

0xemc:
> Awesome I agree it's worth ditching relative imports if we can.
> 
> The convention established so far (except the Sanity Studio packages) has been to use absolute paths with no alias, which is set in tsconfig via the `baseUrl` property
> 
> ie, for importing within a package:
> 
> ```ts
> import { Component } from "components/Component";
> import { myUtil } from "utils/myUtil";
> ```
> 
> And for importing across packages, we use the alias defined in package.json for each package (but in practice only share code from `@klimadao/lib`)
> 
> ```ts
> import { myUtil } from "@klimadao/lib/myUtil";
> ```
> 
> So for this pr:
> 
> * Any reason we don't stick to this convention for `carbonmark-api`?
> * Do we really need `module-alias` package? I'm confident ts gives us what we need out of the box...

Fair enough I've updated the alias to be `src/`. Unfortunately the way the module-alias package works is it does a string replacement looking for the defined alias (in this case src/) so a wild card root isn't really possible.

The reason ts paths is not sufficient is because node has no knowledge of the paths defined in tsconfig. The resulting js built by tsc and isnt computed correctly by node.

In the webpack/next environment my understanding is that it's the bundler that handles these aliases correctly.

If we switch to ts-node rather than plain tsc, these aliases would be resolved via [tsconfig-paths](https://www.npmjs.com/package/tsconfig-paths) again another package but we would get the wildcard behaviour that you mentioned.

ladytrekker (approved):
Thanks. And learned something: NodeJS needs module-alias on top of the TS config paths. Makes sense.

PR #1310, opened 2023-07-12
chore: Carbonmark to be deployed via github workflow

0xemc:
## Description

In order to gain more control over the deployment process:

* Manually build and deploy Carbonmark via a github workflow
* Add a comment to the PR with a link to the preview deployment

## Todo
- [ ] Disable default vercel deploy


sprrwhwk:
We can consider adding a path filter to the workflows. This would for example allow us to only build & deploy carbonmark if something changed inside the `[ '/carbonmark', '/carbonmark-api' ]` paths. Same goes for the other projects.

See [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) for the docs.

PR #1313, opened 2023-07-12
portfolio retire: do not redirect on initializing

ladytrekker:
## Description

Retiring from Portfolio asset always ended in a redirect back to the main portfolio page.

This is the fix.

## Related Ticket

none

## Checklist


jabby09 (approved):
good catch!

fix worked

PR #1316, opened 2023-07-13
Retire: do not continue when costs is invalid

ladytrekker:
## Description

When the costs could not be calculated, the former value of the total costs was not resetted.
Therefore the user could continue with a false total cost value.

Now, the total costs are set to "0" when the API for the total costs calculation throws.
And the form is invalid.

## Related Ticket

none

## Checklist


jabby09:
@Atmosfearful nothing for the QA team to do here until this is merged

jabby09:
@0xemc do you have PR review & merge powers now?

If yes, this one is important given it impacts the credit card payment release we did.

PR #1321, opened 2023-07-14
add torus link below wallet address

### LINKED ISSUE #998: [wallet] Add a View Wallet On Torus button
Users using torus should be shown how to get to their wallet on polygon.tor.us to manage their other crypto assets there.

This might be a good spot, in the sidebar?
image

Keep in mind this should be reuseable for other web wallets like Circle in the future.

sprrwhwk:
## Description

This PR adds a new component, the `ViewWalletButton` that, when clicked, leads people to their torus wallet. It also renames the `CopyAddressButton` into a more generic `CopyValueButton` which behaves exactly like before.

To keep consistency with different wallets, it also introduces a component that takes the name of the previous `CopyAddressButton`. It displays the address as a text and has a separate copy button.

## Related Ticket

Resolves #998 

## Checklist


carbonmark/components/Layout/AddressSection/index.tsx
```diff
@@ -1,32 +1,50 @@
...
+            address={props.address}
+            className="copyButton"
+          />
+          {true /* isTorusWallet */ && (
+            <Anchor
+              className={styles.iconAndText}
+              href={`https://polygon.tor.us/`}
+            >
+              <Text className={styles.externalLink} t="body2" uppercase>
+                View <LaunchIcon />
```
sprrwhwk, line 43: hey @ladytrekker, what are the steps required to prepare this string for lingui?

lib/utils/concatAddress/index.ts
```diff
@@ -1,7 +1,7 @@
 /**
  * @example
- * // 0x123..456
+ * // 0x123…456
  */
 export const concatAddress = (address: string) => {
-  return address.slice(0, 5) + "..." + address.slice(address.length - 3);
+  return address.slice(0, 5) + "…" + address.slice(address.length - 3);
```
sprrwhwk, line 6: this change is an a11y improvement since screen readers can properly recognize this whereas they might struggle with three full stop characters

sprrwhwk: https://tosbourn.com/the-importance-of-using-the-ellipsis-character/

carbonmark/components/Layout/AddressSection/index.tsx
ladytrekker: You just need to wrap the actual string into Trans or t

So like

```TS
  <Text className={styles.externalLink} t="body2" uppercase>
    <Trans>View</Trans>{" "}<LaunchIcon />
   </Text>
```

lib/components/ViewWalletButton/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+} & ButtonBaseProps;
+
+export const ViewWalletButton: FC<Props> = (props) => {
+  const className = cx(styles.viewWalletButton, props.className);
+
+  return (
+    <div className={className}>
+      <span>{concatAddress(props.address)}</span>
+      <ButtonPrimary
+        href="https://polygon.tor.us/"
```
ladytrekker, line 21: I would pass this url as a prop to make this "ViewWalletButton" compatible with future wallets too

lib/components/ViewWalletButton/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+        target="_blank"
+        rel="noopener noreferrer"
+        label="VIEW"
```
ladytrekker, line 24:
```suggestion
        label={t(`View`)}
```

=> for translators

ladytrekker (commented):
Thanks
See my comments below

jabby09:
@sprrwhwk let me know when you want me to review / qa the front end again

lib/components/ViewWalletButton/index.tsx
sprrwhwk: good point, done!

lib/components/ViewWalletButton/index.tsx
sprrwhwk: it doesn't work, as mentioned above. I think I'm missing something

ladytrekker: Ah, I see it now.

You created the component in @klimadao/lib
In this repo there is no lingui.

I would do this:

1. move ViewWalletButton component to carbonmark/components (it won't be in use somewhere else anyways)
2. Import ButtonPrimary from carbonmark/components/Buttons into this component
3. try again with `label={<Trans>View</Trans>}`
4. Here it will work

sprrwhwk: works now, thanks

sprrwhwk:
@jabby09 ready for review

0xemc (approved):

0xemc (commented):
Not sure what's going on in the preview link but it's looking pretty funky:

Screenshot 2023-07-22 at 7 43 17 am

carbonmark/components/Layout/AddressSection/index.tsx
```diff
@@ -21,13 +25,19 @@ export const AddressSection: FC<AddressSectionProps> = (props) => {
...
-          label={concatAddress(props.address)}
-          address={props.address}
-          className="copyButton"
-        />
-      )}
+      {props.address &&
+        (isTorus ? (
+          <ViewWalletButton
+            address={props.address}
+            href="https://polygon.tor.us/"
```
0xMakka, line 32: Nit: Can remove the hardcoded href and use the url which already exists in the constants file.

```
import { urls } from "@klimadao/lib/constants";

<ViewWalletButton
  address={props.address}
  href={urls.polygonTor}
/>
```

carbonmark/components/ViewWalletButton/styles.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const viewWalletButton = css`
+  color: var(--manatee);
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+  align-items: center;
+`;
+
+export const viewButton = css`
+  gap: 4px !important;
```
0xMakka, line 11: Nit: use rem values

```gap: 0.4rem;```

File: carbonmark/components/ViewWalletButton/index.tsx [hunk 7]
```diff
@@ -0,0 +1,33 @@
...
+  address: string;
+} & ButtonBaseProps;
+
+export const ViewWalletButton: FC<Props> = (props) => {
+  const className = cx(styles.viewWalletButton, props.className);
+
+  return (
+    <div className={className}>
+      <span>{concatAddress(props.address)}</span>
+      <ButtonPrimary
```
0xMakka, line 21: Nit: can remove the closing `</ButtonPrimary>` tag and close off the opening tag:

```
<ButtonPrimary  
  {...props}
/>
```

carbonmark/components/ViewWalletButton/styles.ts
sprrwhwk: I actually took this from figma as-is. Is there anything to worry about when changing it from px to rem in this specific case?

carbonmark/components/Layout/AddressSection/index.tsx
sprrwhwk: good point! thanks

carbonmark/components/ViewWalletButton/index.tsx
sprrwhwk: done, thanks!

carbonmark/components/ViewWalletButton/styles.ts
0xMakka: You can always double check to ensure the computed styles are the pixel value you are looking for in devtools.

For example - padding top & bottom here are set as 0.4rem.
Screenshot 2023-07-22 at 19 17 51

Switch from styles to computed and you can see that both values are 4px
Screenshot 2023-07-22 at 19 18 11

sprrwhwk: Thanks, I'll keep that in mind in the future. Had no idea about the computed tab. Done in [975daa4](https://github.com/KlimaDAO/klimadao/pull/1321/commits/975daa466b367c1a080f93693d7b5ccbcf2a8dc3)

Atmosfearful:
> Not sure what's going on in the preview link but it's looking pretty funky:
> 
> Screenshot 2023-07-22 at 7 43 17 am

I can reproduce this as well when connecting with metamask Chrome extension

carbonmark/components/Layout/AddressSection/index.tsx
```diff
@@ -1,8 +1,10 @@
-import { CopyAddressButton } from "@klimadao/lib/components";
+import { urls } from "@klimadao/lib/constants";
 import { Domain } from "@klimadao/lib/types/domains";
-import { concatAddress } from "@klimadao/lib/utils";
+import { concatAddress, isTorusProvider, useWeb3 } from "@klimadao/lib/utils";
 import { Trans } from "@lingui/macro";
+import { CopyAddressButton } from "components/CopyAddressButton";
```
Atmosfearful, line 5: my best guess is that changing this import instead of `@klimadao/lib/components` is what broke the `CopyAddressButton` styles

Atmosfearful (changes requested):

carbonmark/components/Layout/AddressSection/index.tsx
wowselim: You're right, I should have caught this. Fixed in [e5c55ae](https://github.com/KlimaDAO/klimadao/pull/1321/commits/e5c55ae2278f0e525a0d0b5d59dd812de31c98bd)

sprrwhwk: You're right, I should have caught this. Fixed in [e5c55ae](https://github.com/KlimaDAO/klimadao/pull/1321/commits/e5c55ae2278f0e525a0d0b5d59dd812de31c98bd)

jabby09 (approved):
this matches the design and works as expected.

we now have a design mis-match torus and other wallet users - I've asked professor to update the designs. if there are changes we can do in another ticket.

jabby09 (dismissed):
@sprrwhwk 

as per comment here: https://github.com/KlimaDAO/klimadao/issues/998#issuecomment-1651822474

We should update the wallet login so it matches the font size / style / colour of torus change.

image

sprrwhwk:
@Atmosfearful the new CopyValueButton is also used by app/NavMenu. Should I duplicate it and put the copy in lib/components? Seems hacky…

sprrwhwk:
> @Atmosfearful the new CopyValueButton is also used by app/NavMenu. Should I duplicate it and put the copy in lib/components? Seems hacky…

For the record:

After talking to @Atmosfearful we decided to move the CopyValueButton into the lib project.

sprrwhwk:
@Atmosfearful this grew quite in scope. You might want to have another look at it…

Atmosfearful (approved):

Issue #764, opened 2022-10-27
[tech] Speed up the retirement receipt page load time


Atmosfearful:
After a user creates a carbon offset retirement, and clicks the link to their retirement page, there is a few seconds of delay. ?Let's try to improve this.

* [x] prefetching can help
```html
<link
    rel="prefetch"
    href={`https://klimadao.finance/retirements/${data.beneficiary_address}/${data.retirement_index}`}
/>
```
* [x] add a subgraph poller so we don't try navigating / prefetching the page until we know the record exists in subgraph

Atmosfearful:
Related to #1018

jabby09:
@Atmosfearful I added this to the /retire epic

psparacino:
addressed in: https://github.com/KlimaDAO/klimadao/pull/1331

psparacino:
This issue was addressed in PR #1331

PR #1331, opened 2023-07-18
Tufnel/post retirement refactor 1096

### LINKED ISSUE #1096: [retirements] Check for fresh retirement data before navigating to retirement page
Prerequisite for #1080 
Currently, in our retirement pages, we have some complex fallback logic and loading spinners, because the subgraph can be slow. The goal is to get these removed and simplified in #1018 

In order to accomplish this, we must first make some changes at the end of the retirement flows.

### Context
At the end of a retirement, we will show a receipt page

* [x] On the receipt page, which is We should poll the subgraph to wait for a retirement record to be indexed and appear in subgraph before we navigate to this page, that way the user should never encounter a 404.

psparacino:
## Description

On retirement completion, this polls the subgraph for the block number in the transaction receipt. Once the subgraph has reached the txn block, router.prefetch is called on the receipt page while opening up the success modal.

The poller will try for 5 secs (10 tries at half second intervals). If the subgraph has still not reached the transaction block the modal shows a message (below) that the transaction is successful but is slow to index. There is a link to the successful txn on polygon scan and the link to retire more carbon.

On a side note for the future, there are some medium size design differences in the approve, submit and processing modals between the Pool and Portfolio retirement flows.

## Related Ticket

Resolves #1096 and #764 

## Changes

| Pool Notification  |
|--------|
|pool not indexed|

| Portfolio Notification  |
|--------|
|portfolio not indexed|

## Checklist


jabby09:
@psparacino @Atmosfearful also need to consider this flow with paying with credit card to retire a pool project. In that case we end up with the stripe screens so will never see these modals.

Atmosfearful:
@jabby09 because it's in a separate repo, @psparacino will make a ticket and at some point soon we can get it working in the Offsetra/stripe receipt as well.

carbonmark/lib/prepareNavigateToCertificate.ts
```diff
@@ -0,0 +1,28 @@
...
+  retirementIndex: number,
+  blockNumber: number,
+  setSubgraphIndexed: Dispatch<SetStateAction<boolean | "timed out">>
+): Promise<boolean | "timed out"> => {
+  let currentBlock = await queryKlimaBlockNumber();
+  let counter = 0;
+
+  while (currentBlock < blockNumber && counter < 20) {
+    await new Promise((resolve) => setTimeout(resolve, 500));
+    currentBlock = await queryKlimaBlockNumber();
```
psparacino, line 17: throw error instead of "timed out"

carbonmark/lib/actions.retire.ts
```diff
@@ -136,7 +136,11 @@ export const retireCarbonTransaction = async (params: {
   onStatus: OnStatusHandler;
   projectAddress: string;
   isPoolDefault: boolean;
-}): Promise<{ transactionHash: string; retirementIndex: number }> => {
+}): Promise<{
+  transactionHash: string;
+  blockNumber: number;
```
psparacino, line 141: transaction blockNumber

or comments

carbonmark/components/pages/Project/Retire/Pool/RetireForm.tsx
```diff
@@ -258,6 +268,21 @@ export const RetireForm: FC<Props> = (props) => {
...
+    const retirementAddress =
+      inputValues?.beneficiaryAddress ?? (address || "");
+
+    if (retirementBlockNumber !== 0 && retirementIndex) {
+      prepareNavigateToCertificate(
+        router,
+        retirementAddress,
+        retirementIndex,
+        retirementBlockNumber,
+        setSubgraphIndexed
```
Atmosfearful, line 281: instead of a callback, just return one of the strings from the uniion `"pending" | "complete" | "timeout"`

```
const handleIndexingComplete = async () => {
  // const status = await prepareNavigateToCertificate();
  const status = await waitForIndexStatus({ blockNumber });
  prefetch(url);
  setIndexStatus(status);
}
```

carbonmark/components/pages/Portfolio/Retire/RetirementStatusModal/styles.tsx
```diff
@@ -1,12 +1,10 @@
 import { css } from "@emotion/css";
 
 export const title = css`
-  lineheight: 28px;
-  color: #3b3b3d;
   display: flex;
   align-items: center;
   justify-content: center;
-  gap: 5px;
+  gap: 6px;
```
Atmosfearful, line 7: rem plz

Atmosfearful (changes requested):

0xemc (approved):

Atmosfearful (approved):

PR #1339, opened 2023-07-18
Emc/feature/1221 api versioning

### LINKED ISSUE #1221: [carbonmark-api] API Versioning
What: Add versioning to existing endpoints
Why: Allow easy rollback 
How: https://www.fastify.io/docs/latest/Reference/Routes/#version-constraints

Expected Outcome:
[ ] - some_outcome
[ ] - some_outcome

Testing Criteria:
	1. do_something
	2. do_something_else
	3. expect_something

0xemc:
## Description

This PR provides an action that is triggered on push of a git tag of the format `carbonmark-api/*`

It then builds, deploys and aliases that tagged version of the carbonmark-api package.

CNAM record has been setup for `*.api.carbonmark.com` in google domains to accomodate the versioned urls

## Discussion

The reason for using tags for versioning is to avoid having to maintain multiple versions of code continuously (e.g `carbonmark-api/v1/src`,  `carbonmark-api/v2/src`). There are many issues with this approach (types, linting etc).

## Related Ticket

Resolves #1221


wowselim:
With this approach, how would we fix a bug that only affects v1 of the API?

Atmosfearful (approved):

Atmosfearful (approved):

PR #1340, opened 2023-07-18
Fix local api setup and README instructions

### LINKED ISSUE #1338: [API] Fix local setup and README instructions
* [x] The newest version of vercel CLI creates a `.vercel` folder with a `.env.development.local` file
 - dotenv invocations need to be updated to reference this instead of the root `.env.local`
* [x] Save this path to a constant and import it in both locations as `LOCAL_ENV_PATH`
* [x] Update the readme to explain what the expected behavior should be using vercel link && pull
- `vercel link` should be run from the KlimaDAO root
- must be a member of the KlimaDAO project in vercel
- say `Y`es when it asks if we want to link to an existing project, type `carbonmark-api`
- note that vercel pull --environment development should create the file in .vercel folder as noted above

psparacino:
## Description

Updated readme for updating to the newest vercel version, linking to the carbonmark-api vercel project, pulling the env variables, upgrading from previous version and running carbonmark-api locally in dev mode.

Updated path to the folder and variables created by the newest vercel cli.

## Related Ticket

Resolves #1338 

## Checklist

### DIFF
--- a/carbonmark-api/README.md
+++ b/carbonmark-api/README.md
@@ -1,24 +1,37 @@
 <!-- PROJECT LOGO -->
+ver
 <br />
-<div align="center">
-  <a href="https://github.com/github_username/repo_name">
-    <img src="./assets/logo.png" alt="Logo" width="80" height="80">
-  </a>
 
-<h2 align="center">Carbonmark Api</h2>
+<div  align="center">
 
-  <p align="center">
-    A BFF API for Carbonmark
-    <br />
-  </p>
+<a  href="https://github.com/github_username/repo_name">
+
+<img  src="./assets/logo.png"  alt="Logo"  width="80"  height="80">
+
+</a>
+
+<h2  align="center">Carbonmark Api</h2>
+
+<p  align="center">
+
+A BFF API for Carbonmark
+
+<br />
+
+</p>
+
 </div>
 
 ### Built With
 
 - [Fastify](https://fastify.io)
+
 - [Vercel](https://vercel.com)
+
 - [Firebase](https://firebase.google.com)
+
 - [Apollo](https://www.apollographql.com/)
+
 - [Ethers](https://docs.ethers.org/)
 
 <br />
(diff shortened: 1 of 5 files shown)


carbonmark-api/src/app.ts
```diff
@@ -2,9 +2,10 @@ import AutoLoad, { AutoloadPluginOptions } from "@fastify/autoload";
 import * as dotenv from "dotenv";
 import { FastifyPluginAsync } from "fastify";
 import { join } from "path";
+import { LOCAL_ENV_PATH } from "./utils/helpers/utils.constants";
 
 // This is not pretty but we need to reference the env file at the root of the monorepo
-const result = dotenv.config({ path: "../.env.local" });
+const result = dotenv.config({ path: LOCAL_ENV_PATH });
```
0xemc, line 8: Nice one and thanks for the extra docs 👍

Could you use `process.cwd` here plz ([example](https://github.com/KlimaDAO/klimadao/pull/1347/files#diff-bc149aa836d2ff3c5bd38465633d65ab9944fa49efa7687204654e8ee247baf5R7))  and remove the constant from utils?

```suggestion
const result = dotenv.config({ path: join(process.cwd(), ".env.local") });
```

psparacino:
@0xemc I saw your comment about changing to process.cwd in app.ts, but can't find it on here(?). Did you delete and is that change no longer needed?

carbonmark-api/README.md
```diff
@@ -1,24 +1,37 @@
 <!-- PROJECT LOGO -->
+ver
```
Atmosfearful, line 2:
```suggestion
```

Atmosfearful (approved):

0xemc:
> @0xemc I saw your comment about changing to process.cwd in app.ts, but can't find it on here(?). Did you delete and is that change no longer needed?

Hey no problem for now, there were other issues in the serverless version so I reverted it

PR #1354, opened 2023-07-20
Tufnel/Retirements enhancements # 1018

### LINKED ISSUE #1018: [stretch] Refactor & Enhancements for /retirements pages
First we should make some enhancements to the retire flow, which are highlighted here #1096

Finally, we can do this ticket, which describes some code cleanup to simplify /retirements page rendering.

# Considerations
- Currently, we have some complex fallback logic that calls the retirement aggregator storage contract via RPC, if the subgraph data is not available. The reason this is included, is because sometimes the subgraph can be slow to index.
- Instead of complicating this page rendering, I think we should remove this extra logic and simply treat subgraph as "source of truth". _No subgraph record -> show 404._

# Requirements for Single Retirement Page (`/retirements/[address]/[index]`)
* [x] Simplify data fetching so we only use subgraph. If the retirement is not indexed in the subgraph yet, 404. The client can retry if neccesary.
* [x] This data is static so we can set revalidation time to be long if all data is present - 86400 seconds (24 hours)
  * [x] However if the data is not present and we throw 404, then set the revalidation time to be very short - 0 seconds ?
  
# Requirements for Retirements List (`/retirements/[address]`)
Take a peek at get static props and see if there are some improvements we can make.
* [x] Revalidation time should be short, so new retirements appear quickly.

psparacino:
## Description

`/retirements`:

Removed RPC calls for totalRetirements and totalCarbonRetired and using the return from the subgraph instead to get the necessary data. 

Promise.all w/ rpc calls: 800ms to 1.2s
Subgraph only: ~200ms

shortened revalidation to 30 seconds if not found.

`/retirements/[index]`

Removed RPC fallback for subgraph. If no subgraph data exists, directs to a 404.

Changed revalidation times. If the subgraph data exists, revalidates every 24 hours. If no subgraph data, revalidates every second.

## Related Ticket

Resolves #1018 

## Checklist

### DIFF
--- a/carbonmark/pages/retirements/[beneficiary]/index.tsx
+++ b/carbonmark/pages/retirements/[beneficiary]/index.tsx
@@ -54,15 +54,19 @@ export const getStaticProps: GetStaticProps<Props, Params> = async (ctx) => {
     ]);
 
     if (Array.isArray(klimaRetires)) {
-      let total = 0;
       totalRetirements = klimaRetires.length;
 
-      klimaRetires.forEach((retirement) => {
-        total += parseFloat(retirement.amount);
-      });
-      totalCarbonRetired = total.toString();
+      try {
+        totalCarbonRetired = klimaRetires
+          .reduce((acc, retirement) => acc + parseFloat(retirement.amount), 0)
+          .toString();
+      } catch (e) {
+        throw new Error(
+          `Invalid retirement.amount in the array klimaRetires: ${e}`
+        );
+      }
     } else {
-      console.log("klimaRetires is not an array");
+      console.error("klimaRetires is not an array");
     }
 
     if (!translation) {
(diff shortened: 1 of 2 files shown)


carbonmark/pages/retirements/[beneficiary]/index.tsx
```diff
@@ -78,7 +90,7 @@ export const getStaticProps: GetStaticProps<Props, Params> = async (ctx) => {
     console.error("Failed to generate", e);
     return {
       notFound: true,
-      revalidate: 240,
+      revalidate: 30,
```
Comment (Atmosfearful) [inline carbonmark/pages/retirements/[beneficiary]/index.tsx:93]:
```suggestion
      revalidate: 1,
```

Atmosfearful (approved):

Issue #1262, opened 2023-06-30
[seller data] Remove activity data too?


ladytrekker:
## Description

All carbonmark listings needed to be returned.
Therefore, all former listings have been flagged as "deleted: true" on the backend.

The UI is filtering the listings data => give me all with "deleted: false", and renders the user stats accordingly.
Now, all user stats are set to "0" because all listings are deleted.

Still, the activity data exists.
This might lead to confusion as the user might ask, why are all my listings set to "0" but my former activity data (sold, created, bought from ...) still exists?
See an example profile here: https://www.carbonmark.com/users/jabby
And the API data [here](https://api.carbonmark.com/api/users/jabby?type=handle)

## Todo
- [x] discuss what the actual outcome should be (show former listings on the frontend even if deleted? Or remove all activity on the backend?)
- [x] Check what can be done on the frontend or backend accordingly

![Screenshot_2023-06-30_at_9 48 58_AM](https://github.com/KlimaDAO/klimadao/assets/95881624/8028789c-3ad4-43f8-95ec-aee979a1b433)

jabby09:
hi @ladytrekker 

discussed this with @Atmosfearful 

consensus is that in the Stats component, with seller listings removed, showing 0 `tonnes listed` and blank `active listing` makes sense.

the confusion for a user may lie in having zeroed out `tonnes sold` which should be a cumulative stat. Assuming we can access that data for the user/wallet we this data point should be a sum of all the tonnes sold by the seller.

Is that possible?

ladytrekker:
@jabby09 

To calculate the amount of tonnes a user sold, this is the math:
- from every listing get the sum of all values for `totalAmountToSell`
- from every listing get the sum of all values for `leftToSell`
- Return totalAmount - leftToSell => this is the amount a user sold so far
=> see code https://github.com/KlimaDAO/klimadao/blob/staging/carbonmark/lib/listingsGetter.ts#L26-L32

Because every listing of a user has been flagged as "deleted: true" 
it looks like the value for `leftToSell` has been set to "0" for every user as well.

See your own listings data here
- https://api.carbonmark.com/api/users/jabby?type=handle

=> every listing has `leftToSell: 0`
=> not possible to calculate the tonnes your user has still left from all listing and then get the value you've sold

## Use Activity data instead ?

It may be possible to get over every activity entry for a user to
- filter all activities with activityType: "Sold" and the seller ID is this user wallet address
- from these entries get the `amount` and the `price`
[...]

PR #1359, opened 2023-07-21
Get total tonnes sold from activities

ladytrekker:
## Description

See this comment here: https://github.com/KlimaDAO/klimadao/issues/1262#issuecomment-1625470684
Even though the listings are deleted, let's show the tonnes the user sold before when selling listings was still possible.

Problem is, that the listings value `leftToSell` was set to "0" for all listings of a user.
This value was used before to calculate the sum of all sold tonnes before.

Now, the sold amount must be retrieved by iterating over the user's activities data instead.
See comment here https://github.com/KlimaDAO/klimadao/issues/1262#issuecomment-1645503147

**See Preview:**
- [JABBY](https://carbonmark-git-lady-user-stats-klimadao.vercel.app/users/jabby)
- [ATMOS](https://carbonmark-git-lady-user-stats-klimadao.vercel.app/users/atmosfearful)
- [LADY](https://carbonmark-git-lady-user-stats-klimadao.vercel.app/users/ladsysdfsdfsdfsdfsdfsdfsfsdfsdf)

## 🔥 MERGING IS BLOCKED 🔥 

**The API limits the activity data of a user to 10 entries !** 🚒 

When summing up all SOLD activities of a user, the API needs to return ALL entries of this user.
Otherwise the total tonnes sold value is not correct.

## Related Ticket

Resolves https://github.com/KlimaDAO/klimadao/issues/1262

## Checklist

### DIFF
--- a/carbonmark/components/Stats/StatsListings.tsx
+++ b/carbonmark/components/Stats/StatsListings.tsx
@@ -4,7 +4,7 @@ import SavingsOutlinedIcon from "@mui/icons-material/SavingsOutlined";
 import SellOutlinedIcon from "@mui/icons-material/SellOutlined";
 import StoreOutlinedIcon from "@mui/icons-material/StoreOutlined";
 import { Text } from "components/Text";
-import { getTotalAmountSoldFromActivities } from "lib/activitiesGetter";
+import { getTotalTonnesSoldFromActivities } from "lib/activitiesGetter";
 import { getAmountLeftToSell, getTotalAmountSold } from "lib/listingsGetter";
 import { Listing, UserActivity } from "lib/types/carbonmark";
 import { useRouter } from "next/router";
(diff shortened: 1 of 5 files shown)


0xemc (commented):
@ladytrekker 

If you would like to remove the limit on activities you can remove the limit from the gql query in `carbonmark-api/src/graphql/marketplace.gql`

 ```gql
 activities(orderBy: timeStamp, orderDirection: desc, first: 10) {
 ```

 ```gql
 activities(orderBy: timeStamp, orderDirection: desc) {
 ```

Atmosfearful (commented):
Fetching all activity is not a viable long term solution, we should solve this with a direct subgraph query if possible. If not possible now, then in carbonmark v2.

jabby09:
@Atmosfearful should the branch be deleted

PR #1361, opened 2023-07-21
fix:1357 User wallet added to /GET response

0xemc:
## Description

Add missing `wallet` attribute from the GET `/user` requests.

Previously this was added in the middle of the route:

image

It seems to have been lost during refactor. Tests for response content will be added in forthcoming PR 

## Related Ticket

Resolves #13557


jabby09 (approved):
good to go

PR #1362, opened 2023-07-24
fix:1305 Update user serverside on edit profile

### LINKED ISSUE #1035: [listings] Better solution for fractional purchases and / token dust
We can either
- Enable purchasing of fractional amounts less than 1 tonne
- or, Hide fractional amounts and stick to our 1 tonne minimum (but don't hide them in the portfolio)
- or, Enforce 1 tonne intervals universally throughout the app to reduce the occurance of token dust? i.e. make it impossible to purchase 1.1 tonnes.

### LINKED ISSUE #1305: [profile] Error creating/editing profiles
In production, still getting an error message when editing a profile
image

On other PR's in progress, existing profiles are not being displayed properly - this is my jabby profile that works fine in production
image

0xemc:
## Description

Add a server mutation on EditProfile

## Related Ticket

Resolves #1305


carbonmark/lib/pollUntil.ts
```diff
@@ -1,3 +1,9 @@
+// This function polls a given function until it returns a valid result or the maximum number of attempts is reached.
```
Atmosfearful, line 1: fyi jsdoc comment syntax works better for vscode tooltips

```js
/**
 * I appear in tooltips
 */

File: carbonmark/components/pages/Users/SellerConnected/index.tsx [hunk 2]
```diff
@@ -124,8 +125,13 @@ export const SellerConnected: FC<Props> = (props) => {
       // Merge with data from Updated Profile as backend might be slow!
       const newUser = { ...userFromApi, ...profileData };
 
-      // Update the cache only, do not revalidate
+      // Update the cache only, this is an optimistic update
       await mutate(newUser, false);
+      //Update the server directly. This is a temporary solution until we can clean up the mutation logic
+      mutate(handleMutateUserUntil, {
```
Atmosfearful, line 131: if we have the newUser data already why do we need to poll? afaik that poller would only end when new activity is found but `onEditProfile` doesn't create an activity entry

Atmosfearful (changes requested):

carbonmark/components/pages/Users/SellerConnected/index.tsx
0xemc: If you remove this chunk, update your profile and then switch pages via the menu and go back to the profile page then the changes aren't persisted. It seems to pull the previous profile value from the cache.

To be fair the wait until logic confused me significantly so I just stole this from where the server mutation as being done elsewhere. Maybe you have a simpler solution?

Atmosfearful: @0xemc I think you just need to add populateCache:true to line 129, did that not work?

0xemc: I thought the same but then the UI is not updating.

There is something extremely wacky going on with the updating of users. Updating the profile submits a POST request which is immediately followed by 4 GET calls that override the update. I've tried using optimistic data as well but it seems to get overridden by the subsequent GET.

Screenshot 2023-07-26 at 9 38 59 am

I'm loathe to refactor how this is working and spend too much time on it so for now I thought I'd replicate how it was done elsewhere

jabby09 (changes requested):
@0xemc @Atmosfearful still getting an error when editing a profile:
Screenshot 2023-08-30 at 10 26 24 AM

jabby09 (approved):

0xemc:
Closing this as it's super stale.. User tests will hopefully expose any issues here

Issue #1144, opened 2023-05-17
[resources] Not working in some languages


ladytrekker:
## Description

When browsing the /resources page in a language other than english
=> no articles are shown.

**Compare:**
- https://www.klimadao.finance/resources  ✅ 
- https://www.klimadao.finance/de/resources 👎 
- https://www.klimadao.finance/fr/resources  👎 
- https://www.klimadao.finance/zh-CN/resources ✅ 
- https://www.klimadao.finance/es/resources 👎 
- https://www.klimadao.finance/hi/resources  👎

jabby09:
@ShimonD-KlimaDAO @Atmosfearful I'm not fully up-to-speed on translations, is this a known issue?

ShimonD-KlimaDAO:
@jabby09 - the handful of articles showing in zh-CN have been translated. The rest have not been translated, and other languages do not have any translated articles. I believe this was a one-off task initiated by Drew.

I think that every article missing a translation in any language should show in English. Otherwise we have an empty translated resources page. 
@Atmosfearful - could you please reassign this to the appropriate developer?

psparacino:
> > @jabby09 - the handful of articles showing in zh-CN have been translated. The rest have not been translated, and other languages do not have any translated articles. I believe this was a one-off task initiated by Drew.
> > I think that every article missing a translation in any language should show in English. Otherwise we have an empty translated resources page. @Atmosfearful - could you please reassign this to the appropriate developer?
> 
> @psparacino is this something you can pick up? Seems like it should be a quick change.
> 
> Please let me know if you need additional content or requirements.

Sounds good, I'll pick this one up

psparacino:
@jabby09 I have a question on this I can drop into the discord; is it possible to give me back the contributor role so I can get to the engineering channel? I lost it in the hack yesterday. 

The question is that currently I don't see a unique identifier between translated posts. Therefore I don't see a way to filter which posts have translations and then provide the rest in default English. I could do all or nothing, but that doesn't sound like the preferred out come here.

jabby09:
@psparacino will do on discord access

would need @Atmosfearful to address your question

ladytrekker:
@psparacino Concerning your question in discord:

## Extend CMS with more languages

- [ ] extend the post.js schema in /cms repository with all languages [here](https://github.com/KlimaDAO/klimadao/blob/staging/cms/schemas/post.js#L29-L40)
[...]

PR #1373, opened 2023-07-26
Tufnel/resources not showing 1144

psparacino:
## Description

If a language has no featuredArticles or documents translated yet, instead of showing a page with no featuredArticles or documents it will automatically redirect to the english page.

Used a client side re-route in order to keep using getStaticProps.

In Sanity, updated the languages to match production languages and made language selection required.

**NOTE: I tried to deploy the changes to Sanity but it doesn't look like I have push access even though I can now log in.**

## Related Ticket

Issue: #1144 

## Checklist

### DIFF
--- a/site/pages/resources/index.tsx
+++ b/site/pages/resources/index.tsx
@@ -6,10 +6,31 @@ import { GetStaticPropsContext } from "next";
 export async function getStaticProps(ctx: GetStaticPropsContext) {
   try {
     const { locale } = ctx;
-    const documents = await fetchCMSContent("allDocuments", { locale });
-    const featuredArticles = await fetchCMSContent("allFeaturedPosts", {
-      locale,
-    });
+
+    const [
+      enDocuments,
+      enFeaturedArticles,
+      localeDocuments,
+      localeFeaturedArticles,
+    ] = await Promise.all([
+      fetchCMSContent("allDocuments", { locale: "en" }),
+      fetchCMSContent("allFeaturedPosts", { locale: "en" }),
+      fetchCMSContent("allDocuments", { locale }),
+      fetchCMSContent("allFeaturedPosts", { locale }),
+    ]);
+
+    let documents, featuredArticles;
+    // if en has more documents or articles than locale, then show locale documents and articles plus en documents and articles
+    if (
+      localeDocuments.length < enDocuments.length ||
+      localeFeaturedArticles.length < enFeaturedArticles.length
+    ) {
+      documents = [...localeDocuments, ...enDocuments];
+      featuredArticles = [...localeFeaturedArticles, ...enFeaturedArticles];
+    } else {
+      documents = [...enDocuments];
+      featuredArticles = [...enFeaturedArticles];
+    }
 
     const translation = await loadTranslation(locale);
     if (!documents) {


site/lib/queries.ts
```diff
@@ -13,8 +13,7 @@ const hideFromProduction = IS_PRODUCTION
 
 const isDomainKlimadao = '(domain == "klimadao" || !defined(domain))';
 
-const localeFilter =
```
Atmosfearful, line 16: The original requirements, is peculiar and not properly documented, sorry for that.
* English and Chinese are the only languages that we support for blog posts at the moment.
*  If the user is viewing the site in a supported language, only fetch localized blog posts.
*  If non-supported language, fetch English blog posts.

However as I'm writing this, I believe we should simplify this requirement.

* [ ] If the user is viewing the site in a supported language, fetch English AND localized posts.
* [ ] If the user is viewing the site in a non-supported language, fetch English blog posts.

I feel like chinese users will miss out on a lot of content that they would otherwise be interested in (and can read with Google Translate)

Atmosfearful (changes requested):

0xtapi:
Tested this on various different browsers (most Chromium-based, plus Firefox) and on a few mobile browsers, no issues. Your redirect solution seems to be working.

However, from a UX perspective, I think this solution is not ideal. Getting redirected abruptly from your native language into English feels disruptive to me. There is no explanation as to why the user gets redirected, so it creates confusion. What I would do, as a user, is to try changing the language again and then be puzzled as to why it doesn't work.

A more elegant solution, imo, is to implement what Atmos proposed:

- If the user is viewing the site in a supported language, fetch English AND localized posts.
- If the user is viewing the site in a non-supported language, fetch English blog posts.

Additionally, we advise the translators to add a little bit of content to the translated pages. On the current ressources page, it says "Updates and thought leadership from the founders, DAO contributors, advisors and community."

We can tell translators to add a short explainer here, like "Note: These ressources are currently only available in English, with a few select pieces translated into [target language]. We are constantly working on expanding the number of available articles in [target language]."

Users would then retain their native language UI and also understand why the ressources section is behaving in this way.

site/pages/resources/index.tsx
```diff
@@ -6,10 +6,31 @@ import { GetStaticPropsContext } from "next";
...
+      fetchCMSContent("allFeaturedPosts", { locale: "en" }),
+      fetchCMSContent("allDocuments", { locale }),
+      fetchCMSContent("allFeaturedPosts", { locale }),
+    ]);
+
+    let documents, featuredArticles;
+    // if en has more documents or articles than locale, then show locale documents and articles plus en documents and articles
+    if (
+      localeDocuments.length < enDocuments.length ||
+      localeFeaturedArticles.length < enFeaturedArticles.length
```
Atmosfearful, line 26:
```suggestion
      localeDocuments.length < enDocuments.length ||
      localeFeaturedArticles.length < enFeaturedArticles.length
```

Won't this always be true? Unless the local is `en`, in which case we could have skipped the `locale` queries anyways.

site/pages/resources/index.tsx
```diff
@@ -6,10 +6,31 @@ import { GetStaticPropsContext } from "next";
...
+    ) {
+      documents = [...localeDocuments, ...enDocuments];
+      featuredArticles = [...localeFeaturedArticles, ...enFeaturedArticles];
```
Atmosfearful, line 29: Does this break the sorting? Because the localized articles won't always be first.

Unless we are doing more client side sorting that I'm forgetting about, I believe the GROQ query should sort for us.

We just need to update the GROQ query to match this logic:

```
fetch posts where:
  post.locale === `undefined`
  OR post.locale === 'en'
  OR post.locale === $LOCALE
```

Atmosfearful (changes requested):

site/pages/resources/index.tsx
psparacino: The idea was if there were an equal number of locale and en then there were no omissions and just to show the locale language. But sounds like this is moot point given the below feedback (using groq vs handling in getStaticProps)

psparacino:
groq query PR here: https://github.com/KlimaDAO/klimadao/pull/1411

This PR can be closed if the other is merged

Issue #1360, opened 2023-07-21
[tech] Firestore Performance Testing


psparacino:
@Atmosfearful 

Branch with tests on it here: https://github.com/KlimaDAO/klimadao/tree/tufnel/endpoint-performance-918

Tests in `carbonmark-api/src/speedTests`

I realized after creating the tests the the ticket was supposed to come first, sorry about that. Progress so far:

1. Created a separate collection in firestore called `usersTesting`
2. Created 1000 random users in the collection
3. Copied over the 19 existing users into usersTesting so in the usersTesting collection there are 1000 random users and 19 actual users
4. Compared retrieving those 19 users out of the 1019 using three methods:
- `await Promise.all(ids.map(id => db.collection('usersTesting').where('address', '==', id).get()));`
- `await db.getAll(...ids.map(id => db.collection('usersTesting').doc(id)));`
- `await Promise.all(chunks.map((chunk) => db.collection('usersTesting').where('address', 'in', chunk).get()) );`

8. Returning target users out of whole collection :

> whereTest duration: 697 ms
> getAllTest duration: 159 ms
> inTest duration: 156ms

9. If returning all users in a collection:

> whereTest duration: 1077 ms
> getAllTest duration: 975 ms
> inTest duration: 291 ms

where-in wins when returning from a large collection.

Current changes in `fetchMarketplaceListings`.

### End result in performance of `carbonmark-api/src/routes/projects/getById.ts`:

**Original (not using getAll, multiple firestore calls):**

Promise.all in `getByID`: 1.1 -1.4 s

`fetchMarketplaceAndListings`: 918 ms

**Using getAll for one firestore call to create a user map:**

Promise.all in `getById`: 574 ms-682ms

`fetchMarketplaceAndListings`: 167 ms-335ms

**Using where-in for one firestore call to create a user map):**

Promise.all in `getById`: 368ms-537ms

`fetchMarketplaceAndListings`: 143.586ms-280ms

Where-In is by far the fastest in the tests, but that performance jump is not fully reflected in the handler itself yet, so there might be other factors slowing it down

Atmosfearful:
Can you try the `in` operator from firestore:

```
const snapshot = await usersTestingCollection.where('address', 'in', ids).get();
```

Note the limitations in the docs: array can not exceed 30 entries. https://firebase.google.com/docs/firestore/query-data/queries#limitations_3

PR #1376, opened 2023-07-27
Tufnel/refactor fetchMarketplaceListings and firestore calls

### LINKED ISSUE #918: [tech] Very slow initial page loads
The critical pages are (in carbonmark)
- /retirements/[beneficiary]/[index]
- /projects/[project-key]
- /projects/[project-id]/purchase
- /projects/retire/pools/[pool]/
- /purchases/[txn-hash]

I suspect the subgraph is to blame. We need to do some performance profiling here to identify the problem.

Once #855 is merged we can use it to profile in production as well.

psparacino:
## Description

`fetchMarketplaceListings` previously had several, redundant firestore calls which was increasing the page's load time.

Refactored here to one firestore call, using `where-in` to fetch unique users and then using that for the rest of the handler.  Findings summarized here: #1360 

**NOTE: I modified the users collection in staging to have an address field.  In order for this to work for production the users collection will also need to be updated to include this field.**

## Related Ticket

Resolves #918 (resolves one element of the slow page load. Will be breaking out other findings into other tickets).

## Checklist


carbonmark-api/src/utils/helpers/fetchMarketplaceListings.ts
```diff
@@ -71,52 +73,74 @@ export const fetchMarketplaceListings = async ({
...
-          if (buyer.exists) {
-            assign(activityWithHandles, "buyer.handle", buyer.data()?.handle);
-          }
-        }
-        return activityWithHandles;
-      })
+          .where("address", "in", chunk)
+          .get()
+      )
     );
```
sprrwhwk, line 108: This is doing a network call for each item in the chunk array, correct? If so, I think we should reduce it to a single call given that the `in` operator seems to support array values.

See [here](https://firebase.google.com/docs/firestore/query-data/queries?hl=de#in_and_array-contains-any) for an example. What do you think?

psparacino: I believe it's making one network call per `chunk` array; firestore has a 30 value limit per chunk (if you include more values in the array you get this error from firestore:  `details: "'IN' supports up to 30 comparison values."`).  So a 75 value `chunks` array would be split up into 3 `chunk` arrays: 30, 30, 15 for a minimum of three calls. If I'm incorrect though it would be great to reduce the number of calls(!)

sprrwhwk:
Oh, that makes sense now 👍. I wasn't aware that each chunk is an array already.On 28 Jul 2023 18:54, p_sparacino ***@***.***> wrote:
@psparacino commented on this pull request.



In carbonmark-api/src/utils/helpers/fetchMarketplaceListings.ts:
> +    const userDocs = await Promise.all(
+      chunks.map((chunk) =>
+        fastify.firebase
           .firestore()
           .collection("users")
-          .doc(act.seller.id.toUpperCase())
-          .get();
-        if (seller.exists) {
-          activityWithHandles.seller.handle = seller.data()?.handle;
-        }
-        if (act.buyer) {
-          const buyer = await fastify.firebase
-            .firestore()
-            .collection("users")
-            .doc(act.buyer.id.toUpperCase())
-            .get();
-          if (buyer.exists) {
-            assign(activityWithHandles, "buyer.handle", buyer.data()?.handle);
-          }
-        }
-        return activityWithHandles;
-      })
+          .where("address", "in", chunk)
+          .get()
+      )
     );

I believe it's making one network call per chunk array; firestore has a 30 value limit per chunk (if you include more values in the array you get this error from firestore:  details: "'IN' supports up to 30 comparison values.").  So a 75 value chunks array would be split up into 3 chunk arrays: 30, 30, 15 for a minimum of three calls. If I'm incorrect though it would be great to reduce the number of calls(!)

carbonmark-api/src/utils/helpers/fetchMarketplaceListings.ts
```diff
@@ -71,52 +73,74 @@ export const fetchMarketplaceListings = async ({
...
+  const usersById = new Map<string, DocumentData | undefined>();
+
+  if (userIds.size !== 0) {
+    const chunkSize = 30;
+    const chunks : string[][] = [];
+    const ids = Array.from(userIds);
+
+    for (let i = 0; i < ids.length; i += chunkSize) {
+      chunks.push(ids.slice(i, i + chunkSize));
+    }
```
sprrwhwk, line 96: We could make use of lodash's `chunk` function here: https://lodash.com/docs/4.17.15#chunk

sprrwhwk (approved):

Atmosfearful (changes requested):
Very close, just few minor suggestions, nice work on this!

Atmosfearful (approved):
Good to go

PR #1391, opened 2023-08-01
fix:1382 Fiat Credit Limit Error

### LINKED ISSUE #1382: [credit card] 2000 limit error state UI issues
See [correct designs here](https://www.figma.com/file/5GewbdVaM1Q06CwHmF7d3F/%2Fprojects?type=design&node-id=890-29761&mode=design&t=xScP6hjsOt0wjb5H-4).

The error state when exceeding $2000 limit is not displaying properly or matching the designs.

### Non-default pool project...
for example https://www.carbonmark.com/projects/VCS-1718-2013/retire/pools/nct

- [x] Total cost in the Total Price panel should be red
- [x] RETIRE CARBON button should still be greyed out

Screenshot 2023-07-29 at 9 47 53 AM

### Edge case
EDIT: this is a subgraph bug tracked in https://github.com/KlimaDAO/klima-subgraph/issues/76
for example https://www.carbonmark.com/projects/VCS-981-2017/retire/pools/nct
NOTE: this project has two pool listings, not sure if that is the related

- [ ] Total cost in the Total Price panel in this case is red but the total is $0.00 and we are displaying a helper text message as well (not in the design)
- [ ] No helper text message until Pay with: (as per the design)
- [ ] RETIRE CARBON button should still be greyed out

Screenshot 2023-07-29 at 9 53 07 AM

0xemc:
## Description

An incorrect value was being used to evaluate if the total price of the retirement exceeded the total allowable fiat payment amount (by credit card).

Also:
- the totalPrice hidden was field was not being re-evaluated when quantity would change 
- the submit button was not listening to form errors

## Related Ticket

Resolves #1382 

## Changes

https://github.com/KlimaDAO/klimadao/assets/7104689/e9db8837-ff20-4b5f-969f-b85685ff76d1


jabby09 (changes requested):
@0xemc I'm seeing double error message here:

Screenshot 2023-08-01 at 10 48 12 AM

jabby09 (changes requested):
@0xemc as per the ticket, there is also an edge case I can't explain that is still an issue, [see this note](https://github.com/KlimaDAO/klimadao/issues/1382#issue-1827469973)

[Preview link](https://carbonmark-git-emc-fix1382-fiat-credit-limit-error-klimadao.vercel.app/projects/VCS-981-2017/retire/pools/nct)

I get $0 total cost for 1 or 2 tonnes

Screenshot 2023-08-01 at 10 52 03 AM

carbonmark/components/pages/Project/Retire/Pool/TotalValues.tsx
```diff
@@ -76,6 +79,14 @@ export const TotalValues: FC<TotalValuesProps> = (props) => {
     selectiveFee();
   }, []);
 
+  /** Surface form errors */
+  useEffect(() => {
+    const error = Object.values(formState.errors).at(0);
+    if (notNil(error)) {
+      setError(error.message ?? "");
+    }
+  }, [formState.errors]);
```
Atmosfearful, line 88: I know this is not your doing but, why do we even need to set the error.message to state, can't we just render the message right in jsx? useFormContext should trigger the necessary component re-render.

carbonmark/components/pages/Project/Retire/Pool/RetireForm.tsx
```diff
@@ -93,27 +94,26 @@ export const RetireForm: FC<Props> = (props) => {
   useEffect(() => {
     const getFiatMaxBalance = async () => {
       try {
-        const fiatInfo = await getFiatInfo();
+        const { MAX_USDC } = await getFiatInfo();
```
Atmosfearful, line 97: Let's just remove this async logic and keep use a constant `MAX_USDC` value of `2000`. Offsetra agreed to keep this much liquidity at all times.

The staging wallet only has a few bux in it, but that's OK because when the user does a $2000 checkout in Staging via Stripe Test Mode, it will default to `0.002 tonnes` under the hood

Atmosfearful:
> @0xemc as per the ticket, there is also an edge case I can't explain that is still an issue, [see this note](https://github.com/KlimaDAO/klimadao/issues/1382#issue-1827469973)
> 
> [Preview link](https://carbonmark-git-emc-fix1382-fiat-credit-limit-error-klimadao.vercel.app/projects/VCS-981-2017/retire/pools/nct)
> 
> I get $0 total cost for 1 or 2 tonnes
> 
> Screenshot 2023-08-01 at 10 52 03 AM

Out of scope for this PR, I narrowed this down to a separate issue https://github.com/KlimaDAO/klima-subgraph/issues/76

carbonmark/components/pages/Project/Retire/Pool/RetireInputs.tsx
```diff
@@ -332,7 +328,11 @@ export const RetireInputs: FC<Props> = (props) => {
               message: t`Could not calculate Total Cost`,
             },
             validate: {
```
Atmosfearful, line 330: ok this is wild, it's an invisible InputField used to add extra validations?

image

If it's not too much trouble I would tear this out, and move the validations to the quantity input (`validate` object or callback) https://www.react-hook-form.com/api/useform/register/
If that's not feasible then we could use a useEffect which `watch`es the formstate and invokes `setError`

carbonmark/components/pages/Project/Retire/Pool/RetireInputs.tsx
```diff
@@ -332,7 +328,11 @@ export const RetireInputs: FC<Props> = (props) => {
...
-              moreThanZero: (v) => Number(v) > 0 || t`Total Cost is required`,
+              //Add form level validation for totalPrice exceeding the fiat max
+              exceedFiatMax: (v) =>
+                !(
+                  paymentMethod === "fiat" && Number(props.fiatMax) < Number(v)
+                ) || getValidations().totalPrice.max.message,
```
Atmosfearful, line 335: I think this is what's causing the double error that jabby noted in his screenshot.

Atmosfearful (changes requested):
Just resolve that double error message however you can,
and set that MAX_USDC to a constant. The rest are just suggestions at your discretion.

jabby09 (approved):
preview link on the PR still shows double message but appears fixed in staging so assume this is good to go.

jabby09:
@Atmosfearful this appears to be resolved

PR #1399, opened 2023-08-03
Duplicate projects on /projects page

### LINKED ISSUE #1161: [carbonmark-api] Duplicate projects in /projects list
2015 of this project appears twice https://www.carbonmark.com/projects/VCS-875-2015
(search the name to reproduce)
Might be related to #1160

psparacino:
## Description

Duplicate projects showed on the `/projects` page. The cause seemed to be multiple tokens and prices; one project would show per token/price.

In this solution, the fetched projects are filtered to find multiple instances of the same project-key+vintage, and then select only the lowest priced option to display on the `/projects` page. 

When the user navigates to the individual project page they'll still see both the lowest priced option as well as any others.

## Related Ticket

Resolves #1161

## Checklist

### DIFF
--- a/carbonmark/hooks/useFetchProjects.ts
+++ b/carbonmark/hooks/useFetchProjects.ts
@@ -18,28 +18,31 @@ export const useFetchProjects = () => {
   const allProjects = data?.filter(negate(isNil)) ?? [];
 
   // group projects by key and vintage to find same projects with multiple prices
-    const groupedProjects = allProjects.reduce<{ [key: string]: Project[] }>((grouped, project) => {
+  const groupedProjects = allProjects.reduce<{ [key: string]: Project[] }>(
+    (grouped, project) => {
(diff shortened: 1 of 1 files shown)


0xemc (changes requested):
Great pickup 👍 

This uniqueness test must have got lost during the refactor. 

I've re-added it in this PR:

https://github.com/KlimaDAO/klimadao/pull/1400

Up to you @psparacino @Atmosfearful if you want to keep this logic in the FE code too.. 

Though I'd suggest extracting it to be a generic "remove duplicates from array" utility so we don't bloat these files

psparacino:
@0xemc Given your PR addresses this on the api level we can probably close this PR as it seems redundant to also include on the FE

0xemc:
> @0xemc Given your PR addresses this on the api level we can probably close this PR as it seems redundant to also include on the FE

Ok, fair enough. Closed

PR #1406, opened 2023-08-07
Support search by project id- 1029

### LINKED ISSUE #1029: [projects] Searching for a project id does not work as expected
Unsure if this is an explicit requirement but intuitively searching for a project's registry id (e.g VCS-1052) should return that project in the search results.

https://registry.verra.org/app/projectDetail/VCS/1052

https://www.carbonmark.com/projects?search=VCS-1052

https://www.carbonmark.com/projects/VCS-1052-2012

image

**US**
As a user when I view a listing on an existing registry website, I want to easily find that listing by it's ID in the Carbonmark Projects page by searching for that ID.

psparacino:
## Description

Searching by project keys was previously unsupported as search input queries only supported name searches. This PR adds project keys to possible search inputs. This query format should also be more extensible as different search input requirements can be added to the `or` block.

## Related Ticket

Resolves #1029 

## Checklist


0xemc (approved):
Nice one 👍 works like a charm

PR #1440, opened 2023-08-16
Add separate prettier checks

### LINKED ISSUE #1438: [CI/CD] Move Prettier formatting checks to a GitHub action
In the klimadao monorepo I would like to set up a GHA that runs `prettier --check` on every commit and throws a red status if an issue is surfaced.

# background
To reduce build times and simplify our dependency tree I removed prettier checks from all packages and added a `prettier-check-all` command to the top level of npm workspaces`. I'd like to bring it back in the form of a parallel GHA instead of running it before each build.

# Acceptance criteria:
- [ ] Every commit to every branch in the `klimadao` monorepo should trigger a `check formatting` status check
- [ ] It should run in parallel with the other actions & bots (vercel)
- [ ] the status check should throw a red X when `prettier check` catches something

## Stretch goals
Bonus points if you can only check directories with changes i.e. with `git diff`
  1. get list of top level dirs containing changes with `git diff --name-only [TARGET_BRANCH] | awk -F'/' '{print $1}' | sort -u`
    - for `staging`, check against `main`
    - for branches that are not `staging` or `main`, i.e. any working branch, check against `staging`
  2. pass this list of dirs as an argument to `prettier --check` so that only changed files are checked.

0xAurelius:
## Description

Runs prettier checks as separate GitHub Action workflow

Only checks files that have been modified (relative to staging for non-staging branches, relative to main for staging)

## Related Ticket

Resolves #1438 

## Checklist

### DIFF
--- a/.github/workflows/check_formatting.yml
+++ b/.github/workflows/check_formatting.yml
@@ -1,0 +1,54 @@
+name: Check formatting
+on:
+  push:
+    branches: "*"
+  pull_request:
+    branches: "*"
+
+jobs:
+  prettier:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Get Prettier version
+        id: prettier-version
+        run: >
+          echo "version=$(cat package.json | grep '\"prettier\": \"' |
+          cut -d ':' -f 2 | tr -d '"' | tr -d ' ' | tr -d ,)" >> $GITHUB_OUTPUT
+
+      - name: Get Prettier organize imports version
+        id: prettier-org-import-version
+        run: >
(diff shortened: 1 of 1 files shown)


.github/workflows/prettier.yml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  prettier:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Install Prettier
+        run: npm install -g prettier
```
Atmosfearful, line 17: ah we also need to match the package.json.prettier config as I've opted out of a v3 prettier convention

Atmosfearful (approved):

PR #1447, opened 2023-08-17
Makka / #1230 add one percent to total cost

### LINKED ISSUE #1230: [form] Validation check on wallet balance
[Discord thread](https://discord.com/channels/841390338324824096/1112500261894754324) for context.

When redeeming (buy) or retiring from a pool, when the user completes the form we approve more than the quoted cost (1%) to account for slippage or pool price movement. The rationale being, if we pass the exact cost to maxAmountIn, then even 0.000001 usdc of price movement would cause a txn to fail.

Currently the form validation only compares cost to balance, not cost plus the slippage buffer. This can create an edge case issue if the user is trying to maximize tonnes to use all of the balance of the pay with token in their wallet; in that the form will pass validation but will fail in the confirm modal because of the added 1%.

The ticket adds a more robust validation at the form stage to account for the 1% slippage buffer.

## Requirements

- [ ] In Carbonmark, on Retire and Buy forms for Pool listings, need to add additional form validation to ensure the Pay with balance is the cost + 1%. If not, need to display the form error message. Design [here](https://www.figma.com/file/5GewbdVaM1Q06CwHmF7d3F/%2Fprojects?type=design&node-id=785-28351&t=AGyn7q0nJKgFKs64-4).
- [ ] In the Retirement Aggregator, on form page https://app.klimadao.finance/#/redeem and https://app.klimadao.finance/#/offset, need to add additional form validation to ensure the Pay with balance is the cost + 1%. If not, need to display the form error message. Design [here](https://www.figma.com/file/HXwNOLfM9CQHgwmouz0rpk/CARBON-OFFSETTING?type=design&node-id=2305-12141&t=eUJtBhZaKiaIPuja-4).
- [ ] NOTE...this validation step is not required on Retire and Buy forms for Seller listing in Carbonmark because their is no slippage.

0xMakka:
## Description

Adds a 1% fee to total cost on retire/purchase pools.

- [x] Add error message and styling when insufficient balance.

## Related Ticket

Resolves #1230

## Checklist


jabby09:
@0xMakka is this one ready for QA?

0xMakka:
> @0xMakka is this one ready for QA?

@jabby09 I think you should be good to start looking at it. I pushed some changes earlier on the /redeem page on the app, however I was having issues with Polygon RPC so couldn't fully verify on the /offset page.

0xtapi:
Helloes,

+1% total cost on Carbonmark works, no issues. Checked it on both Purchase and Retire. ✅

+1% total cost also works on KlimaDAO /redeem (✅), but not on KlimaDAO /offset (❌).

0xMakka:
> Helloes,
> 
> +1% total cost on Carbonmark works, no issues. Checked it on both Purchase and Retire. ✅
> 
> +1% total cost also works on KlimaDAO /redeem (✅), but not on KlimaDAO /offset (❌).

@0xtapi thanks for this, I was having issues with polygon-rpc at the time when working on the offset part so the balance wasn't loading for me to verify it was working as intended. Will pick this up again today and provide a fix for the /offset 👍

0xtapi:
> @0xtapi thanks for this, I was having issues with polygon-rpc at the time when working on the offset part so the balance wasn't loading for me to verify it was working as intended. Will pick this up again today and provide a fix for the /offset 👍

I believe it, lots of RPCs have been giving me trouble in the past few days. Better luck next time 🙃

carbonmark/lib/getPoolData.ts
```diff
@@ -9,3 +12,29 @@ export const getPoolTokenType = (pool: Uppercase<PoolToken>): CarbonToken =>
 
 export const getDefaultPoolFromPrices = (prices: Price[]) =>
   prices.find((p) => p.isPoolDefault);
+
+export const getPoolApprovalValue = (
+  cost: string,
+  paymentMethod: CarbonmarkPaymentMethod,
+  maxDecimals?: number
```
Atmosfearful, line 19: we should use `getTokenDecimals` and derive it from the payment method, not a configurable param

carbonmark/components/pages/Project/Purchase/Pool/PurchaseModal.tsx
```diff
@@ -87,6 +91,8 @@ export const PurchaseModal: FC<Props> = (props) => {
     (showSuccessScreen && t`Purchase successful`) ||
     t`Confirm Purchase`;
 
+  console.log("props.approvalValue", props.approvalValue);
```
Atmosfearful, line 94: plz remove

carbonmark/components/pages/Project/Purchase/Pool/PurchaseModal.tsx
```diff
@@ -16,6 +16,10 @@ export interface Props {
     value: string;
     token: CarbonmarkToken;
   };
+  approvalValue?: {
```
Atmosfearful, line 19: this purchase modal will always be USDC, so I believe `token` can be removed and approvalValue can just be a value

Atmosfearful (changes requested):
I haven't QAd it hands on yet, but I don't think this satisfies all the ticket requirements.
Unfortunately link to the designs is broken in the ticket, waiting on @jabby09 to help me track that down, if memory serves me correctly we should have a *different* error message that tells the user they need to have enough to cover cost +1 for slippage.

In the meantime here are some change requests-- (don't mess with error messages until we can peek at the designs)

jabby09:
> if memory serves me correctly we should have a _different_ error message that tells the user they need to have enough to cover cost +1 for slippage.

@0xMakka @Atmosfearful yes that is correct about an error message; see the requirements in #1230 
As an example, you can see the design implementation for the RA https://www.figma.com/file/HXwNOLfM9CQHgwmouz0rpk/CARBON-OFFSETTING?type=design&node-id=2305-12141&t=eUJtBhZaKiaIPuja-4

@0xtapi when you did the QA for this, did you confirm the error message as per this requirement?

> In Carbonmark, on Retire and Buy forms for Pool listings, need to add additional form validation to ensure the Pay with balance is the cost + 1%. If not, need to display the form error message.

carbonmark/components/pages/Project/Purchase/Pool/PurchaseModal.tsx
0xMakka: Done

carbonmark/lib/getPoolData.ts
0xMakka: My bad on this one, removed `maxDecimals` and now using `getTokenDecimals` from the payment method.

0xtapi:
1% increased cost on /offset works now ✅

@jabby09 

> when you did the QA for this, did you confirm the error message as per this requirement?

No, I did not. Makka said the only change that was made was the added +1% cost, so that's the only thing I took a look at. You are right that it currently does not show the error message as seen in the designs. I'll make it more clear next time if I actually QA'd for everything in the related ticket or just for whatever fix is mentioned in the OP.

Which means that, for now, the balance error messages are missing ❌.

0xMakka:
> 1% increased cost on /offset works now ✅
> 
> @jabby09
> 
> > when you did the QA for this, did you confirm the error message as per this requirement?
> 
> No, I did not. Makka said the only change that was made was the added +1% cost, so that's the only thing I took a look at. You are right that it currently does not show the error message as seen in the designs. I'll make it more clear next time if I actually QA'd for everything in the related ticket or just for whatever fix is mentioned in the OP.
> 
> Which means that, for now, the balance error messages are missing ❌.

This was my bad, I missed this error message on the designs. Apologies guys, will have a fix in for this today.

jabby09:
@0xtapi this is ready now for re-testing and requirements verification

0xtapi:
I think we're still not quite there, but we're getting closer. Give me a quick sanity check here please: The "1% warning message" should display whenever your balance isn't at least the total cost plus 1%, right?

If so, this still needs some work. Check screenshots below:
nowarn
warn

First one is my balance exceeding the total cost by about 0.04%, second screenshot is my balance not exceeding the total cost (where it correctly spits out the message).

The warning should trigger on the first screenshot as well because my balance is not higher than total cost + 1% (my balance would need to be about 703 USDC in this example), right? So if the total cost is 100 USDC, the warning should trigger for all balance values below 101 USDC, correct?

If these assumptions are true, this still needs some more tweaking 🔧

0xMakka:
> I think we're still not quite there, but we're getting closer. Give me a quick sanity check here please: The "1% warning message" should display whenever your balance isn't at least the total cost plus 1%, right?
> 
> If so, this still needs some work. Check screenshots below: nowarn warn
> 
> First one is my balance exceeding the total cost by about 0.04%, second screenshot is my balance not exceeding the total cost (where it correctly spits out the message).
> 
> The warning should trigger on the first screenshot as well because my balance is not higher than total cost + 1% (my balance would need to be about 703 USDC in this example), right? So if the total cost is 100 USDC, the warning should trigger for all balance values below 101 USDC, correct?
> 
> If these assumptions are true, this still needs some more tweaking 🔧

@0xtapi just a note the "cost" price includes the 1% fee. Here is two scenarios I ran through when testing.

**Offset View:**

Scenario 1:

- Tonnes = 6.19081
- Cost without 1% fee = 3.071108 USDC
- Cost with 1% fee -> 3.071108 + 1% = 3.101819

My exact USDC balance is 3.101819

In this case, no error message is shown as there is enough balance to cover the fee and the cost

Screenshot 2023-09-13 at 11 14 34

-------

Scenario 2:

- Tonnes = 6.190811
- Cost without 1% fee = 3.071109 USDC
- Cost with 1% fee -> 3.071109 + 1% = 3.101820

My exact USDC balance is 3.101819

In this case, an error message is shown as there isn't enough balance to cover the fee and the cost.

Screenshot 2023-09-13 at 11 14 48

-------

**Redeem View:**

8.21585

Scenario 1:

- Tonnes = 8.21585
- Cost without 1% fee = 3.071108 USDC
- Cost with 1% fee -> 3.071108 + 1% = 3.101819

My exact USDC balance is 3.101819

In this case, no error message is shown as there is enough balance to cover the fee and the cost

Screenshot 2023-09-13 at 11 29 23

-----------

Scenario 2: 

- Tonnes = 8.215852
- Cost without 1% fee = 3.071109 USDC
- Cost with 1% fee -> 3.071109 + 1% = 3.101820

My exact USDC balance is 3.101819

In this case, an error message is shown as there isn't enough balance to cover the fee and the cost.

Screenshot 2023-09-13 at 11 30 58

------

Not sure if I'm missing something else here, but will be sure to jump back on this and address if it is still an issue?

jabby09:
@0xMakka @0xtapi I'm having trouble following the RA screenshots above and for some reason I cannot currently get my metamask wallet to connect properly to the RA to test myself. However, in Carbonmark...

Apologies if the requirements were not clear in https://github.com/KlimaDAO/klimadao/issues/1230

Note, the 1% to account for slippage or pool price movement is automatically added when the transaction request is sent to the contract. This step happens here in the Confirm Retirement modal when the user clicks APPROVE.

This will fail every time if the user wallet does not have enough USDC to cover the total cost + 1%.

Screenshot 2023-09-13 at 9 46 42 AM

The objective is this ticket is to validate on the form before the user gets to the Confirm modal.

In the example below:

- My USDC balance is $5.50
- Therefore, I should not be able to submit this form for any total cost + 1% that exceeds $5.50
- Basically that works out to ~$5.45 (i.e. $5.45 x 1.01 = $5.50)
- In the example, somewhere close to 14.44 tonnes should trigger the validation error but it does not

Screenshot 2023-09-13 at 9 55 37 AM

The same logic applies to the retirement aggregator form.

0xMakka:
> @0xMakka @0xtapi I'm having trouble following the RA screenshots above and for some reason I cannot currently get my metamask wallet to connect properly to the RA to test myself. However, in Carbonmark...
> 
> Apologies if the requirements were not clear in #1230
> 
> Note, the 1% to account for slippage or pool price movement is automatically added when the transaction request is sent to the contract. This step happens here in the Confirm Retirement modal when the user clicks APPROVE.
> 
> This will fail every time if the user wallet does not have enough USDC to cover the total cost + 1%.
> 
> Screenshot 2023-09-13 at 9 46 42 AM
> The objective of this ticket is to validate on the form before the user gets to the Confirm modal.
> 
> In the example below:
> 
> * My USDC balance is $5.50
> * Therefore, I should not be able to submit this form for any total cost + 1% that exceeds $5.50
> * Basically that works out to ~$5.45 (i.e. $5.45 x 1.01 = $5.50)
> * In the example, somewhere close to 14.44 tonnes should trigger the validation error but it does not
> 
> Screenshot 2023-09-13 at 9 55 37 AM
> The same logic applies to the retirement aggregator form.

@jabby09 thanks for the above message. So, I think the part I might be confusing is that currently the total cost on the card includes the 1% fee. From what I am now gathering; the total cost should be displayed without the 1% fee added, but when adding the error handling message, if the user doesn't have the total cost + 1% only at that point should the error message be displayed?

If this is the case (my bad at reading the requirements wrong), I'll rework it again; when looking at the RA app, I seen there is a tooltip on the total cost which says the total cost includes slippage and aggregation fee of 1% so I assumed it should be added to the total cost by default.

jabby09:
> thanks for the above message. So, I think the part I might be confusing is that currently the total cost on the card includes the 1% fee. From what I am now gathering; the total cost should be displayed without the 1% fee added, but when adding the error handling message, if the user doesn't have the total cost + 1% only at that point should the error message be displayed?
> 
> If this is the case (my bad at reading the requirements wrong), I'll rework it again; when looking at the RA app, I seen there is a tooltip on the total cost which says the total cost includes slippage and aggregation fee of 1% so I assumed it should be added to the total cost by default.

hey @0xMakka okay now I'm confused, hahaha

don't do anything at this point, I think we need to talk to @Atmosfearful about this one

jabby09 (changes requested):
@0xMakka 

the logic now seems to be working as expected:
Screenshot 2023-09-29 at 10 30 09 AM

However, the error message should read: `Your balance much equal at least 1% more than the cost of the transaction.`

Similar to the message on the retirement aggregator:
Screenshot 2023-09-29 at 10 31 24 AM

FYI...I can't seem to get the RA to load properly, just get a loading balance spinner forever so I can't really test the RA implementation of this.

0xMakka:
> #1230

@jabby09 The error message is to be shown under the pay with dropdown, couldn't see it in the designs linked. I thought I had it added but will take another look. 

With regards the loading issues, this is the rpc issue I've been looking into as polygon-rpc is constantly failing for me which prevents the balance from being able to be loaded.

jabby09:
> The error message is to be shown under the pay with dropdown, couldn't see it in the designs linked. I thought I had it added but will take another look.

ah, okay. it was not showing up for me on Carbonmark

jabby09 (changes requested):
@0xMakka please check the style on the pay with element and font size for the error message.

Used to seeing this...
Screenshot 2023-09-30 at 9 59 47 AM

With USDC seeing this...
Screenshot 2023-09-30 at 10 00 49 AM

0xMakka:
> @0xMakka please check the style on the pay with element and font size for the error message.
> 
> Used to seeing this... Screenshot 2023-09-30 at 9 59 47 AM
> 
> With USDC seeing this... Screenshot 2023-09-30 at 10 00 49 AM
> 
> The other thing to check / keep in mind is the changes in #1459 and how the error messages are displayed.

@jabby09 My bad on this, I copied the font size from figma, but all the error messages on carbonmark are a slightly smaller size than the figma file (16px in figma, 14px on carbonmark) - so I've just updated to 14px to be consistent with what's currently on carbonmark. I also noticed I missed the red border on the dropdown so pushed a fix for this also.

jabby09 (approved):

0xemc (approved):
This PR looks like it's passed @jabby09 QA but has been up for a couple months now, can you resolve the conflicts and then we can close it @0xMakka

0xMakka:
> This PR looks like it's passed @jabby09 QA but has been up for a couple months now, can you resolve the conflicts and then we can close it @0xMakka

Good shout @0xemc, yeah other higher priority tasks came up so this one was just left sitting. Just fixed all conflicts there now 👍

Atmosfearful (approved):

PR #1460, opened 2023-08-21
Makka/1423 change asset details card to accordion

### LINKED ISSUE #1423: [forms] Make Asset Details component closed by default on page load
On both /purchase and /retire forms for Pool Listings, the current Asset Details component is open on form page load.

Screenshot 2023-08-14 at 4 24 41 PM

The purpose of this workstream is to change this so the Asset Details component is closed by default on page load.

[Design here](https://www.figma.com/file/5GewbdVaM1Q06CwHmF7d3F/Projects?type=design&node-id=1261-15460&mode=design&t=8dyo3qp0Bv3QcM7Z-4).

Screenshot 2023-08-14 at 4 26 08 PM

0xMakka:
## Description

Change asset details card to an accordion on both /purchase and /retire forms for Pool Listings.

## Related Ticket

Resolves #1423

## Changes

| Before  | After (Collapsed)  | After (Expanded)  |
|---------|--------|--------|
|Screenshot 2023-08-21 at 09 58 24|Screenshot 2023-08-21 at 09 57 37|Screenshot 2023-08-21 at 09 57 46|

## Checklist


Atmosfearful (changes requested):
Edge case bug with an easy fix.

The accordion causes open tooltips to move out-of-place.
To reproduce:
1. Tab-navigate to reach the accordion, you will pass over a tooltip, which will trigger the tooltip to open.
2. The tooltip stays open as is tippy default behavior when tab-navigating.
3. Pressing spacebar on the accordion to expand it, will cause the parent sticky container to expand, which causes the tooltip to shift.

Solution:
* [ ] create a new stacking context for tippy on the parent card container by applying `position: relative` (see screenshot)
image
* [ ] leave a comment in the CSS explaining why this gets a position relative.
```
position: relative creates a new stacking context for tooltips to attach to, otherwise the next closest stacking context would be the sticky column which can change sizes and cause the tooltip to shift.
```

0xMakka:
> Edge case bug with an easy fix.
> 
> The accordion causes open tooltips to move out-of-place. To reproduce:
> 
> 1. Tab-navigate to reach the accordion, you will pass over a tooltip, which will trigger the tooltip to open.
> 2. The tooltip stays open as is tippy default behavior when tab-navigating.
> 3. Pressing spacebar on the accordion to expand it, will cause the parent sticky container to expand, which causes the tooltip to shift.
> 
> Solution:
> 
> * [x]  create a new stacking context for tippy on the parent card container by applying `position: relative` (see screenshot)
>   image
> * [x]  leave a comment in the CSS explaining why this gets a position relative.
> 
> ```
> position: relative creates a new stacking context for tooltips to attach to, otherwise the next closest stacking context would be the sticky column which can change sizes and cause the tooltip to shift.
> ```

Great catch and thank you for also providing the solution.

Atmosfearful (approved):

PR #1505, opened 2023-09-04
cleanup carbonmark-data css

### LINKED ISSUE #1462: [carbon-dashboard] Cleanup carbonmark-data css
* There's some places where we use `px` units. We should set the root font size to `10px` and use `rem` as a unit throughout the project.
* The desktop sidebar uses a fixed position but we can probably convert it to something that's more predictable and easier to work with.

sprrwhwk:
## Related Ticket

Resolves #1462 

## Checklist

### DIFF
--- a/carbonmark-data/components/Layout/style.module.scss
+++ b/carbonmark-data/components/Layout/style.module.scss
@@ -5,55 +5,55 @@
 
   @include desktop {
     position: fixed;
-    width: 180px;
+    width: 18rem;
     overflow: auto;
-    padding: 48px 20px 48px 20px;
-    gap: 40px;
+    padding: 4.8rem 2rem 4.8rem 2rem;
+    gap: 4rem;
     display: grid;
     text-align: center;
     background-color: var(--surface-04);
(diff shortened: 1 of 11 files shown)


carbonmark-data/components/Layout/style.module.scss
```diff
@@ -116,7 +116,8 @@
 
 .mobileBottomNavItem {
   display: flex;
-  height: 44px;
+  padding: 0.4rem;
+  height: 3.6rem;
```
biwano, line 120: Why is this not 4.4rem?

biwano (approved):

carbonmark-data/components/Layout/style.module.scss
sprrwhwk: Unless I'm missing something:
`3.6 + 0.4 + 0.4 = 4.4`

sprrwhwk: This is from figma

grafik

PR #1507, opened 2023-09-04
Project Info Subgraph Fallback

### LINKED ISSUE #974: [portfolio] Use CMS endpoint for reading unlisted project data
## Description

In order to render the portfolio and the create-listing dropdown for every token asset a user owns,
the frontend checks for each token if:
- is this token already listed as a project on Carbonmark? => get the project data from Carbonmark API
- If not => is this a C3 token? => create a contract for this C3 token and call contract methods to get the project data
- is this a TCo2 token? => create a contract for this TCo2 token and call contract methods to get the project data

See the [code here](https://github.com/Atmosfearful/bezos-frontend/blob/bezos-main/carbonmark/lib/actions.ts#L251-L268)

=> The creation of token contracts can be replaced with a new API endpoint which reads data from the CMS!

## TODO
- [x] wait until the new endpoint for the CMS token data is live
- [x] replace the contract methods with Carbonmark API endpoint requests to get the project data for each token asset a user own

psparacino:
## Description

Addressed #974 . The project info is available on the `polygon-bridged-carbon` subgraph, so the subgraph is used in place of the cms endpoint mentioned in the issue.

I kept the contract calls as a last fallback. So for project info it will currently check the api, if api fails -> subgraph, if subgraph fails -> rpc call.

## Related Ticket

Resolves #974 

## Checklist

### DIFF
--- a/carbonmark/lib/actions.ts
+++ b/carbonmark/lib/actions.ts
@@ -1,7 +1,7 @@
 import C3ProjectToken from "@klimadao/lib/abi/C3ProjectToken.json";
 import IERC20 from "@klimadao/lib/abi/IERC20.json";
 import TCO2 from "@klimadao/lib/abi/TCO2.json";
-import { addresses } from "@klimadao/lib/constants";
+import { addresses, subgraphs } from "@klimadao/lib/constants";
 import { AllowancesToken } from "@klimadao/lib/types/allowances";
 import { formatUnits } from "@klimadao/lib/utils";
 import { Contract, ethers, providers, Transaction, utils } from "ethers";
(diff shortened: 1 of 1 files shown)


carbonmark/lib/actions.ts
```diff
@@ -277,8 +277,14 @@ export const addProjectsToAssets = async (params: {
         // project exists on API
         if (projectFromApi && projectFromApi.key) {
           project = projectFromApi;
+          // if no project on API, try subgraph
+        } else if (projectId) {
```
Atmosfearful, line 281: The API should handle all the fallbacks ideally, and should merge in CMS data if present.

As ladytrekker said in #974 
> The creation of token contracts can be replaced with a new API endpoint which reads data from the CMS!

In other words she means we should rely on the /users endpoint for all assets and listings, and it should handle merging data and fallbacks.

Apologies for that the ticket requirements were not clear. We should deprioritize this until we have CMV2 working and then reassess.

Atmosfearful (changes requested):

carbonmark/lib/actions.ts
psparacino: sounds good. I'll back burner this until after CMv2 is deployed and then will revisit.

Atmosfearful:
Replaced by cleanups that were in https://github.com/KlimaDAO/klimadao/pull/1640

PR #1522, opened 2023-09-07
chore: Remove /api prefix from api

0xemc:
## Description

- Removes the `/api` prefix for all routes
- moves extra routes out of `app.ts`


Atmosfearful:
Is this on latest staging commit? Looks like carbonmark is failing from the api change, but I thought we fixed the version to v1.1...

Atmosfearful (approved):

PR #1529, opened 2023-09-09
chore:1448 API Typebox Typing & Refactor

0xemc:
## Related Ticket

Resolves #<issue-number>

## Checklist


carbonmark-api/src/plugins/open-api.ts
```diff
@@ -31,6 +41,20 @@ For a developer guides and example implementations, or to learn more about Carbo
     /** The API version */
     version: packageJson.version,
   },
+  components: {
+    schemas: {
```
Atmosfearful, line 45: I'm afraid this makes our docs more confusing, because the schemas are all listed in the nav menu, but they don't clarify where in the api they are found, nor does the endpoint reference the relevant schema.

I think $refs would be the solution here, and I believe stoplight supports $refs, but for now lets just tear them out from this file to remove them from the nav menu until we figure out a good solution

carbonmark-api/src/models/User.model.ts
```diff
@@ -0,0 +1,20 @@
...
+  username: Type.String(),
+  description: Type.String(),
+  profileImgUrl: Nullable(Type.String()),
+  updatedAt: Type.Number(),
+  createdAt: Type.Number(),
+  wallet: Type.String(),
+  listings: Type.Array(Listing),
+  activities: Type.Array(Activity),
+  //@todo replace unknown with Holding model type
+  assets: Type.Array(Type.Unknown()),
```
Atmosfearful, line 17: cant slap this in real quick? so we dont forget

here is a sample `asset` (aka Holding?) from the prod api
```json
    {
      "id": "0x20a580444dd4a90cc8990da7b480c5e3d605a26f1d9121d3a940e71929d4f6a7707b159a585cbb97",
      "token": {
        "id": "0x1d9121d3a940e71929d4f6a7707b159a585cbb97",
        "name": "C3 Tokens: C3T-VCS-51-2014",
        "symbol": "C3T-VCS-51-2014",
        "decimals": 18
      },
      "amount": "30000000000000000"
    }
 ```

File: carbonmark-api/src/routes/categories/categories.ts
Comment (Atmosfearful) [inline carbonmark-api/src/routes/categories/categories.ts:1]: Thanks for taking the initiative to rename folders, they had been bugging me.
However the naming convention we've followed in other packages is:
`/directory/file.ts` or `/directory/file/index.ts`-- but I avoid `/directory/file/file.ts`

Because the redundant imports always irked me-- e.g. 
```ts
import Component from "../directory/Component/Component"; //🤮
```

Anyways we should just name the handler based on it's `method`:
```
- 📂 routes/
  - 📂 categories/
    - 📄 get.ts
    - 📄 get.schema.ts
  - 📂 projects/
    - 📄 get.ts
    - 📄 get.schema.ts
    - 📂 [id]/
      - 📄 get.ts
      - 📄 get.schema.ts
```

Atmosfearful (approved):
incredible this API package is really looking 🔥 🔥 🔥 now!

A few suggestions at your discretion.

carbonmark-api/src/models/Category.model.ts
```diff
@@ -0,0 +1,4 @@
+import { Static, Type } from "@sinclair/typebox";
+
+export const Category = Type.Object({ id: Type.String() });
+export type CategoryT = Static<typeof Category>;
```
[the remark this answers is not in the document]
0xemc: Cool, done 👍

carbonmark-api/src/models/CommonSchema.model.ts
```diff
@@ -0,0 +1,12 @@
+import { Type } from "@sinclair/typebox";
+import { COMMON_SCHEMA_URI, NetworkParam } from "./NetworkParam.model";
```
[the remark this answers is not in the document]
0xemc: Can do, `app.constants.ts` it is

carbonmark-api/src/models/User.model.ts
0xemc: Yep done, see `Asset.model.ts`

carbonmark-api/src/plugins/open-api.ts
0xemc: The main issue here is that this object is where `feTS` extracts the models used to generate the types in the Carbonmark app :(

I initially tried to use #refs however after a couple of hours sunk I couldn't get it to play nicely with Typebox (I was even messing around with schema resolvers etc..)

carbonmark-api/src/routes/categories/categories.ts
0xemc: Sure, it was half in this direction already. I feel like it's the lesser of two weavels though because cmd+p now gives us many `get.ts` files (likewise for `index.tsx`).

Anyway, for the sake of consistency I've updated them all to match their method. It looks cleaner this way for sure.

PR #1538, opened 2023-09-11
Various API Fixes

0xemc:
Description
This PR refines the user registration and update process in the Carbonmark API. It adds checks for JWT_SECRET, improves error messages and adds pattern validation for user handles.

Changes
* Added JWT_SECRET environment variable check
* Improved error messages for user registration
* Added pattern validation for user handles
* Added tests for user creation and update endpoints
* Updated Firebase admin import
* Added ability to disable authentication during testing


carbonmark-api/src/plugins/bearer.ts
```diff
@@ -1,16 +1,27 @@
...
   });
 
   await fastify.decorate(
     "authenticate",
     async function (request, reply): Promise<void> {
       try {
-        await request.jwtVerify();
+        /** We need the ability to disable authentication when testing */
+        if (!process.env.IGNORE_AUTH) {
+          console.log("SHOULDNT BE HERE");
```
Atmosfearful, line 22: leftover debug comment?

carbonmark-api/test/routes/users/create.test.ts
```diff
@@ -0,0 +1,62 @@
...
+  afterEach(async () => {
+    delete process.env.IGNORE_AUTH;
+    await app.close();
+  });
+  beforeEach(async () => {
+    process.env.IGNORE_AUTH = "true";
+    app = await build();
+  });
+
+  test("should block ethereum address as handle when creating a user", async () => {
```
Atmosfearful, line 47: mind throwing a few happy path tests in there-- want to make sure a DeFi or ReFi brand can represent themselves, some have 0x in their names
`district0x`
`0xProtocol`
`0xymoron`

carbonmark-api/src/plugins/bearer.ts
```diff
@@ -1,16 +1,27 @@
 import fastifyJwt from "@fastify/jwt";
 import fp from "fastify-plugin";
+import { isNil } from "lodash";
+
+const SECRET = process.env.JWT_SECRET;
```
Atmosfearful, line 5: pls add to .env.example and make a note when this variable is required (only for developing on /users POST and PATCH yes?

Atmosfearful (changes requested):

carbonmark-api/src/plugins/bearer.ts
0xemc: Whoops..

carbonmark-api/src/plugins/bearer.ts
0xemc: This is required whenever  the `authenticate` decorator is used. So not just for development but in production too. It is coming from Vercel so it shouldnt need any additional manual steps

carbonmark-api/src/routes/users/create.ts
```diff
@@ -72,21 +85,22 @@ const handler = (fastify: FastifyInstance) =>
...
       .get();
-    // If no documents are found, return a 404 error
+
     if (!userSnapshot.empty) {
       return reply.code(403).send({
         code: 403,
-        error: "This user is already registered!",
+        error: "A user with this handle is already registered!",
       });
     }
```
sprrwhwk, line 105: I find these checks a little confusing:

* `user.exists == true` => **wallet** already registered
* `!userSnapshot.empty == true` => **user** with this handle is already registered

Especially since the former checks explicitly if `user.exists`…

carbonmark-api/test/routes/users/create.test.ts
0xemc: Sure thing, done

carbonmark-api/src/routes/users/create.ts
0xemc: I can see why, in the first case we're confirming that no Document in firebase exists with that wallet (this is the primary key).

In the second we want to make sure that no one is creating a new account with a new address (primary key) but using a handle that someone else has already set on their Document.

Does that make sense?

carbonmark-api/src/routes/users/put.ts
```diff
@@ -21,6 +21,19 @@ const handler = (fastify: FastifyInstance) =>
           profileImgUrl && profileImgUrl.length ? profileImgUrl : null,
       };
 
+      // Check if the handle already exists in our database
```
Atmosfearful, line 24: we only want to do this for `POST` (new user creation). `PUT` is for edits, this will incorrectly throw for all edits because the handle already exists.

Actually this draws my attention to an existing mistake in this PUT endpoint-- we shouldn't pass `handle` to firestore.update()-- the original requirement was to not support handle changes after profile creation

Atmosfearful (changes requested):

carbonmark-api/src/routes/users/post.ts
```diff
@@ -31,35 +31,34 @@ const handler = (fastify: FastifyInstance) =>
...
-      await fastify.firebase
+      const document = await fastify.firebase
         .firestore()
         .collection("users")
         .doc(wallet.toUpperCase())
         .set(createData);
 
       // If the document is successfully created, return the request body
-      return reply.send(request.body);
+      return reply.send(document);
```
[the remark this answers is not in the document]
0xemc: Good catch, reverted to request body for now

carbonmark-api/src/routes/users/put.ts
0xemc: Fair enough, updated and removed handle.

Atmosfearful (approved):

PR #1551, opened 2023-09-14
handles MinPurchaseRequired Error

### LINKED ISSUE #1536: [retire] Pay with CC error & modal UI confusion
testing retire with CC and receipt generation re: #1365 

Form inputs (I was not logged in so had to enter the beneficiary wallet)

Screenshot 2023-09-11 at 4 20 58 PM

Get this error message:

Screenshot 2023-09-11 at 4 19 26 PM

Attempted to try again (both as logged in user and not logged) and get this weird UI bug in the modal?

Screenshot 2023-09-11 at 4 21 15 PM

psparacino:
**NOTE: this PR is dependent on the fiat api PR merge.**

## Description

Handles the MinPurchaseRequired error from the api.

If retirement amount is below the minimum purchase required, the modal informs the user and only gives the option to go back. It also clears the quantity in the retire inputs so that they are forced to enter in another number, as the Retire Carbon button will be disabled until they do.

**NOTE: until the fix on the fiat repo is merged there will still be quote error on the one ton. This error is due to slippage being rounded off inexpensive tons.**

## Related Ticket

Resolves #1536 

## Changes

| Before  | After  |
|---------|--------|
|
image|
 |Screen Shot 2023-09-14 at 1 09 30 PM|

## Checklist

- [X] I have run `npm run build-carbonmark` without errors

### DIFF
--- a/carbonmark/components/pages/Project/Retire/Pool/CreditCardModal.tsx
+++ b/carbonmark/components/pages/Project/Retire/Pool/CreditCardModal.tsx
@@ -24,29 +24,28 @@ export const CreditCardModal: FC<Props> = (props) => {
       onToggleModal={props.onCancel}
     >
       <div className={styles.confirmCreditCard}>
+        {!props.isRedirecting && !props.checkoutError && (
+          <>
+            <Text>
+              <Trans>
+                To complete your credit card transaction, you will be redirected
+                to Stripe, our payment processing partner.
+              </Trans>
+            </Text>
+            <ButtonPrimary onClick={props.onSubmit} label={t`Continue`} />
+            <CarbonmarkButton
+              onClick={
+                !!props.checkoutError ? props.onFiatError : props.onCancel
+              }
+              label={t`Go back`}
+            />
+          </>
+        )}
         {!props.isRedirecting &&
-          (!props.checkoutError ? (
+          !!props.checkoutError &&
+          props.checkoutError.length > 0 && (
             <>
-              <Text>
-                <Trans>
-                  To complete your credit card transaction, you will be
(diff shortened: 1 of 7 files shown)


jabby09 (commented):
@psparacino confused on why we are doing this check on submit of the form vs on the form itself?
Screenshot 2023-09-16 at 12 08 48 PM

We have other validations already on the forms - can we not so the same with this check?

Screenshot 2023-09-16 at 12 12 58 PM

psparacino:
> @psparacino confused on why we are doing this check on submit of the form vs on the form itself? Screenshot 2023-09-16 at 12 08 48 PM
> 
> We have other validations already on the forms - can we not so the same with this check?
> 
> Screenshot 2023-09-16 at 12 12 58 PM

This error comes from the api on submission, thus this was the most straightforward method of initially handling. Yes, that makes sense for consistency though; I  edited this now so the min price is now available on the api before submission to enable an error before submission (not pushed yet). So the flow will be:

1) If the price is too low it will show an error, although they will still be able to click retire and continue (same flow as if the amount is exceeded)
2) If they click continue in the modal, the modal will display the error with only the option to go back and clear the quantity from the form.

Atmosfearful (approved):
lgtm will wait for QA

jabby09 (changes requested):
@psparacino looking good in terms of the end-to-end transaction and receipt generation.

Just a minor copy change to simplify and make consistent.

Let's say "Credit card minimum purchase is $1.00" in both locations (as shown below)

Screenshot 2023-09-21 at 10 22 12 AM

jabby09 (approved):

PR #1556, opened 2023-09-19
Emc/chore/1492-testnet-api-support

0xemc:
# Description 

Updates the gql_sdk to accept a network parameter that is passed to all requests and defaults to "polygon".

FFYI this PR does not generate new types. If the schemas between the test and production nets are expected to differ then we will need to come up with a solution.

## Remaining
- [x] Update Mumbai URLS


carbonmark-api/src/app.ts
```diff
@@ -45,6 +45,15 @@ const app: FastifyPluginAsync<AppOptions> = async (
       error: error.message,
     });
   }));
+
+  fastify.addHook("preValidation", (request: any, reply: any, done) => {
```
Atmosfearful, line 49: what's this?

carbonmark-api/src/app.constants.ts
```diff
@@ -6,11 +6,20 @@ const GRAPH_API_ROOT = "https://api.thegraph.com/subgraphs/name";
...
+    marketplace: `${GRAPH_API_ROOT}/najada/marketplace-matic`,
+    assets: `${GRAPH_API_ROOT}/cujowolf/klima-refi-current-holdings`,
+    offsets: `${GRAPH_API_ROOT}/klimadao/polygon-bridged-carbon`,
+    tokens: `${GRAPH_API_ROOT}/klimadao/klimadao-pairs`,
+    digitalCarbon: `${GRAPH_API_ROOT}/klimadao/polygon-digital-carbon`,
+  },
+  mumbai: {
+    marketplace: `${GRAPH_API_ROOT}/cujowolf/carbonmark-mumbai`,
+    assets: `${GRAPH_API_ROOT}/cujowolf/klima-refi-current-holdings`,
+    offsets: `${GRAPH_API_ROOT}/klimadao/polygon-bridged-carbon`,
```
Atmosfearful, line 19: I like this approach but this is a problem, these graphs don't exist for mumbai (because many of the contracts don't exist on mumbai either), and probably won't for a long time.

Another challenge is that, even though I originally wanted mumbai marketplace subgraph to match the mainnet one, in reality there are some differences, and as they continue to develop/ship Marketplace features, there might be moments of divergence where changes make it to testnet before mainnet.

In my PR trying to get the frontend working in a basic state, I ended up just hot swapping and maintaining two sets of types, which also sucks, but has the benefit of explicitly exposing type differences to the developer...

PR #1579, opened 2023-09-25
[api] v2.0.0 version increment

Atmosfearful:
## Description

Just incrementing package.json


Atmosfearful:
I'm reviewing tickets to see if there is anything I should slide in before we increment and fix the frontend version (just to avoid a second go-around)

0xemc (approved):
Approving but you don't want to isolate the api changes made in (atmos/v2-frontend-plumbing) to here so that they match this new version?

Atmosfearful:
blocked until we have mainet URLS

PR #1585, opened 2023-09-26
feature: 1168 Remove country and category from /projects

0xemc:
## Description
 Remove category and country from the `/projects` endpoint given the schema changes made to the marketplace subgrpah


carbonmark-api/src/routes/projects/get.ts
```diff
@@ -44,19 +46,20 @@ const handler = (fastify: FastifyInstance) =>
...
-      search: request.query.search ?? "",
-    };
+    const allOptions = await getDefaultQueryArgs(fastify);
 
     const [marketplaceProjectsData, poolProjectsData, cmsProjects, poolPrices] =
       await Promise.all([
-        gqlSdk.marketplace.findProjects(query),
-        gqlSdk.offsets.findCarbonOffsets(query),
+        gqlSdk.marketplace.findProjects({
+          vintage: args.vintage ?? allOptions.vintage,
```
Atmosfearful, line 54: any possibility these can be empty strings? if yes then use `||` instead of `??`

carbonmark-api/src/routes/projects/get.ts
```diff
@@ -88,26 +91,38 @@ const handler = (fastify: FastifyInstance) =>
...
+          : true
+      )
+      .forEach((project) => {
+        if (
+          !isValidMarketplaceProject(project) ||
+          !CreditId.isValidProjectId(project.key)
+        ) {
+          return;
+        }
+        const [standard, registryProjectId] = project.key.split("-");
```
Atmosfearful, line 113: fyi the splitProjectId util already invokes isValidProjectId and throws so you might be able to shave that if/type guard a few lines up (not sure)
```suggestion
        const [standard, registryProjectId] = CreditId.splitProjectId(project.key);
```

Atmosfearful (approved):

PR #1588, opened 2023-09-26
#1489 refactor(/projects): only fetch unexpired listings

### LINKED ISSUE #1489: [api] "List projects" endpoint changes for CM contract v2
See https://api.carbonmark.com/#/paths/api-projects/get

**Requirements:**
- [x] Add a property to indicate if listing is “expired” and / or “invalid"
- [x] Update GQL query to not return expired listings

Edit: let's bump the following two requirements to a follow up PR, they are needed for relaunch.
* Add a filter to not return "expired listings"
* Add a filter to not return "invalid listings”

Atmosfearful:
## Description

Closes #1489


carbonmark-api/src/graphql/marketplace.gql
```diff
@@ -45,10 +45,14 @@ query getUserByWallet($wallet: Bytes) {
   }
 }
 
-query findProjects($search: String, $vintage: [BigInt!]) {
+query findProjects(
```
0xemc, line 48: Will need to add 

```ts
 expiresAfter: Type.Optional(
    Type.String({
      description:
        "Only return listings that expire after this timestamp (Unix seconds)",
      default: "Current system timestamp.",
      examples: ["1620000000"],
    })
  ),
```
to the schema for get and provide a default value in the handler (ts is complaining)

Also does this need to be on the /projects/[id] handler too?

0xemc (approved):

0xemc (commented):
Looks like TSC is failing, need to provide some default value on /get

PR #1624, opened 2023-10-03
[api] include project data in Listings model

Atmosfearful:
# Description

Brings back the marketplace project query as a temporary workaround, passes this data down to the listing.project entity. Changes affect /projects, /projects/:id, and /users/:id

Eliminates more bignumber strings from API responses.

Creates a shared formatting util for formatting Listings.


Atmosfearful:
apologies for breaking protocol I need to expedite this one

PR #1641, opened 2023-10-06
Emc/chore/1639 integrate swr with typesafe api client

0xemc:
## Description

As part of PR #1623 it became clear that we were not leveraging swr when interacting with the API using FeTS. This PR replaces FeTS with a full SDK including SWR hooks generated via [kubb](https://www.kubb.dev/quick-start). I tried a few generators and found Kubb to be the best documented, well put together and regularly active (though it has far less stars than some other libraries)

**Note** Everything inside `.generated` probably doesn't need a review so can be skipped

## Changes

* nested NetworkParamModel as `Type.Ref`
* added missing wallet param on /users PUT
* removed manually created swr hooks
* replaced FeTS with Kubb
* increment API to 2.0.0-11

## Note
In order to avoid having to extract the api changes in to a seperate PR I've updated the workflow to allow changes to package.json to trigger the release workflow (see https://github.com/KlimaDAO/klimadao/pull/1645)

If we close this after 1645 then we should automatically get a release.


0xemc:
@biwano @Atmosfearful @psparacino @0xMakka bump on this one.

This will help expose errors and inconsistencies in responses from the api. We will know about more issues at compile time.

carbonmark/.generated/carbonmark-api-sdk/types/Def0.ts
```diff
@@ -0,0 +1,7 @@
+export type Def0 = {
```
sprrwhwk, line 1: Can we give this `Def0` a more meaningful name such as `BlockchainNetwork` or just `Network` during code generation?

carbonmark/.generated/carbonmark-api-sdk/types/Def1.ts
```diff
@@ -0,0 +1 @@
+export type Def1 = "polygon" | "mumbai";
```
sprrwhwk, line 1: Same as the comment in Def0

carbonmark/.generated/carbonmark-api-sdk/types/DetailedProject.ts
```diff
@@ -0,0 +1,234 @@
+export const type2 = {
+  Feature: "Feature",
+} as const;
+export type Type2 = (typeof type2)[keyof typeof type2];
+export const geometryType2 = {
+  Point: "Point",
+} as const;
+export type GeometryType2 = (typeof geometryType2)[keyof typeof geometryType2];
```
sprrwhwk, line 8: The names could be improved here as well

carbonmark/lib/actions.ts
```diff
@@ -317,7 +317,7 @@ export const addProjectsToAssets = async (params: {
     });
     const projectIds = Array.from(projectIdSet);
     const projects = await Promise.all(
-      projectIds.map((id) => getCarbonmarkProject(id))
+      projectIds.map((id) => getProjectsId(id))
```
sprrwhwk, line 320: Shouldn't this ideally be called `getProjectById`, leaving out the plural and adding `by`?

sprrwhwk: Same for purchaseById etc.

carbonmark/.generated/carbonmark-api-sdk/clients/getUsersWalletorhandle.ts
```diff
@@ -0,0 +1,29 @@
...
+  GetUsersWalletorhandleQueryParams,
+  GetUsersWalletorhandleQueryResponse,
+} from "../types/GetUsersWalletorhandle";
+
+/**
+ * @description Get a user's profile and activity
+ * @summary User details
+ * @link /users/:walletOrHandle
+ */
+export async function getUsersWalletorhandle<
```
sprrwhwk, line 14:
```suggestion
export async function getUsersWalletorHandle<
```

sprrwhwk (approved):
Just a couple of naming improvements I thought we could add

carbonmark/lib/api/swr.keys.ts
```diff
@@ -0,0 +1,30 @@
...
+  GetUsersWalletorhandlePathParams,
+  GetUsersWalletorhandleQueryParams,
+} from ".generated/carbonmark-api-sdk/types";
+import { urls } from "lib/constants";
+
+/**
+ * Construct stable URL as key for SWR. When this key changes, data is re-fetched.
+ * Always includes network default 'polygon'
+ */
+export const getUsersWalletorhandleKey = (
```
0xMakka, line 11: Nit: As @sprrwhwk mentioned in other comments, renaming this and all other associated names to:

```suggestion
export const getUsersWalletOrHandleKey = (
```

0xMakka (approved):
Nice job @0xemc.

0xemc:
> Just a couple of naming improvements I thought we could add

@psparacino Unfortunately I don't have a heap of control over the format of the contents in `.generated` I'd need to rename the routes in the api since the naming is generated from the schema.. It may be possible but would require some messing around :(

To be revisited?

carbonmark/.generated/carbonmark-api-sdk/clients/getUsersWalletorhandle.ts
0xemc: Good point, updated

PR #1656, opened 2023-10-12
Add handle to /users put

0xemc:
## Description

"Handle" attribute was missing from /users PUT schema but was expected by EditProfile component.

- increments API version
- adds attribute to request and response


carbonmark/lib/constants.ts
```diff
@@ -149,3 +150,5 @@ export const urls = {
 export const DEFAULT_NETWORK = config.networks[ENVIRONMENT] as
   | "mainnet"
   | "testnet";
+
+console.log(config.urls.api[ENVIRONMENT]);
```
Atmosfearful, line 154: ✂️

carbonmark/lib/constants.ts
```diff
@@ -52,6 +52,7 @@ export const ENVIRONMENT: Environment =
     preview: IS_PREVIEW,
   }).first() ?? "preview";
 
+console.log(process.env.NEXT_PUBLIC_CARBONMARK_API_URL);
```
Atmosfearful, line 55: ✂️

carbonmark/.generated/carbonmark-api.schema.ts
```diff
@@ -12,7 +12,7 @@ export default {
       "name": "MIT",
       "url": "https://github.com/KlimaDAO/klimadao/blob/main/LICENSE"
     },
-    "version": "2.0.0-6"
+    "version": "2.0.0-8"
```
Atmosfearful, line 15: how were you able to generate this if v2.0.0-8 does not have a release yet?

This is why the build fails i assume

Atmosfearful (changes requested):

carbonmark/lib/constants.ts
0xemc: Whoops..

carbonmark/.generated/carbonmark-api.schema.ts
0xemc: 1. Run the local dev server: `npm run dev-carbonmark-api`
2. run `cd carbonmark && NEXT_PUBLIC_CARBONMARK_API_URL=http://localhost:3003 npm run generate:types` 
3. Set the `NEXT_PUBLIC_USE_PREVIEW_CARBONMARK_API` in vercel to this branch (the first build when doing this will always fail because the preview deployment hasn't finished until after the first pass)

Atmosfearful (approved):

PR #1693, opened 2023-10-17
fetchIsNewHandle to use response status instead of error message

0xemc:
## Description

`fetchIsNewHandle` was using the error message returned rather than the error status. 

* Added FetchError
*Check for error status 404 instead of string compare

### DIFF
--- a/carbonmark/lib/api/client.ts
+++ b/carbonmark/lib/api/client.ts
@@ -12,11 +12,24 @@ export const client = createClient<NormalizeOAS<typeof schema>>({
     // If the status code is not in the range 200-299,
     // we still try to parse and throw it.
     if (!res.ok) {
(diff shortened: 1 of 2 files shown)


carbonmark/lib/api/client.ts
```diff
@@ -12,11 +12,24 @@ export const client = createClient<NormalizeOAS<typeof schema>>({
...
+      throw new FetchError((await res.json()).message, res.status);
     }
 
     return res;
   },
 });
+/** Extend Error to allow network status to be returned */
+class FetchError extends Error {
+  constructor(
+    public message: string,
```
Atmosfearful, line 24: neat never seen this property assignment shorthand before

Atmosfearful (approved):

PR #1709, opened 2023-10-18
Feature/etherspot transaction buidler widget

poocart:
## Description

This PR includes Etherspot SDK (https://etherspot.io) implementation to enable cross-chain staking using Etherspot Buidler (https://buidler.etherspot.io) depdendency.

It enables users to stake Klima using any asset on supported chains and receive sKlima on Polygon.

It was QA tested on our end, but feel free to test it once again on your end.

image

## Changes

| Before  | After  |
|---------|--------|
|
image
|image|

-->

## Checklist


Atmosfearful:
Approved `app` preview deployment will take a closer look soon.

Atmosfearful (changes requested):
The addition of this dependency has increased the size of the app bundle by 240%
- In addition I have major security concerns about auto-connecting our users wallets to a third party who can trigger transaction requests.
- I am also concerned about users who may think this is something we have built and maintained-- we don't want users coming to us with transaction issues or bug reports that we can't resolve.
-  Because this is a separate protocol and dapp, in my opinion it is a better UX to transfer the user to your domain (this is also better for your organization, for brand exposure and to capture users).

I apologize if Archi or someone else made promises to you guys & you spent significant time on this widget integration, but I suggest we take this in a different direction:
1. The UI or widget should be hosted on your domain and maintained by you guys
2. We can update the "cross-chain staking" call to action, update the URL, and keep it where it is.

I will close this PR as most the changes won't be relevant for the above solution.

Side note: this "buy" button seems broken, says 'deploying...' but nothing happens
image

PR #1729, opened 2023-10-23
Makka / 1333 - upgrade wallet connect to v2

### LINKED ISSUE #1333: Upgrade WalletConnect web3-provider to ethereum-provider
> https://docs.walletconnect.com/2.0/advanced/migration-from-v1.x/dapps/#web3-provider

> If you are using @walletconnect/web3-provider, we stopped supporting this provider in favor of the improved version published under @walletconnect/ethereum-provider and you will be able to check the latest version on NPM [here](https://npmjs.com/package/@walletconnect/ethereum-provider).

0xMakka:
## Description

Upgrade wallet connect to v2. 

**Note:** We need to create a projectId on https://cloud.walletconnect.com. I've temporarily created one and added it to enable testing but will remove once the official projectId is set up and added as an env variable.

Tests to cover:
- Login with wallets listed in the login modal
- Logout with all wallets and ensure user is logged out correctly.
- Switch accounts when logged in and ensure wallet switch works as expected.
- Login/Logout with a wallet in the wallet connect list and ensure it connects successfully with the correct address

## Related Ticket

Resolves #1333 

## Checklist

### DIFF
--- a/lib/.env.example
+++ b/lib/.env.example
@@ -1,4 +1,0 @@
-# Create a file called .env.local and provide these keys.
-
-# Required for Wallet Connect v2
-NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID=12345
--- a/lib/.env.example
+++ b/lib/.env.example
@@ -1,0 +1,4 @@
+# Create a file called .env.local and provide these keys.
+
+# Required for Wallet Connect v2
+NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID=12345
--- a/lib/constants/index.ts
+++ b/lib/constants/index.ts
@@ -226,6 +226,8 @@ export const FALLBACK_BLOCK_RATE = 2.3;
 
 export const API_BASE_URL = "https://www.klimadao.finance/api";
 
+export const WALLETCONNECT_PROJECT_ID = "51d2bf0bcd3abd445f503942db20b1f9"; // @todo - Makka add projectId
+
 /** CMS stuff  */
 export const SANITY_STUDIO_API_PROJECT_ID = "dk34t4vc";
 export const SANITY_STUDIO_API_DATASET = "production";
@@ -302,7 +304,3 @@ export const verra = {
 export const goldStandard = {
   projectDetailPage: `${REGISTRIES.GoldStandard.url}/projects/details`,
 };
-
-export const WALLETCONNECT_PROJECT_ID =
-  process.env.NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID ??
-  "51d2bf0bcd3abd445f503942db20b1f9"; // @todo Makka - remove hardcoded projectId after official projectId setup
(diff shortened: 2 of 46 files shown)


0xtapi:
Hey there,

I encountered no technical issues using WalletConnect to connect to Ledger Live on both KlimaDAO and Carbonmark. Seems to work as intended. I also tried reconnecting/disconnecting a few times and switching to MetaMask in between to check if this would somehow screw it up - but it didn't. Tested it on Firefox as well just to make sure. No issues.

I did run into a minor problem when I manually refused the connection, however. See screenshot:

cmblue

1. It uses "Carbonmark Blue" instead of "Klima Green" for the button element (which is also the case in the current live environment).
2. Black text on gray background isn't impossible to read, but it's not optimal. Switching to white would be better.

With a few minor design tweaks, this should be good to go!

lib/.env.example
```diff
@@ -0,0 +1,4 @@
+# Create a file called .env.local and provide these keys.
+
+# Required for Wallet Connect v2
+NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID=12345
```
Atmosfearful, line 4: unfortunately everything in `lib` should remain environment-agnostic so all variables must be passed in :-/

lib/constants/index.ts
```diff
@@ -302,3 +302,7 @@ export const verra = {
 export const goldStandard = {
   projectDetailPage: `${REGISTRIES.GoldStandard.url}/projects/details`,
 };
+
+export const WALLETCONNECT_PROJECT_ID =
+  process.env.NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID ??
```
Atmosfearful, line 307: I'll try to get this set up in our environments by EOD today

lib/utils/useProvider/index.tsx
```diff
@@ -199,15 +205,16 @@ export const useProvider = (): Web3ModalState => {
...
     web3state.provider.provider.on("chainChanged", handleChainChanged);
     /** For WalletConnect and Coinbase disconnections */
     web3state.provider.provider.on("disconnect", handleDisconnect);
 
     return () => {
-      web3state.provider.provider.removeListener(
-        "accountsChanged",
-        handleAccountsChanged
+      web3state.provider.provider.removeListener("accountsChanged", (accts) =>
+        handleAccountsChanged(accts)
```
Atmosfearful, line 217: T inline arrow function creates a new instance, so the `removeListener` might not be able to find and remove the handler that was instantiated earlier.

That's why I reuse the same reference to `handleAccountsChanged` directly

Atmosfearful (changes requested):

lib/.env.example
0xMakka: Ah okay, apologies my bad I should have realised when there was no example env file in the lib directory. I'll remove this file and leave just leave the projectId key in `lib/constants/index.ts`.

lib/utils/useProvider/index.tsx
0xMakka: Yeah that makes sense. Typescript was giving me trouble here hence the change to it - will take another look.

Atmosfearful (approved):

Atmosfearful:
Works great with Torus! And the modal appears on all 3 applications. Nice work!

Issue #1750, opened 2023-10-24
[carbonmark-api] `/users/[id]` - Support `deleted` param and set to `false` by default


Atmosfearful:
**Describe the bug**
I.e. do not show deleted listings in the default API response

Atmosfearful:
alternatively, `active: true` by default. I believe these are redundant properties

PR #1775, opened 2023-10-26
[carbonmark] fix safemath

Atmosfearful:
## Description
Converting a decimal with .toString() was creating strings with exponent syntax `1.12e10`which would cause the ethers parseUnits to fail.
Now we do all the math with strings and convert to number at the end.

Also snuck in a change to the gql query to not return deleted listings related to #1750

## Checklist


0xMakka (approved):
Good to know, and apologies I didn't spot this when making the changes.

PR #1813, opened 2023-11-01
Emc/chore/1638-users-tests

0xemc:
## Description

`/Users` endpoint tests:

  - "by wallet",
  - "by handle",
  - "with invalid handle",
  - "with invalid address or handle",

The following:

  - "with mumbai network parameter",
  - "with expires after parameter"

have been skipped until issues around param matching and nock can be resolved.


Atmosfearful (approved):
Are the skipped tests intentional? If so include a `why` comment

PR #1828, opened 2023-11-05
Emc/feature/1823-demo-package

0xemc:
## Description

`/retirement-demo`


Atmosfearful:
Let's save this branch for future reference. Sad we didn't get more use out of this due to EcoRegistry speedbumps but it was a good hustle on your end 🙏

PR #1831, opened 2023-11-06
carbon: make tabs and options sticky on mobile

### LINKED ISSUE #1763: [klima data - mobile] suboptimal button alignments on token details
Page: https://carbon-git-staging-klimadao.vercel.app/token-details

Design:
Screenshot 2023-10-25 200655

Current implementation:
Screenshot_20231025-193004-925

While this does seem to match design - the upper two button rows seem to be aligned to the start and end of that row and the lower button row seems to be centered - I think it's not the greatest look. Even if the code says it's perfect, to the eye (or at least my eye) the alignments seem off due to the different widths of the button row elements. Maybe we can design a better solution? Pinging @theprofessora for thoughts!

sprrwhwk:
## Description

This PR makes the mobile tab selector as well as the chart options sticky.

## Related Ticket

Resolves #1763 

## Checklist

### DIFF
--- a/carbon/components/MobileTabSelector/index.tsx
+++ b/carbon/components/MobileTabSelector/index.tsx
@@ -11,7 +11,7 @@ export const MobileTabSelector = <T extends string>(props: {
   value: string;
   options: Array<TypedOption<T>>;
   onSelectionChanged: (value: T) => void;
-  className: string;
+  className?: string;
 }) => {
   const handleChange = (event: SelectChangeEvent) => {
     props.onSelectionChanged(event.target.value as T);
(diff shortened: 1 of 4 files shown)


carbon/components/MobileTabSelector/index.tsx
```diff
@@ -11,15 +11,15 @@ export const MobileTabSelector = <T extends string>(props: {
...
+  className?: string;
 }) => {
   const handleChange = (event: SelectChangeEvent) => {
     props.onSelectionChanged(event.target.value as T);
   };
 
   return (
     <Select
-      className={`${styles.select} ${props.className}`}
+      className={`${styles.select} ${props.className ?? ""}`}
```
biwano, line 22: I do not thing you need the  `?? ""` part

biwano (approved):
Beautiful

jabby09 (approved):

carbon/components/MobileTabSelector/index.tsx
sprrwhwk: I forgot to consider this, I’ll create a separate PR. I’m sorry!

PR #1833, opened 2023-11-06
carbon: make the whole fact clickable

### LINKED ISSUE #1768: [klima data - mobile] info hover impossible to use
See screenshot:
Screenshot_20231025-210352-570

User is supposed to be able to hover over the small info button right behind "SELECTIVE COST". This is not a problem on Desktop, but nearly impossible on Mobile, because 1.) button is too small to touch, 2.) you can't hover with your finger.

But the information presented is still very useful because otherwise the user would have no idea why "PRICE PER TONNE" and "SELECTIVE COST" are different.

So I wonder if we can come up with a better solution. @theprofessora, here's a ping for you!

sprrwhwk:
## Related Ticket

Resolves #1768

## Checklist

### DIFF
--- a/carbon/components/cards/overview/TokensPriceCard/index.tsx
+++ b/carbon/components/cards/overview/TokensPriceCard/index.tsx
@@ -1,7 +1,6 @@
 import { formatTonnes } from "@klimadao/lib/utils";
 import { t } from "@lingui/macro";
 import { InfoOutlined } from "@mui/icons-material";
-import { Tooltip } from "@mui/material";
 import PercentageChange from "components/PercentageChange";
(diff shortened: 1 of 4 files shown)


jabby09 (commented):
@sprrwhwk the target has changed and looks good

in the process, did you adjust the font size of the info message? It is very very small, not readable really.

sprrwhwk:
> @sprrwhwk the target has changed and looks good
> 
> in the process, did you adjust the font size of the info message? It is very very small, not readable really.

Can you give it another try? I tweaked the font size.

jabby09 (approved):
all good now!

PR #1859, opened 2023-11-10
Makka / 1791 prevent users from retiring listed tonnes

### LINKED ISSUE #1791: [carbonmark] `/portfolio/[asset]/retire` form should prevent users from retiring listed tonnes.
Currently its possible to use this form to retire tonnes you already have listed
image

currently this button is disabled if `unlistedTonnes === 0`

but this is not enough. The button takes you to a form, and this form needs needs to prevent you from retiring more than your unlisted balance.

# Solution

in /portfolio/[asset]/retire page

1. Fetch the connected /user
2. Utilize the utils we have defined in `listings.utils.tsx`
3. Only allow the user to retire unlisted balance (total asset amount, minus listed tonnes)

May require changes to `getServerSideProps` ?https://github.com/KlimaDAO/klimadao/blob/1aedc35cf348203e5e804469a8e2e5d8f751b9ab/carbonmark/pages/portfolio/%5Bproject_token_0x%5D/retire/index.tsx#L11

Or probably can be done all client side fetching w/ swr ?

0xMakka:
## Description

Prevents users from being able to retire already listed tonnes, using the unlistedBalance util.

## Related Ticket

Resolves #1791 

## Checklist

### DIFF
--- a/carbonmark/components/pages/Portfolio/AssetProject/index.tsx
+++ b/carbonmark/components/pages/Portfolio/AssetProject/index.tsx
@@ -1,3 +1,4 @@
+import { cx } from "@emotion/css";
 import { Trans } from "@lingui/macro";
 import { ButtonPrimary } from "components/Buttons/ButtonPrimary";
 import { CarbonmarkButton } from "components/CarbonmarkButton";
@@ -82,9 +83,12 @@ export const AssetProject: FC<Props> = (props) => {
       <div className={styles.buttons}>
         <ButtonPrimary
           label={<Trans>Retire</Trans>}
-          disabled={unlistedBalance <= 0}
           href={`/portfolio/${props.asset.token.id}/retire`}
-          renderLink={(linkProps) => <Link {...linkProps} />}
+          renderLink={(linkProps) => (
+            <div className={cx({ [styles.disabled]: unlistedBalance <= 0 })}>
+              <Link {...linkProps} />
+            </div>
+          )}
           onClick={() => {
             LO.track("Retire: Retire Button Clicked");
           }}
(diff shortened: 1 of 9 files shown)


jabby09:
> Edit after Mainnet test ;)
> 
> Solution seems to work! I cannot enter any value higher than the amount that's unlisted ✅
> 
> However, there's a small issue here:
> 
> Screenshot 2023-11-11 141906
> 
> Bottom arrow section always shows the right value. Upper arrow section always shows NaN (I tried with different values). Could be confusing to users and should be fixed (but I'm not sure if this is related to this PR?).

@0xtapi I can't repeat this error in the PR preview link
Available to retire tonnes showing properly. I'm using Brave.

Screenshot 2023-11-11 at 9 45 06 AM

0xMakka:
@0xtapi I'm the same as @jabby09, I can't reproduce the issue with the tonnes value on the side card. Have checked it with a few listings.

Screenshot 2023-11-13 at 10 32 35

jabby09 (approved):

0xtapi (approved):

Atmosfearful (approved):
Looks good thanks
cleanup idea: see if you can overwrite props.asset.balance in the parent level.

What i mean is remove the unlisted balance calculations out of `RetireForm/index.tsx` here:
image

to somewhere higher in the component tree, and just pass down a computed `asset` prop with `balance` overwritten by the actual `unlistedBalance` value. Or, alternative, pass down a new property `unlistedBalance`

Just reduces the amount of prop drilling (no more need to pass down `user`, for example)

But you gotta double check not to break anything that expects the real user total balance

PR #1865, opened 2023-11-13
[API] Fix selective redemption price

### LINKED ISSUE #1394: [api] Selective redemption price estimation is incorrect
In the handler for `/project/[project-id]` we calculate prices for each pool. We first get the pool prices from the graph, which should be fairly accurate. Then we manually derive the cost of selectively redeeming 1 tonne from that based on the known redemption fee % for each pool.

This calculation occurs here:

https://github.com/KlimaDAO/klimadao/blob/staging/carbonmark-api/src/utils/helpers/fetchAllPoolPrices.ts#L13

Percentages are hardcoded here https://github.com/KlimaDAO/klimadao/blob/staging/carbonmark-api/src/routes/projects/projects.constants.ts#L33

It seems we have missed the 1% aggregator fee and sushiswap fees. See [the new fee docs](https://docs.carbonmark.com/get-started/understanding-fees-on-carbonmark)

For debugging, query the aggregator directly to get a real time price for 1 tonne.

biwano:
## Description

Fix selective redemption price computation API side. Another PR will be needed to use these updated prices on the frontend

## Related Ticket

#1394

## Checklist

### DIFF
--- a/carbonmark-api/README.md
+++ b/carbonmark-api/README.md
@@ -24,8 +24,13 @@ You will need to be authenticated to the following in order to run the project:
 
 ### Link your local copy of the project with vercel
 
+Install the vercel CLI
 
-Run the following from the root of the repository, accepting all options:
+```sh
+npm i -g vercel
+```
+
+Run the following from the root of the repository, accepting all options and using KlimaDAO as the project scope:
 
 ```sh
 vercel link -p carbonmark-api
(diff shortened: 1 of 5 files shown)


Atmosfearful (approved):

0xemc (approved):

0xemc (approved):
Would it be beneficial to have some tests for these calculations? (i.e the 1% fee etc)

biwano:
> Would it be beneficial to have some tests for these calculations? (i.e the 1% fee etc)

Yes. I did not address this. Will do another PR.

PR #1884, opened 2023-11-14
API: enable the use of paths from root with tsnd

biwano:
## Description

This PR may enable us to replace relative imports like
`import { CreditId } from "../../../utils/CreditId";`
with
`import { CreditId } from "src/utils/CreditId";`

## Related Ticket

NA

## Changes

| Before  | After  |
|---------|--------|
|img here |img here|

-->

## Checklist


biwano:
Fails on Vercel. We would need that: https://dev.to/omills/path-mapping-with-vercel-functions-nodejs-typescript-22li

Issue #1377, opened 2023-07-27
[API] Move project activity to a new /activity route 


Atmosfearful:
Currently when fetching a `Project` from `/projects/[id]` we also fetch recent activity for that project.

We have identified this as a major performance cost in #1360 and #1367 - I also anticipate more features around sorting, filtering and exploring that activity.

For these reasons I think it makes sense to treat `Activity` as it's own resources and serve it from a sub-path
`/projects/[id]/activity`

This is a breaking change with a new API version. A separate ticket will be needed to consume this new activity feed on the frontend.

jabby09:
@biwano here is a new issue for you when you are free of Klima Data tasks

biwano:
@Atmosfearful If I understand the release process I will make 3 pull requests:

- [X] Modify the API to Create the new activity route: #1888 
- [ ] Modify the UI to use the new activity route
- [ ] Modify the API to remove activity information when querying projects

PR #1893, opened 2023-11-15
Use the new activity route in Carbonmark UI

biwano:
## Description

- Use the new activity route in carbonmark UI

Should be merged after https://github.com/KlimaDAO/klimadao/pull/1984

## Related Ticket

Resolves https://github.com/KlimaDAO/klimadao/issues/1377

## Changes

| Before  | After  |
|---------|--------|
|img here |img here|

## Checklist


biwano:
@Atmosfearful @0xemc When I check this PR preview, my code does not work. Should I investigate or is this normal because my new API endpoint is not yet deployed in production?

0xemc:
> @Atmosfearful @0xemc When I check this PR preview, my code does not work. Should I investigate or is this normal because my new API endpoint is not yet deployed in production?

@biwano I've set your branch to the NEXT_PUBLIC_USE_PREVIEW_CARBONMARK_API  env var in Vercel and redeployed your Carbonmark build.

This means that your API from this PR will be targeted by the Carbonmark package.

Once that redeploy finishes try and access that route from the app again. (Additionally you should be able to check that the endpoint exists on the preview deployment [here](https://carbonmark-api-git-carbonmark-usenewactivityroute-klimadao.vercel.app/)

carbonmark-api/src/routes/projects/[id]/activity/get.ts
Comment (0xemc) [inline carbonmark-api/src/routes/projects/[id]/activity/get.ts:1]: Are you able to make the route `/routes/activity` and then the user of the endpoint can provide filter options vi query params?

I think a route per entity is more re-usable and allows us to handle future use cases without re-work.

0xemc (changes requested):

carbonmark-api/src/routes/projects/[id]/activity/get.ts
Comment (biwano) [reply carbonmark-api/src/routes/projects/[id]/activity/get.ts:1]: Agreed. What do you think of having the `routes/activitiy` as a  "main" endpoint AND the `/project/[id]/activity` as an "alias" endpoint. I would like that as a casual API user because it seems intuitive.
@Atmosfearful @0xemc

Comment (0xemc) [reply carbonmark-api/src/routes/projects/[id]/activity/get.ts:1]: Sure that works, as long as the alias just calls `/activity` with some query params for id similar to the `/project` endpoint

carbonmark-api/src/routes/activities/get.utils.ts
```diff
@@ -0,0 +1,40 @@
...
+import { fetchProjectActivities } from "../../utils/helpers/fetchMarketplaceListings";
+import { stringsToActivityTypes } from "../../utils/helpers/utils";
+import { schema } from "./get.schema";
+
+/**
+ * This handler fetches activities for a carbon project
+ *
+ * @returns
+ */
+export const getActivities = async (
```
0xMakka, line 18: Thoughts on having this util function return the activities and let the routes handle replies/returning the data? There would be no need to pass the fastify instance and reply param to this function?

For example:
```
export const getActivities = async (
  query: Static<typeof schema.querystring>,
  network?: NetworkParam
): Promise<Activity[]> => {
  // ...
  return await fetchProjectActivities(sdk, { ... });
};
```

And then in the route files:

```
const handler = (fastify: FastifyInstance) =>
  function (
    request: FastifyRequest<{
      Querystring: Static<typeof schema.querystring>;
    }>,
    reply: FastifyReply
  ): Promise<Activity[]> {
    const activities = getActivities(request.query, request.query.network);
    return reply.status(200).header("Content-Type", "application/json; charset=utf-8").send(activities);
  };
  ```

File: carbonmark-api/src/routes/activities/get.utils.ts
Comment (0xemc) [inline carbonmark-api/src/routes/activities/get.utils.ts:1]: I'm not sure this file is necessary, we have a `getActivities` helper that uses a `fetchProjectActivities` helper. Can the logic in this file be put in the handler directly instead?

Comment (biwano) [reply carbonmark-api/src/routes/activities/get.utils.ts:1]: It is used in both `route/activities` and `projects/[id]/activity`. It serves as a layer of abstraction so we are sure that both routes behave the same. I could put this logic in the new ``activities.utils.ts` file though

File: carbonmark/components/pages/Projects/index.tsx [hunk 2]
```diff
@@ -70,11 +71,11 @@ const Page: NextPage = () => {
...
-  if (!router.isReady) {
-    // need to prevent the initial grid view from flashing initially
-    // after setting the layout as "list" and reloading the browser.
-    return null;
-  }
+  // need to prevent the initial grid view from flashing initially
+  // after setting the layout as "list" and reloading the browser.
+  useEffect(() => {
+    setReady(router.isReady);
+  }, [router.isReady]);
```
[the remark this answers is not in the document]
biwano: So this is the hydratation fix. See https://github.com/KlimaDAO/klimadao/pull/1943

And we do not show anything before the router is ready so the layout does not flash (shows grid layout and when the router is ready show list layout)

carbonmark-api/src/utils/helpers/fetchMarketplaceListings.ts
```diff
@@ -96,21 +118,59 @@ export const fetchMarketplaceListings = async (
...
+    data = await sdk.marketplace.getAllActivities({
+      activityType,
+    });
+  }
+  const activitiesWithProfiles = await mapUserToActivities(
+    data.activities,
+    fastify
+  );
+  return activitiesWithProfiles;
 };
```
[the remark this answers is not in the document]
biwano: Done in #dccf9345e
I called the file `activities.util.ts` to follow the `users.utils.ts` pattern

carbonmark-api/src/routes/activities/get.utils.ts
0xemc: Fair enough 👍

carbonmark/lib/api/client.ts
```diff
@@ -19,11 +17,28 @@ export type ApiError<TError> = {
...
+      if (Array.isArray(value))
+        value.forEach((v) => {
+          query.push(`${key}=${encodeURIComponent(v)}`);
+        });
+      else if (value) return query.push(`${key}=${encodeURIComponent(value)}`);
+    });
+    return `?${query.join("&")}`;
+  }
+  return "";
+}
```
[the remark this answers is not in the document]
biwano: Héhé, nice catch. It returns `?`
Fixed in #a671cde86

carbonmark/components/pages/Project/index.tsx
```diff
@@ -45,6 +48,15 @@ export type PageProps = {
 
 const Page: NextPage<PageProps> = (props) => {
   const { data: project } = useGetProjectsId(props.projectID);
+  const { data: activities } = useGetProjectsIdActivity(props.projectID, {
```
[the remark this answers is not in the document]
biwano: Done in 29fd4a049a3726b46ce178b2f9004ac1f10547e3

The preloading for the project data was broken too because of the way the SDK generates SWR keys vs how they were generated before. Please check.

It must be broken in other pages too. But maybe we can make a global check in a separate PR.

biwano:
> I just realised also that the `API_PROD_URL`
> 
> https://github.com/KlimaDAO/klimadao/blob/03dcdd07cc9c0441ddbfc74df9bb770907d9a082/carbonmark/lib/constants.ts#L31
> 
> Will need to be incremented to match the new API version as well for this to work in production.

Done in 0684e205713a40c87c737ae39b7ce6ae8d0b6e00

Should we make this direcly match the value in `carbonmark-api/package.json` ?

carbonmark/components/pages/Project/index.tsx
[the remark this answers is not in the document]
0xemc: Yea unfortunately this was an unexpected result of Kubb. Technically the preloading wasnt working anyway because we weren't leveraging swr on any of our calls before anyway. Will fix asap

0xemc:
> > I just realised also that the `API_PROD_URL`
> > https://github.com/KlimaDAO/klimadao/blob/03dcdd07cc9c0441ddbfc74df9bb770907d9a082/carbonmark/lib/constants.ts#L31
> > 
> > Will need to be incremented to match the new API version as well for this to work in production.
> 
> Done in [0684e20](https://github.com/KlimaDAO/klimadao/commit/0684e205713a40c87c737ae39b7ce6ae8d0b6e00)
> 
> Should we make this direcly match the value in `carbonmark-api/package.json` ?

Yes exactly

carbonmark/lib/api/client.ts
```diff
@@ -19,11 +17,25 @@ export type ApiError<TError> = {
...
+      if (Array.isArray(value))
+        value.forEach((v) => {
+          query.push(`${key}=${encodeURIComponent(v)}`);
+        });
+      else if (value) return query.push(`${key}=${encodeURIComponent(value)}`);
+    });
+    if (query.length) return `?${query.join("&")}`;
+  }
+  return "";
+}
```
0xemc, line 38:
```suggestion
/**
 * Transforms a request params object into a valid query string
 */
export function serializeParams(params: ParamsObject): string {
  // Remove null and undefined values
  const validparams = pickBy(params, notNil);
  // Convert to valid url params
  const urlparams = new URLSearchParams(validparams).toString();
  // Prepend '?' if required
  return urlparams ? `?${urlparams}` : "";
}

```

carbonmark/.generated/carbonmark-api-sdk/client.ts
```diff
@@ -19,11 +17,25 @@ export type ApiError<TError> = {
...
+      if (Array.isArray(value))
+        value.forEach((v) => {
+          query.push(`${key}=${encodeURIComponent(v)}`);
+        });
+      else if (value) return query.push(`${key}=${encodeURIComponent(value)}`);
+    });
+    if (query.length) return `?${query.join("&")}`;
+  }
+  return "";
+}
```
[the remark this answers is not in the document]
biwano: > I think we can do this a little more comprehensibly:
> 
> Did you find some instances where arrays were being passed?

@0xemc 
I am passing arrays in this PR: https://github.com/KlimaDAO/klimadao/blob/0684e205713a40c87c737ae39b7ce6ae8d0b6e00/carbonmark/components/pages/Project/index.tsx#L61C34-L61C34

Here is the rationale behind that:
- we want (I think) to declare arrays in the fastify schema:  https://github.com/KlimaDAO/klimadao/blob/0684e205713a40c87c737ae39b7ce6ae8d0b6e00/carbonmark-api/src/routes/activities/get.schema.ts#L15 . 
Note that this is not what we do now. We declare strings and we manually split them: https://github.com/KlimaDAO/klimadao/blob/0684e205713a40c87c737ae39b7ce6ae8d0b6e00/carbonmark-api/src/routes/projects/get.ts#L41
- To make query arrays parameters work natively with fastify, they must be formatted this way: ?`country=belgium&country=france` instead of `country=belgium,france` (which is URLSearchParams behaviour).
- Thats is why, I am handling the parameters manually, 

WDYT?

carbonmark-api/src/utils/helpers/activities.utils.ts
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Query the subgraph for project activities
+ */
+export const fetchProjectActivities = async (
+  sdk: GQL_SDK,
+  { projectId, activityType, fastify }: ActivitiesParams
+): Promise<Activity[]> => {
+  let data;
+  // FIXME: Is it possible to do that with the same gql query template?
```
Atmosfearful, line 73: Hmm not sure what this means, do you mean is it possible to replace the if/else into a single statement?

Atmosfearful (approved):
Looks beautiful 💯 💯 💯

carbonmark-api/src/utils/helpers/activities.utils.ts
biwano: > Hmm not sure what this means, do you mean is it possible to replace the if/else into a single statement?

Yeah exactly.

biwano:
> @biwano Looks beautiful 💯 💯 💯 I would try to parallelize that request in the projects page and then we should be good to go. Hopefully no conflicts with the other /projects PR?

There will be, all my PRs are conflicting with each other, merging them is going to be a pain. Better be careful.

carbonmark/lib/api/client.ts
biwano: Same as above?

carbonmark/pages/projects/[project_id]/index.tsx
```diff
@@ -22,6 +29,9 @@ export const getStaticProps: GetStaticProps<PageProps, Params> = async (
     if (!project) {
       throw new Error("No project found");
     }
+    const activities = await getProjectsIdActivity(params.project_id, {
```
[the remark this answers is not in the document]
Comment (biwano) [reply carbonmark/pages/projects/[project_id]/index.tsx]: Done in 0c9dead53

Atmosfearful (dismissed):
Please separate the API changed into a different PR that we can merge and deploy first.
This will reduce conflicts and make it easier to roll back the FE if we have issues.

biwano:
> Please separate the API changed into a different PR that we can merge and deploy first. This will reduce conflicts and make it easier to roll back the FE if we have issues.

Done here https://github.com/KlimaDAO/klimadao/pull/1984

carbonmark/components/pages/Project/index.tsx
```diff
@@ -274,7 +274,6 @@ export const Project: NextPage<PageProps> = (props) => {
         fetcher,
         fallback: {
           [`/projects/${props.projectID}`]: props.project,
-          [`/projects/${props.projectID}/activity`]: props.activities,
```
0xemc: Was this causing issues?

carbonmark/lib/constants.ts
```diff
@@ -28,7 +28,7 @@ const SHORT_COMMIT_HASH = process.env.NEXT_PUBLIC_VERCEL_GIT_COMMIT_SHA?.slice(
 );
 
 /** When incrementing this API version, be sure to update TypeScript types to reflect API changes */
-export const API_PROD_URL = "https://v2.1.0.api.carbonmark.com";
+export const API_PROD_URL = "https://v3.0.0.api.carbonmark.com";
```
0xemc, line 31: Better to do this bump and the carbonmark changes in a seperate PR

0xemc (changes requested):
Need to split these changes up as per usual, one PR for Carbonmark one for the API

0xemc (approved):

PR #1894, opened 2023-11-15
No longer fetch activities when querying a project

### LINKED ISSUE #1377: [API] Move project activity to a new /activity route

biwano:
## Description

The API no longer fetch activities when fetching a project by id
Should not be merged before https://github.com/KlimaDAO/klimadao/pull/1893

## Related Ticket

PR 3/3 
Resolves #1377

## Checklist


biwano:
Closed in favor of #1893

PR #1912, opened 2023-11-20
Carbonmark: Prevent carbonmark user fetch from api from failing

### LINKED ISSUE #1911: [profile] 500 error when accessing some user profiles
Some 0x addresses are causing a 500 error when accessing the profile page.

From user reporting issue:
> This occurs for me when creating a new profile for an address that hasn't previously connected to CM before, as well as when using wallet addresses that previously had a Pledge page set up but haven't connected to CM.
> 
> We have a few partners that want to utilize a CM profile page to indicate ownership over an address by uploading their logo. Unfortunately they're getting this error when trying to access that page.

image

For example:
- Address from a partner that produces this error: https://www.carbonmark.com/users/0xbefd98974f844f464e387737229624128a295d92. Note that their retirement page works fine.
Screenshot 2023-11-20 at 9 25 16 AM
- Another example: https://www.carbonmark.com/users/0x5146c0cEb234B8856079dBB9b312E958FEF86924
- This user (jabby) works correctly https://www.carbonmark.com/users/0x6476edf98Bdc5EE62f9D4E4BF8932d69Ed685483. Resolves to https://www.carbonmark.com/users/jabby

Wondering if this has something to do with users who have profiles created (and therefore handles) vs those who do not.

biwano:
## Description

Prevent fetching carbonmark user from api from failing when rendering the profile page

## Related Ticket

Resolves #1911 

## Checklist

### DIFF
--- a/carbonmark/pages/users/[user].tsx
+++ b/carbonmark/pages/users/[user].tsx
@@ -5,6 +5,7 @@ import { VALID_HANDLE_REGEX } from "lib/constants";
 import { loadTranslation } from "lib/i18n";
 import { getAddressByDomain } from "lib/shared/getAddressByDomain";
 import { getIsDomainInURL } from "lib/shared/getIsDomainInURL";
+import { User } from "lib/types/carbonmark.types";
 import { GetStaticProps } from "next";
 import { ParsedUrlQuery } from "querystring";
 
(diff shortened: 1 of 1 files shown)


biwano:
@0xemc @Atmosfearful 
The issue was due to a change to the client.ts file: https://github.com/KlimaDAO/klimadao/blame/8ce464eb32804749bb760d43a5aae15e25f3b263/carbonmark/lib/api/client.ts
The client now throws exceptions when the response from the API is invalid. Functions must catch those errors.

I note that it is a temporary solution. So we should be careful when this changes again.

jabby09 (approved):

PR #1926, opened 2023-11-24
Makka / #1132 - replace browser injected wallet provider icon

### LINKED ISSUE #1132: [connect modal] Fix default wallet provider display on login modal
image

When user is using Frame or similar non-standard browser-based wallets, we should have a cleaner default logo with the right sizing so it's consistent with the other options

Most other sites use the 🤑 emoji 
image

0xMakka:
## Description

Replaces the default browser injected wallet icon on the login modal.

## Related Ticket

Resolves #1132 

## Changes

| Before  | After  |
|---------|--------|
|Screenshot 2023-11-24 at 15 05 10|Screenshot 2023-11-24 at 15 04 53|

## Checklist


lib/components/ConnectModal/index.tsx
```diff
@@ -1,6 +1,6 @@
 import { cx } from "@emotion/css";
+import AccountBalanceWalletIconDefault from "@mui/icons-material/AccountBalanceWallet";
```
sprrwhwk, line 2:
```suggestion
import AccountBalanceWalletIcon from '@mui/icons-material/AccountBalanceWallet';
```

Should do the trick, no?

0xMakka: @sprrwhwk There is two other icon imports in this file `Close` and `MailOutline`, the reason I added the "Default" text to the end of the string was to be consistent with the naming of the other two imports, and below due to esm import issues, default needs to be called on the icons after importing.

```
import MailOutlineIconDefault from "@mui/icons-material/MailOutline";
import AccountBalanceWalletIconDefault from "@mui/icons-material/AccountBalanceWallet";

// ems modules and javascript are strange so we import like this
const MailOutlineIcon = (MailOutlineIconDefault as any).default as any;
const AccountBalanceWalletIcon = (AccountBalanceWalletIconDefault as any)
  .default as any;
```

biwano (approved):

PR #1929, opened 2023-11-27
Solves inconsistencies in API return values

### LINKED ISSUE #1013: [performance] Convert BigNumbers further up the chain
## Description

From a PR comment:

> Not a change request just something to keep in mind: would be nice to convert these BigNumbers further up the chain if possible, which is what I do in the app as soon as they are fetched before setting to redux state. So components only ever deal with stringified numbers.
> 
> Ideally the backend just gives us strings, we can ask them.
> 
> Unless of course we need to do actual bignumber math, in which case we need to keep them as bignumbers.

_Originally posted by @Atmosfearful in https://github.com/Atmosfearful/bezos-frontend/pull/48#discussion_r1083195081_
            

## Current state

At the moment, all pages fetch their data directly yon the server from the API.
The numbers in the API data are BigBumbers.
These are passed down to the page, which is currently converting all BigNumbers to numbers.

See all types here:
https://github.com/Atmosfearful/bezos-frontend/blob/bezos-main/lib/types/carbonmark.ts

biwano:
## Description

Makes all timestamps returned as seconds since epoch (numbers)

See this comment for the rationale behind those changes (to be discussed): https://github.com/KlimaDAO/klimadao/issues/1013#issuecomment-1827162622

## Related Ticket

Resolves #1013

## Checklist


carbonmark-api/src/utils/helpers/users.utils.ts
```diff
@@ -78,5 +84,21 @@ export const getProfileByHandle = async (params: {
...
+export function getUiUserProfile(profile: UserProfile): UiUserProfile {
+  return {
+    address: profile.address,
+    description: profile.description,
+    handle: profile.handle,
+    username: profile.username,
+    createdAt: (profile.createdAt / 1000).toFixed(0),
+    updatedAt: (profile.updatedAt / 1000).toFixed(0),
+  };
+}
```
0xemc, line 104: Is the main purpose of this function to format the createdAt and updatedAt timestamps? 

If so it might be better to do this explicitly as a seperate utility at the point of render. This will also avoid the need for another manual type that needs to be maintained`UiUserProfile`.

Unless there's another benefit I'm missing here?

biwano: > Is the main purpose of this function to format the createdAt and updatedAt timestamps?
Yup
> 
> If so it might be better to do this explicitly as a seperate utility at the point of render. This will also avoid the need for another manual type that needs to be maintained`UiUserProfile`.
> 
> Unless there's another benefit I'm missing here?

The thing is that user model transformation happens also in the projects route so I would need a model transformation function anyways.
`get.ts -> fetchmarketplacelisting.ts -> users.utils.ts`

And given how the existing code is structured, it would feel awkward for that route, the handler should not have to know that this particular piece of information needs conversion.

0xemc: Yah I agree I guess my point is that the UI (App or consumer of the API) should be responsible for any formatting. 

The API should aim to be client agnostic. Do you know what I mean?

0xemc (changes requested):
We'll need a version bump on the `package.json` too btw

carbonmark-api/src/utils/helpers/users.utils.ts
biwano: I was not working on carbonmark when the project started so I do not have the history. But I noticed that timestamps are converted to strings on purpose (same for prices) so I wanted to stick with that original design.
But on principle I completly agree with you with returning timestamps as numbers and will modify the PR in that direction.
I tag @Atmosfearful just in case there is a real rationale behind this conversion for timestamps.

Atmosfearful (approved):
Looks good to me, appreciate the cleanup! Will let @0xemc do final review and merge

biwano:
@O

> Looks good to me, appreciate the cleanup! Will let @0xemc do final review and merge

@0xemc If you can take a look.

carbonmark-api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@klimadao/carbonmark-api",
-  "version": "2.0.3",
+  "version": "4.0.0",
```
0xemc, line 3: This might be the wrong version bump depending on which PR is merged first.. If you intend to wait on this #1984 then this is ok.

carbonmark-api/src/utils/helpers/utils.ts
```diff
@@ -228,3 +228,14 @@ export const isMatchingCmsProject = (
...
   project: CarbonProject
 ) => project?.registryProjectId === projectId && project.registry === registry;
+
+export function formatGraphTimestamps<
+  T extends { createdAt: string | null; updatedAt: string | null },
+>(data: T) {
+  const partialData = omit(data, ["createdAt", "updatedAt"]);
+  return {
+    ...partialData,
+    createdAt: Number(data.createdAt),
```
0xemc, line 238: Are the graph time stamps not in milliseconds? i.e we don't need to divide by 1000

If so can you put a short comment in to mention this?

```suggestion
/** Note: Graph timestamps are in seconds **
    createdAt: Number(data.createdAt),
```

carbonmark-api/test/routes/projects/get.test.ts
```diff
@@ -200,16 +200,18 @@ describe("GET /projects", () => {
...
+            ...pick(marketplace.projectWithListing.listings![0], [
+              "active",
+              "batchPrices",
+              "batches",
+              "deleted",
+              "id",
+              "tokenAddress",
+            ]),
+            createdAt: 1234,
+            updatedAt: 1234,
```
0xemc, line 213:
```suggestion
              "tokenAddress",
              "createdAt,
              "updatedAt"
            ]),

```

0xemc (approved):

0xemc:
> > Looks good to me, appreciate the cleanup! Will let @0xemc do final review and merge
> 
> @0xemc If you can take a look.

Looks good to me, merge away

carbonmark-api/test/routes/projects/get.test.ts
biwano: This would not work since we are converting createdAt and updatedAt to numbers.

PR #1933, opened 2023-11-28
Emc/fix-request-params-on-kubb-hooks

0xemc:
## Description
Resolves:

1. Issue caused by incorrect formatting of params in `lib/api/client.ts`. This is the client that the SDK uses to fetch and params were being parsed as `[Object]` or `undefined`
2. Uses `shouldFetch` on the swr `use` hook to ensure we only fetch when an address for the user has been provided. This will stop a lot of silent errors in the console.
3. Includes `ConditionalErrorBoundary` which allows a component to catch an error (404 in the case of an unknown user) and still render the child. (client side)
4. The [user] route should now allow 404 exceptions when initially rendering (serverside)
5. Project filters not working due to swr hooks not taking params as part of their key

FYI the 500 and 404 errors are a result of the app being strict on 404 errors since the sdk upgrade. This PR will enable those errors to be triggered without crashing the app.


0xemc:
@jabby09 this PR could use your QA. Things to check:

- [ ] project filtering and searching in `/projects`
- [ ] Log in and account creation with new address
- [ ] Profile Editing (change username etc)

Thank you 🙏

carbonmark/components/pages/Projects/ProjectsController/index.tsx
```diff
@@ -7,22 +6,32 @@ import { LoginButton } from "components/LoginButton";
...
+  value === "" ? undefined : value;
 const ProjectsController = () => {
   const { isDesktop } = useResponsive();
   const router = useRouter();
   const { params, updateQueryParams } = useProjectsParams();
   const [showFilterModal, setShowFilterModal] = useState(false);
-  const { data: projects = [] } = useGetProjects();
+
+  /** convert all string arrays to comma separated string as per the api's expectation */
+  const mappedParams = mapValues(params, pipe(joinArray, emptyToUndefined));
```
biwano, line 32: I think we should make the API expect html form post style query parameters because fastify uses nodejs queryparser which natively expect this.
ie. 
`?category=Energy+Efficiency&category=Forestry`
instead of:
`?category=Energy+Efficiency%2CForestry`

but maybe later

carbonmark/hocs/ConditionalErrorBoundary/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+    >
+      <Component {...props} />
+    </ConditionalErrorBoundary>
+  );
+
+  /** Add another display name for the wrapped version */
+  ComponentWithConditionalErrorBoundary.displayName = `withConditionalErrorBoundary(${displayName})`;
+
+  return ComponentWithConditionalErrorBoundary;
+}
```
biwano, line 63: Real nice!

biwano (approved):

jabby09:
> this PR could use your QA

@0xemc going to hand this one off to @0xtapi

carbonmark/pages/users/[user].tsx
```diff
@@ -70,7 +83,16 @@ const resolveDomain = async (params: { domain: string; locale?: string }) => {
  * Attempts to resolve a valid user handle. Throws if handle can't be resolved.
  * */
 const resolveHandle = async (params: { handle: string; locale?: string }) => {
-  const carbonmarkUser = await getUsersWalletorhandle(params.handle);
+  let carbonmarkUser: User | null = null;
+  try {
+    carbonmarkUser = await getUsersWalletorhandle(params.handle);
```
Comment (0xMakka) [inline carbonmark/pages/users/[user].tsx:88]: Nit: might be a good chance to rename this?

```suggestion
    carbonmarkUser = await getUsersWalletOrHandle(params.handle);
```

0xMakka (approved):

carbonmark/pages/users/[user].tsx
Comment (0xemc) [reply carbonmark/pages/users/[user].tsx:88]: Unfortunately this one is generated from the API spec too. Will see what we can do.

carbonmark/components/pages/Projects/index.tsx
```diff
@@ -28,12 +30,25 @@ const views = {
   map: LazyLoadingMapView,
 };
 
+const joinArray = (value: string | string[]): string =>
```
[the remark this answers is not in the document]
0xemc: Hoping that this solution will be temporary.. fingers crossed

PR #1943, opened 2023-12-01
carbonmark: Fix project's page hydratation issue

biwano:
## Description

Fixes the hydratation issue on the `/projects` page when there are no query parameters

## Checklist


Atmosfearful:
Was there a ticket for this @biwano ?

biwano:
> Was there a ticket for this @biwano ?

No, I just created one: https://github.com/KlimaDAO/klimadao/issues/1959

biwano:
@Atmosfearful I went for rendering server side at every request. 
I made a few more changes to make the best of that situation:
 - Use query parameters server side
 - Displays projects on first render
 - Only make a request on the client when a query parameter is changed by the user

@0xemc If you have some time to take a look too,

If that's good:
 - We will need a QA pass
 - maybe we can extract a pattern for other pages.

carbonmark/pages/projects/index.tsx
```diff
@@ -22,17 +22,21 @@ export interface ProjectsPageStaticProps {
...
-    const translation = await loadTranslation(ctx.locale);
+    const [projects, vintages, countries, categories, translation] =
+      await Promise.all([
+        fetchProjects(ctx.query),
+        getVintages(),
+        getCountries(),
+        /** @note because the API is returning trailing empty spaces on some categories, trim them here */
+        getCategories().then((categories) =>
+          categories.map(update("id", trim))
+        ),
```
Atmosfearful, line 37: We can remove the comment, empty strings are no longer a problem.
This can become
```
.then(({ id }) => id)
```

Atmosfearful (changes requested):

Issue #1944, opened 2023-12-01
[profile | portfolio] Screen flashing and content regressions on Profile & Portfolio pages


jabby09:
Related issues:
#1940 
#1063 

This PR https://github.com/KlimaDAO/klimadao/pull/1942 was introduced to fix profile bugs in #1940

This appears to have introduced a new set of issues:

- [x] When I access another users public profile (for example https://www.carbonmark.com/users/offsetra) the initial screen view flashes the Listings - No active listings page before rendering properly

- [x] A new login with no Profile created should still show the temp avatar and wallet and stats / activity components. The figma design for a seller that has logged in but not yet set up their profile is [here](https://www.figma.com/file/tnBViWV19Ln9jjFDekksY7/Profile?type=design&node-id=13-13294&mode=design&t=g8WQBIgLuA3J3cXB-4); assuming the Marketplace navigation item is still active; i.e. start from the marketplace, click login, go through the login modal, end up on this page.

<img width="754" alt="Screenshot 2023-12-01 at 9 32 44 AM" src="https://github.com/KlimaDAO/klimadao/assets/103535195/7813b642-184e-471f-a419-d95457e04cbd">
vs 
<img width="705" alt="Screenshot 2023-12-01 at 9 33 27 AM" src="https://github.com/KlimaDAO/klimadao/assets/103535195/0c6a4c8d-ddcc-4341-80bd-b16d43acc941">

- [x] Recent portfolio page content changes that were merged as part of #1063 not longer seem to be there anymore. Currently if you do not have a profile, the Portfolio page is completely blank.

Atmosfearful:
> Using social logins, if I've logged in and then log out, if I login again in the same browser session with a different social login (for example google instead of discord) the web3 auth appears to send me to the correct location but I still get connected to carbonmark with the previous user (in this case, my discord user). Basically it appears that within a browser session, if I log out if does not fully release the account I was logged in as.

I think this one should be moved to a separate ticket, I think we can merge michaels PR to resolve the other more pressing issues

jabby09:
> I think this one should be moved to a separate ticket, I think we can merge michaels PR to resolve the other more pressing issues

done

Issue #1063, opened 2023-04-20
[portfolio] Default page content changes


jabby09:
When a new user logs in, and they have not yet completed their seller profile, and they also have no carbon token assets in their wallet, this is the current UX...
<img width="1106" alt="Screenshot 2023-11-24 at 9 23 08 AM" src="https://github.com/KlimaDAO/klimadao/assets/103535195/7545b91c-8c36-47ee-a02b-1f4b2f1eedc3">

- [ ] Change the page copy to:

> This is your portfolio, here you will find any of the supported carbon assets you own. Assets shown here can be retired or listed for sale.
>
> You do not currently have any supported carbon assets available for listing.
>  
> To get started, be sure to complete your link:[profile] and then head to the link:[marketplace] to purchase carbon assets.

jabby09:
@0xMakka this is another quick but nice UX enhancement

PR #1945, opened 2023-12-01
Emc/fix/1944-profile-portfolio-issues

0xemc:
## Description

* Replace a `=== null` check that was failing because of `undefined`
* Pass the User object directly from getStaticProps rather than through SWRConfig

Resolves the following from #1944:
- [x] When I access another users public profile (for example https://www.carbonmark.com/users/offsetra) the initial screen view flashes the Listings - No active listings page before rendering properly

`JABBY CONFIRMS FIXED`

- [x] Recent portfolio page content changes that were merged as part of https://github.com/KlimaDAO/klimadao/issues/1063 not longer seem to be there anymore. Currently if you do not have a profile, the Portfolio page is completely blank.

`JABBY CONFIRMS FIXED`

- [x] A new login with no Profile created should still show the temp avatar and wallet and stats / activity components. The figma design for a seller that has logged in but not yet set up their profile is [here](https://www.figma.com/file/tnBViWV19Ln9jjFDekksY7/Profile?type=design&node-id=13-13294&mode=design&t=g8WQBIgLuA3J3cXB-4); assuming the Marketplace navigation item is still active; i.e. start from the marketplace, click login, go through the login modal, end up on this page.

`JABBY CONFIRMS FIXED`

### DIFF
--- a/carbonmark-api/src/models/User.model.ts
+++ b/carbonmark-api/src/models/User.model.ts
@@ -1,4 +1,5 @@
 import { Static, Type } from "@sinclair/typebox";
+import { isNil } from "lodash";
 import { ActivityModel } from "./Activity.model";
 import { AssetModel } from "./Asset.model";
 import { ListingModel } from "./Listing.model";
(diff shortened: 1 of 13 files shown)


jabby09 (commented):
@0xemc which two issues does this PR fix?

0xemc:
> @0xemc which two issues does this PR fix?

@jabby09 I'll be adding them in the description as I go. If you'd like to confirm on the preview link that'd be good or wait until they're all finshed

0xemc:
@jabby if you're ok with these fixes we can close this PR and tackle the others after. 

I need to clarify the logged in user without an account requirement too btw

CC: @Atmosfearful

jabby09:
> if you're ok with these fixes we can close this PR and tackle the others after.

@0xemc checking now

PR #1952, opened 2023-12-02
fix:1951 error.status -> error.data.statusCode

0xemc:
## Description

We were incorrectly accessing the error status code with the new kubb hooks


0xtapi:
Tested, played around with it a bit, couldn't find any bugs. Ship it!

PR #1962, opened 2023-12-04
Makka / #1953 content changes for unregistered message

### LINKED ISSUE #1953: [portfolio] Content change for users without a CM profile
Currently this is the message we show on a Portfolio page if the logged in user does NOT have a CM profile.

image

Please change the content to the following:

> This is your portfolio. Here you will find supported carbon assets you own. Assets shown here can be retired or listed for sale in the marketplace once you have a profile.
> 
> You must have a profile created to view carbon assets you own in your Carbonmark portfolio.
> 
> To get started, be sure to complete your [profile](https://www.carbonmark.com/users/0xFA37c2042f470517f15172a239723F3eb8f1Bd6A) and then head to the [marketplace](https://www.carbonmark.com/projects) to purchase carbon assets.

Note...the profile link should obviously be constructed to use the wallet of the connected user

0xMakka:
## Description

This PR makes a change to the content shown to logged in users that do not have a CM profile.

## Related Ticket

Closes #1953 

## Notes For QA
* [ ] Verify that the message reflects the requirements outlined in #1953, also check:
       - When clicking on the `profile` link in the content, it should use the wallet address of the connected user.
       - When clicking on the `marketplace` link in the content, it should bring the user back the marketplace home page.

- Portfolio View

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
   },
   "devDependencies": {
     "@lingui/cli": "^4.4.0",
+    "@lingui/macro": "^4.6.0",
     "concurrently": "^7.6.0",
     "prettier": "3.0.2",
     "prettier-plugin-organize-imports": "^3.2.3"
--- a/package.json
+++ b/package.json
@@ -50,7 +50,6 @@
   },
   "devDependencies": {
     "@lingui/cli": "^4.4.0",
-    "@lingui/macro": "^3.17.2",
     "concurrently": "^7.6.0",
     "prettier": "3.0.2",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -50,6 +50,7 @@
   },
   "devDependencies": {
     "@lingui/cli": "^4.4.0",
+    "@lingui/macro": "^4.6.0",
```
biwano, line 53: I think you did not want to change the version of @lingui/macro in this PR.

biwano (changes requested):

PR #2012, opened 2023-12-11
Simplify profile auth & signature flow

Atmosfearful:
## Description

As we prepare for wallet abstraction (web2 auth) this old JWT verification method should be deprecated.

Switched to a more straightforward signature + nonce approach for every request.

### Old way
/verify was used to verify a users signature and mint a JWT for the session
The JWT was included in the edit requests
### New way
POST and PUT requests expect an `Authorization` header which should include a signed message.
The string to be signed is included as a vercel env var, and will be shared with the frontend

# Note
The frontend needs to be updated to consume v3.0.0 and pass the signature to the header.

# QA
User with a nonce:
https://carbonmark-api-git-atmos-simplify-auth-klimadao.vercel.app/users/atmosfearful

User without a nonce (backwards compat)
https://carbonmark-api-git-atmos-simplify-auth-klimadao.vercel.app/users/jabby


carbonmark-api/src/routes/users/put.ts
```diff
@@ -11,12 +13,32 @@ const handler = (fastify: FastifyInstance) =>
...
+      });
+      if (!isAuthenticated) {
+        return reply.status(403).send();
+      }
+
       const updatedData = {
-        username,
-        description,
+        username: username || userDoc.username || userDoc.handle,
+        description: description || userDoc.description || null,
```
biwano, line 38: Description is optionalin the UI. 
I think that with this code we discard "" as a value for description, which prevents the users from deleting the description. (I did not test)

carbonmark-api/src/routes/users/put.ts
```diff
@@ -11,12 +13,32 @@ const handler = (fastify: FastifyInstance) =>
...
         updatedAt: Date.now(),
-        profileImgUrl:
-          profileImgUrl && profileImgUrl.length ? profileImgUrl : null,
+        profileImgUrl: profileImgUrl || userDoc.profileImgUrl || null,
```
biwano, line 40: Same

carbonmark-api/src/utils/firebase.utils.ts
```diff
@@ -1,4 +1,41 @@
...
+  /** Editable username */
+  username: string;
+  /** Optional profile description */
+  description?: string | null;
+  /** Optional image url */
+  profileImgUrl?: string | null;
+  /** Nonce, incremented once per edit, may not be present
+   * Ensures the same message hash can never be reused (replay attack) */
+  nonce?: number;
+}
```
biwano, line 27: It would be nice if this was a model (like Asset.model.ts).
In my carbon provenance PR. I needed to return user information (in particular profileImgUrl) and I ended up creating a UserProfile model which is in fact your FirestoreUserDoc interface.

carbonmark-api/src/utils/firebase.utils.ts
```diff
@@ -1,4 +1,41 @@
...
+}): Promise<null | FirestoreUserDoc> => {
+  const userDocRef = await params.firestore
+    .collection("users")
+    .doc(params.docId.toUpperCase())
+    .get();
+  if (!userDocRef.exists) return null;
+  const userDoc = userDocRef.data();
+  if (!isFirestoreUserDoc(userDoc)) return null;
+  return userDoc;
+};
```
biwano, line 41: There is a users.utils.ts where there are functions to fetch user information from firebase. We should regroup them.

biwano (changes requested):

carbonmark-api/src/utils/crypto.utils.ts
```diff
@@ -25,3 +20,24 @@ export const formatUSDC = (n: bigint | string): string => {
...
+ * Uses ECDSA under the hood to recover the signerWalletAddress
+ */
+export const verifyProfileSignature = (params: {
+  nonce?: number;
+  signature: string;
+  expectedAddress: string;
+}) => {
+  if (!params.signature) return false;
+  // Backwards-compat: nonce may be undefined, append empty string
+  const expectedMessage = SIGN_PROFILE_MESSAGE + `${params?.nonce}`;
```
sprrwhwk, line 35: Wouldn't this append `null` or `undefined`?
```suggestion
  // Backwards-compat: nonce may be undefined, append empty string
  const expectedMessage = SIGN_PROFILE_MESSAGE + params?.nonce ?? "";
```

carbonmark-api/src/routes/users/post.ts
```diff
@@ -44,11 +47,23 @@ const handler = (fastify: FastifyInstance) =>
...
+      // first signature is without nonce
+      expectedAddress: createData.address,
+      signature: request.headers.authorization?.split(" ")[1] || "",
+    });
+
+    if (!isAuthenticated) {
+      return reply.status(403).send({
+        error: "Unauthorized: invalid signature.",
+      });
+    }
```
0xemc, line 65: This might be better handled in some middleware like the old `bearer.ts`, what do you think?

I'm all for not being religious about duplicated code but having the same logic like this in seperate files leaves the possibility open that the logic is changed in one place but not in all cases leading to lots of headaches.

carbonmark-api/src/routes/users/put.ts
```diff
@@ -11,12 +13,32 @@ const handler = (fastify: FastifyInstance) =>
...
+      }
+
+      const isAuthenticated = verifyProfileSignature({
+        nonce: userDoc.nonce,
+        expectedAddress: userDoc.address,
+        signature: request.headers.authorization?.split(" ")[1] || "",
+      });
+      if (!isAuthenticated) {
+        return reply.status(403).send();
+      }
```
0xemc, line 34: As above, express plugins are designed for this I think.

carbonmark-api/src/app.constants.ts
```diff
@@ -56,3 +56,6 @@ export const TOKEN_ADDRESSES = {
 export const RPC_URLS = {
   polygonTestnetRpc: "https://rpc-mumbai.maticvigil.com",
 };
+
+/** Message shared with frontend, to be combined with user's nonce and signed by private key. */
+export const SIGN_PROFILE_MESSAGE = `${process.env.SIGN_PROFILE_MESSAGE}`;
```
0xemc, line 61: Is this different from the current `AUTHENTICATION_MESSAGE` in vercel? 

image

carbonmark-api/src/models/User.model.ts
```diff
@@ -17,6 +17,7 @@ export const UserModel = Type.Object({
   listings: Type.Array(ListingModel),
   activities: Type.Array(ActivityModel),
   assets: Type.Array(AssetModel),
+  nonce: Type.Optional(Type.Number()),
```
0xemc, line 20: I'm probably misunderstanding something. Isn't the nonce only supposed to be valid for a single auth attempt to prevent MITM attacks?

carbonmark-api/src/routes/users/post.schema.ts
```diff
@@ -12,7 +12,10 @@ export const RequestBody = Type.Object(
   { required: ["handle", "username", "wallet", "description"] }
 );
 
-export type CreateUserResponse = Static<typeof RequestBody>;
+export const ResponseBody = Type.Object({
+  address: Type.String({ minLength: 26, maxLength: 64 }),
```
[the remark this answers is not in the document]
Atmosfearful: good call, added a regex for 0x addresses and consumed that directly in the schema

carbonmark-api/src/routes/users/put.ts
Atmosfearful: Great catch, I also caught a mismatch in the schema which was disallowing empty string. I'll see if I can add a testcase for this.

carbonmark-api/src/utils/firebase.utils.ts
Atmosfearful: good call

carbonmark-api/src/utils/crypto.utils.ts
Atmosfearful: You are right! Will add a test and fix this

carbonmark-api/src/app.constants.ts
Atmosfearful: renamed to make it specific for this operation (profile puts/post)

carbonmark-api/src/models/User.model.ts
Atmosfearful: Correct, it's included here because the client needs to know the correct nonce to sign. It's valid for a single successful PUT or POST and can never be reused (because it increases monotonically).

Previously we used random nonces which theoretically could reappear and thus be reused (to repeat an edit, so not really a problem but still technically worse)

carbonmark-api/src/routes/users/post.ts
Atmosfearful: that... took wayyyyy longer than I care to admit... but I got it working!

Atmosfearful:
Oh no... tests pass locally but not remotely... nightmare scenario. Will look deeper tmr

0xemc:
> Oh no... tests pass locally but not remotely... nightmare scenario. Will look deeper tmr

Looks like it's failing for me on local too: `test/routes/users/post.test.ts` & `test/routes/users/put.test.ts`

carbonmark-api/test/test.utils.ts
```diff
@@ -1,39 +1,49 @@
...
-//Renable the `bearer.ts` plugin
-export function enableAuth() {
-  delete process.env.IGNORE_AUTH;
+  // allows us to keep invoking `mockFirestore()` in our tests to apply new mock implementations
+  if (overrides) {
+    firestore.mockImplementation(() => ({
+      ...firestoreMethods,
+      ...overrides,
+    }));
+  }
```
0xemc, line 48: Nice 👍

carbonmark-api/src/utils/crypto.utils.ts
```diff
@@ -25,3 +20,24 @@ export const formatUSDC = (n: bigint | string): string => {
...
+ * Uses ECDSA under the hood to recover the signerWalletAddress
+ */
+export const verifyProfileSignature = (params: {
+  nonce?: number;
+  signature: string;
+  expectedAddress: string;
+}) => {
+  if (!params.signature) return false;
+  // Backwards-compat: nonce may be undefined, append empty string
+  const expectedMessage = SIGN_PROFILE_MESSAGE + (params?.nonce || "");
```
0xemc, line 35: Is it possible that `params?.nonce` is also 0 (initially?) causing the falsey operator to return ""?

I'm always cautious of the `||` operator because of these bugs

0xemc (approved):
Ticked assuming the failing unit tests don't reveal anything nasty in the logic.

0xemc:
Converting to draft while tests are being fixed

Atmosfearful:
Tests pass! Getting this on staging so we can fix an edit profile issue (nonce is missing on recent API deployments)

PR #2023, opened 2023-12-13
[FE] Makka / #1992 pay with bank transfer FE changes

0xMakka:
## Description

This PR adds the ability to Retire via bank transfer, more detailed descriptions are available in the main ticket - #1992  

## Related Ticket

Closes #1992
Also related to & relies on #2024 

## Notes For QA

QA is required as this PR adds a new feature to pay with bank transfer on retirement pools.
* [x] Allow a user to choose Bank transfer from the `Pay with` list.
* [x] `How many tonnes of carbon would you like to retire?` is the only required field at this stage - the "Retire Carbon" button should be disabled if nothing is entered in this field. The other fields are optional. If nothing is entered into the Beneficiary wallet address field, the connected users wallet address will be used.
* [x] After entering a quantity and clicking "Retire Carbon", the user should be presented with a modal popup according to the figma designs where they can "Continue" or "Go back". 
      - Upon clicking "Go back" the modal should be closed.
      - Upon clicking "Continue", the user should be brought to the "Pay with Bank Transfer" form where they are presented with a more detailed form with both required and optional fields. Any details entered on the previous "Retire Carbon" form should be carried over and filled in on this form.
* [x] On the "Pay with Bank Transfer" form the "Submit" button should be disabled if the required fields aren't filled out (denoted with an * on the label). As the user types in the input fields, error messages should be shown if the data is incorrect on a required field.
* [x] Upon submitting the form, a POST request is made to the carbonmark api at the following endpoint `/retire/bank-transfer` with the form data in the body of the request. The api then makes a call to the hubspot form api with the form data. 
     - If the request is successful, a 200 response will be returned and the user will be shown the "Form complete" screen with the previously entered data filled out.
     - If the request is unsuccessful, a 400 response will be returned and an error message will be displayed above the "Submit" button.

Specific pages, components or journeys that might be affected:
- Pool Retirements Page

PR #2024, opened 2023-12-13
[API] Makka/1992 pay with bank transfer api

0xMakka:
## Description

This PR adds a new api route which handles sending the "pay with bank transfer" form data to the hubspot api.

Note: The FE in #2023 relies on this branch in order to be able to submit details.

Todo:
- [x] Write tests for the new api route.

## Related Ticket

Related to #2023

### DIFF
--- a/carbonmark-api/postman_collection.json
+++ b/carbonmark-api/postman_collection.json
@@ -550,6 +550,48 @@
         }
       },
       "response": []
+    },
+    {
+      "name": "retireViaBankTransform",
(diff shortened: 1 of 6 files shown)


0xMakka:
This PR is no longer required. Based on @Atmosfearful's feedback in [#2023](https://github.com/KlimaDAO/klimadao/pull/2023), it makes sense that the api logic belongs in the nextjs api and not the public api.

Issue #1936, opened 2023-11-29
[retire] Carbon provenance implementation


jabby09:
### Related docs & links

[Shipped Figma Design](https://www.figma.com/file/7N5csmr1NPeKefGh5MuGeh/Transfer-History?type=design&node-id=1-90&mode=design&t=rDcTnXPNjtix3jl0-0)

Related subgraph work: https://github.com/KlimaDAO/klima-subgraph/pull/75

[Design/ discovery discord thread ](https://discord.com/channels/1135827610434289727/1173629852050526289)

### Requirements & user journey

- In the figma design file, find the DEVS START HERE section
- Note the use of the material-ui react timeline component (https://mui.com/material-ui/react-timeline/)
- User journey starts from either a retirement receipt page or retirement receipt pdf
- Also not the any design differences for desktop vs mobile (included in file)

**Retirement Receipt Page**

- Remove the COPY LINK button in the current implementation
<img width="685" alt="Screenshot 2023-11-29 at 2 37 31 PM" src="https://github.com/KlimaDAO/klimadao/assets/103535195/88fbd273-2e6f-459d-bc5c-b2e00f6de472">

- As per the new [design](https://www.figma.com/file/7N5csmr1NPeKefGh5MuGeh/Transfer-History?type=design&node-id=18-1037&mode=design&t=rDcTnXPNjtix3jl0-4), move `VIEW ON POLYGONSCAN` button to the left and add a `CARBON PROVENANCE` button
<img width="694" alt="Screenshot 2023-11-29 at 2 40 38 PM" src="https://github.com/KlimaDAO/klimadao/assets/103535195/f68401ab-3ce2-4f06-95e8-629be870652e">

**PDF Receipt Page**
- As per the new [design](https://www.figma.com/file/7N5csmr1NPeKefGh5MuGeh/Transfer-History?type=design&node-id=18-1131&mode=design&t=rDcTnXPNjtix3jl0-4), add a View carbon provenance link to the exported pdf
<img width="616" alt="Screenshot 2023-11-29 at 2 43 30 PM" src="https://github.com/KlimaDAO/klimadao/assets/103535195/90a19c82-5c6a-40f2-9618-b26b9cfe6bd4">

**Carbon Provenance Page (NEW)**
- This is a new page
[...]

PR #2034, opened 2023-12-15
[Provenance API] Adds retirements routes

biwano:
## Description

Adds retirement routes in the API
- Get a retirement
- Get a retirement provenance records
- Search retirments (only via account and retirement_index for now)

## Related Ticket

Also related to #1936


0xemc (changes requested):
I'm not sure I agree with the very specific routes. A single handler on `/retirements` would be more flexible and long term sustainable. I can't imagine having to create new handlers & routes and tests every time we use a different provider or retirement type.

Can we please handle this with a single handler and request params or the body of a POST request instead?

biwano:
> I'm not sure I agree with the very specific routes. A single handler on `/retirements` would be more flexible and long term sustainable. I can't imagine having to create new handlers & routes and tests every time we use a different provider or retirement type.
> 
> Can we please handle this with a single handler and request params or the body of a POST request instead?

Agreed on the fact that the name was very poorly chosen. my purpose was not to have a route by provider. But as an API user I would not like to use request parameters for a route that returns a single retirement. I would like to keep that for a route that filters retirements based on multiple parameters.
So I changed the route to just delete the klima part of the path. The idea is that when an API user wants to get a single retirement he passes the account_id and the retirement_index as part of the path. This would mean that we would consider the couple (account_id, retirement_index) the default `key` for a retirement.
LMKWYT

0xemc:
> > I'm not sure I agree with the very specific routes. A single handler on `/retirements` would be more flexible and long term sustainable. I can't imagine having to create new handlers & routes and tests every time we use a different provider or retirement type.
> > Can we please handle this with a single handler and request params or the body of a POST request instead?
> 
> Agreed on the fact that the name was very poorly chosen. my purpose was not to have a route by provider. But as an API user I would not like to use request parameters for a route that returns a single retirement. I would like to keep that for a route that filters retirements based on multiple parameters. So I changed the route to just delete the klima part of the path. The idea is that when an API user wants to get a single retirement he passes the account_id and the retirement_index as part of the path. This would mean that we would consider the couple (account_id, retirement_index) the default `key` for a retirement. LMKWYT

Hmm it seems like a lot of extra complexity to avoid the user having to extract the first element from an array. If we were to provide a `/retirements` endpoint would that require and additional handler?

 I would much prefer that we have a single point of access for Retirements (e.g `/retirements`) and the user has the flexibility to reduce their results how they wish.
 
 Unless theres another reason that a single result without params or a POST body is preferable to developers accessing our API?

biwano:
> > > I'm not sure I agree with the very specific routes. A single handler on `/retirements` would be more flexible and long term sustainable. I can't imagine having to create new handlers & routes and tests every time we use a different provider or retirement type.
> > > Can we please handle this with a single handler and request params or the body of a POST request instead?
> > 
> > 
> > Agreed on the fact that the name was very poorly chosen. my purpose was not to have a route by provider. But as an API user I would not like to use request parameters for a route that returns a single retirement. I would like to keep that for a route that filters retirements based on multiple parameters. So I changed the route to just delete the klima part of the path. The idea is that when an API user wants to get a single retirement he passes the account_id and the retirement_index as part of the path. This would mean that we would consider the couple (account_id, retirement_index) the default `key` for a retirement. LMKWYT
> 
> Hmm it seems like a lot of extra complexity to avoid the user having to extract the first element from an array. If we were to provide a `/retirements` endpoint would that require and additional handler?
> 
> I would much prefer that we have a single point of access for Retirements (e.g `/retirements`) and the user has the flexibility to reduce their results how they wish.
> 
> Unless theres another reason that a single result without params or a POST body is preferable to developers accessing our API?

It would work with the single retirements handler. But I think it is good have this pattern where as an API user you can filter instances of a model on one side, and you can get an instance of a model by id on the other side. 
It enables you to deal with instance of models instead arrays of instance, Instead of checking that you have a single result in your response array. You are ensured to have an instance of a model or a 404 error.
Plus, this is what we do already for projects:
- `/projects` => Filter projects
- `/projects/:id` => Get single projects
----------
- `/retirements` => Get multiple retirements
- `/retirements/:id` => Get single retirement (where `:id` is `/:account_id/:retirement_index`

Seems like a common pattern to me when building APIs.
The `id` could be the transactionHash instead of `/:account_id/:retirement_index` though

0xemc:
> > > > I'm not sure I agree with the very specific routes. A single handler on `/retirements` would be more flexible and long term sustainable. I can't imagine having to create new handlers & routes and tests every time we use a different provider or retirement type.
> > > > Can we please handle this with a single handler and request params or the body of a POST request instead?
> > > 
> > > 
> > > Agreed on the fact that the name was very poorly chosen. my purpose was not to have a route by provider. But as an API user I would not like to use request parameters for a route that returns a single retirement. I would like to keep that for a route that filters retirements based on multiple parameters. So I changed the route to just delete the klima part of the path. The idea is that when an API user wants to get a single retirement he passes the account_id and the retirement_index as part of the path. This would mean that we would consider the couple (account_id, retirement_index) the default `key` for a retirement. LMKWYT
> > 
> > 
> > Hmm it seems like a lot of extra complexity to avoid the user having to extract the first element from an array. If we were to provide a `/retirements` endpoint would that require and additional handler?
> > I would much prefer that we have a single point of access for Retirements (e.g `/retirements`) and the user has the flexibility to reduce their results how they wish.
> > Unless theres another reason that a single result without params or a POST body is preferable to developers accessing our API?
> 
> It would work with the single retirements handler. But I think it is good have this pattern where as an API user you can filter instances of a model on one side, and you can get an instance of a model by id on the other side. It enables you to deal with instance of models instead arrays of instance, Instead of checking that you have a single result in your response array. You are ensured to have an instance of a model or a 404 error. Plus, this is what we do already for projects:
> 
> * `/projects` => Filter projects
> * `/projects/:id` => Get single projects
> 
> * `/retirements` => Get multiple retirements
> * `/retirements/:id` => Get single retirement (where `:id` is `/:account_id/:retirement_index`
> 
> Seems like a common pattern to me when building APIs. The `id` could be the transactionHash instead of `/:account_id/:retirement_index` though

>You are ensured to have an instance of a model or a 404 error.

This makes sense though I'm not sure how much extra utility it gives the consumer.

Having `/:id` routes makes sense because it's a single unique identifier. The `/:accoount_id/:another_random_variable` is really trying to squash query paramaters (or post body filters) in to a route parameter.

At the end of the day if you can argue the case that this approach will have the same re-usability, DRY benefits and flexibility (for if the user wants to fetch retirements under some other constraints) as if we were to simply allow filtering through a request body then I'm for it. Can the current approach do that?

carbonmark-api/src/utils/helpers/utils.constants.ts
```diff
@@ -1,9 +1,9 @@
 import { TOKEN_ADDRESSES } from "../../app.constants";
 
 // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- unable to type environment variables
-const ENV = (process.env.VERCEL_ENV ?? "development") as
-  | "development"
-  | "production";
+const ENV = (
+  process.env.VERCEL_ENV !== "production" ? "development" : "production"
+) as "development" | "production";
```
biwano, line 6: I do not know how it worked before for preview environments

carbonmark-api/src/models/UserProfile.model.ts
```diff
@@ -0,0 +1,15 @@
...
+//This model matches the document structure in https://console.firebase.google.com/project/klimadao-staging
+export const UserProfileModel = Type.Object({
+  handle: Nullable(Type.String()),
+  username: Type.String(),
+  description: Nullable(Type.String()),
+  profileImgUrl: Nullable(Type.String()),
+  updatedAt: Type.Number(),
+  createdAt: Type.Number(),
+  address: Type.String(),
+});
```
0xemc, line 13: Nice ❤️

carbonmark-api/src/routes/retirements/get.ts
```diff
@@ -0,0 +1,79 @@
...
+
+type FilterItem = string | number | null;
+type FilterRecords =
+  | { [k: string]: FilterItem | FilterItem[] | Filter }
+  | FilterItem
+  | FilterItem[]
+  | null;
+type Filter = FilterRecords | FilterRecords[];
+
+// Handler function for the "/retirements/klima/:account_id/:retirement_index" route
```
0xemc, line 25: This might be stale too?

carbonmark-api/src/routes/retirements/get.ts
```diff
@@ -0,0 +1,79 @@
...
+    // Build GQL query
+    const SearchKlimaRetirementDocument = gql`
+    query searchRetirements {
+      retires(where: ${gqlWhereFragment(retireFilter)}) {
+        ...RetireFragment
+        credit {
+          ...CarbonCreditFragment
+        }
+      }
+    }
```
0xemc, line 62: Any reason we can't add this to the `/gql` folder and interact via the sdk?

carbonmark-api/src/utils/helpers/utils.constants.ts
0xemc: Did this fix and issue you were having? the original seems fine to me (VERCEL_ENV is only undefined when not on VERCEL (so it has a value in preview environments)

carbonmark-api/src/utils/helpers/utils.ts
```diff
@@ -275,3 +275,17 @@ export function asResponse<ReplyType>(
...
+ * @returns
+ */
+export function gqlWhereFragment<T>(obj: T) {
+  // See https://stackoverflow.com/questions/11233498/json-stringify-without-quotes-on-properties
+  // JSON stringify
+  let json = JSON.stringify(obj);
+  // Remove double quotes around attribute names
+  json.replace(/\\"/g, "\uFFFF"); // U+ FFFF
+  return (json = json.replace(/"([^"]+)":/g, "$1:").replace(/\uFFFF/g, '\\"'));
+}
```
0xemc, line 291: I can't see this being used anywhere?

carbonmark-api/test/routes/retirements/[id]/get.test.ts
Comment (0xemc) [inline carbonmark-api/test/routes/retirements/[id]/get.test.ts:1]: can you add a test for non-existent hash plz

carbonmark-api/test/routes/retirements/[id]/provenance/get.test.mocks.ts
```diff
@@ -0,0 +1,11 @@
+import nock from "nock";
+import { GRAPH_URLS } from "../../../../../src/app.constants";
+import { fixtures } from "../../../../fixtures";
+const mockProvenanceRecord = fixtures.digitalCarbon.provenanceRecord;
+
+export const mockDigitalCarbonProvenanceRecords = () =>
+  nock(GRAPH_URLS["polygon"].digitalCarbon)
+    .post("")
```
Comment (0xemc) [inline carbonmark-api/test/routes/retirements/[id]/provenance/get.test.mocks.ts:8]: Please add the query name lookup so that we dont trample on other requests to `.digitalCarbon`

As an example:

https://github.com/KlimaDAO/klimadao/blob/512ee69aa794e276bcd9cf0526360ff99cb11960/carbonmark-api/test/routes/projects/get.test.mocks.ts#L61

carbonmark-api/test/routes/retirements/[id]/provenance/get.test.ts
Comment (0xemc) [inline carbonmark-api/test/routes/retirements/[id]/provenance/get.test.ts:1]: A test case for invalid hash would be great here too

carbonmark-api/test/routes/retirements/get.test.mocks.ts
```diff
@@ -0,0 +1,53 @@
...
+  jest.mock("../../../src//utils/helpers/users.utils", () => {
+    const carbonProjectsUtils = jest.requireActual(
+      "../../../src/utils/helpers/users.utils"
+    );
+    return {
+      ...carbonProjectsUtils,
+      getProfileByAddress: jest.fn(() => {
+        return mockProfile;
+      }),
+    };
```
0xemc, line 43: Can you use nock or the mockFirebase utility instead

https://github.com/KlimaDAO/klimadao/blob/512ee69aa794e276bcd9cf0526360ff99cb11960/carbonmark-api/test/test.utils.ts#L9

carbonmark-api/src/utils/helpers/utils.constants.ts
biwano: @0xemc  Yeah, it was refusing to deploy on preview environmenets because `process.env.VERCEL_ENV` was equal to `preview`

carbonmark-api/src/routes/retirements/[id]/get.ts
```diff
@@ -0,0 +1,45 @@
+import { FastifyInstance, FastifyReply, FastifyRequest } from "fastify";
+import { gql_sdk } from "../../../utils/gqlSdk";
+import { formatRetirement } from "../../../utils/helpers/retirements.utils";
+import { getProfileByAddress } from "../../../utils/helpers/users.utils";
+import { Params, Querystring, schema } from "./get.schema";
+
+// Handler function for the "/retirements/klima/:account_id/:retirement_index" route
```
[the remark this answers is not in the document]
Comment (biwano) [reply carbonmark-api/src/routes/retirements/[id]/get.ts]: Done in f034c071d

carbonmark-api/src/routes/retirements/get.ts
biwano: Done in 3ce2fc8af

carbonmark-api/src/routes/retirements/get.ts
biwano: @0xemc Because the where clause is dynamic. . I searched for ways to do it via the SDK but could not find any.
If you look at the `/projects` endpoint we have the same issue but we solve it in a convoluted way (by passing all possible values to the query arguments when we do not want to filter). There it is not possible because there is an infinite number of beneficiaryAddresses and retirement indexes.

carbonmark-api/src/utils/helpers/utils.ts
biwano: It is in [carbonmark-api/src/routes/retirements/get.ts](https://github.com/KlimaDAO/klimadao/pull/2034/files/5ea83c65f65f4b4794be22d5749045e68a1812a6#diff-fc9cd5d99f9a8cc740683d1aa0e159094aa7c770d59f9f7c6f77daa41596c94e)

carbonmark-api/test/routes/retirements/[id]/provenance/get.test.mocks.ts
Comment (biwano) [reply carbonmark-api/test/routes/retirements/[id]/provenance/get.test.mocks.ts]: Done in cea91c911

carbonmark-api/test/routes/retirements/get.test.mocks.ts
biwano: Done in 39cb34634

carbonmark-api/test/routes/retirements/[id]/get.test.ts
Comment (biwano) [reply carbonmark-api/test/routes/retirements/[id]/get.test.ts:1]: done in 54ba40476

carbonmark-api/test/routes/retirements/[id]/provenance/get.test.ts
Comment (biwano) [reply carbonmark-api/test/routes/retirements/[id]/provenance/get.test.ts:1]: Done in 54ba40476

carbonmark-api/src/routes/retirements/[id]/provenance/get.schema.ts
```diff
@@ -0,0 +1,22 @@
+import { Type } from "@sinclair/typebox";
+import { ProvenanceRecordModel } from "../../../../models/ProvenanceRecord.model";
+import { params, querystring } from "../get.schema";
```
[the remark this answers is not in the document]
Comment (0xemc) [reply carbonmark-api/src/routes/retirements/[id]/provenance/get.schema.ts]: Yep that's fine. As long as the expected params for a route should be clear in the `.schema` file for that route (without having to go digging)

carbonmark-api/src/utils/helpers/utils.constants.ts
0xemc: When you say "refusing the deploy" do you mean the postman tests were failing? Because they should be on this preview build (we're trying to create a release that points to a broken subgraph)

carbonmark-api/src/routes/retirements/get.ts
0xemc: Hmm ok this is something that is going to come up again. We should be able to simply go:

```gql
query searchRetirements($where: Retire_filter) {
```

However I can see that the generated SDK is still just returning all results..

I'll create a ticket to address this.

0xemc (approved):

PR #2049, opened 2023-12-20
Refactor /projects and /projects/:id routes and change /user/:walletorHandle routes

### LINKED ISSUE #1035: [listings] Better solution for fractional purchases and / token dust

### LINKED ISSUE #2020: [pricing] QA regression and sanity review of project pricing
@0xtapi there has been a bunch of regression happening on the site and I need someone with fresh eyes to take a pass at project "pricing" across the site. I'm seeing variations if I'm on the Marketplace page vs a Project Details page vs the Retire page and we should identify where we have issues and where we are sourcing price data for all these locations (i.e. api endpoint vs subgraph etc).

**Pages to review:**
- https://www.carbonmark.com/projects
- https://www.carbonmark.com/retire
- Related project details pages
- https://data.klimadao.finance/overview/price-of-digital-carbon

**Examples of things I'm noticing:**
- Project prices in CM do not typically match exactly what we report for price and selective cost on https://data.klimadao.finance/overview/price-of-digital-carbon Why??
- Related to #1975 ...these 4 projects represent the default projects in the pool, i.e. the projects with no selective cost fees. Are these prices correct on https://www.carbonmark.com/retire and are they being maintained on other screens (project detail pages, retire or buy forms)? For example:

Screenshot 2023-12-12 at 4 59 27 PM
Screenshot 2023-12-12 at 4 59 08 PM

- Sort the marketplace cards from low to highest price. Find this project, today at $0.50. Where does $0.50 come from? Now click through to the project detail page, price is $0.67??

Screenshot 2023-12-12 at 4 50 20 PM

biwano:
## Description

This PR is intented to solve the discrepency on the way prices are computed in the API. It intraduces a massive refactor of the /projects and /projects:id routes so both of them use the same algorithm to compute a project instance. 
This is a proposal. If we use this PR I think we will need to QA the whole application.
Here is what is inside:

- Introduces a minSupply argument to the `/projects` and `/projects/:id` `/user/:walletOrHandle` routes
- Listings and credits are filtered out AT QUERY LEVEL if the supply available in less than the minSupply argument.
- Introduces a `buildProjectEntry` method that composes a Project instance based on CMS project content, polygon-digital-carbon subgraph data, 

Depending on what we decide for the release process, I will create another PR with only the API changes

## Related Ticket

Closes #2020
Closes #1035

## Notes For QA

- The whole Application

Other notes:
- This is a work in progress, please wait that there is an engineering consensus about the PR before QAing.


carbonmark-api/src/graphql/digitalCarbon.gql
```diff
@@ -18,12 +18,16 @@ query getDigitalCarbonProjectsCountries {
...
+  $projectID: String!
+  $vintage: Int
+  $minSupply: BigInt
+) {
   carbonProjects(where: { projectID: $projectID }) {
     ...DigitalCarbonProjectFragment
     carbonCredits(where: { vintage: $vintage }) {
       ...CarbonCreditFragment
-      poolBalances {
+      poolBalances(where: { balance_gt: $minSupply }) {
```
sprrwhwk, line 30: @biwano `min` implies `gte`, not `gt`

carbonmark-api/src/graphql/digitalCarbon.gql
```diff
@@ -54,7 +59,7 @@ query findDigitalCarbonProjects(
     ...DigitalCarbonProjectFragment
     carbonCredits(where: { vintage_in: $vintage }) {
       ...CarbonCreditFragment
-      poolBalances {
+      poolBalances(where: { balance_gt: $minSupply }) {
```
sprrwhwk, line 62: same as above. should be `gte`

carbonmark-api/src/graphql/marketplace.gql
```diff
@@ -61,26 +69,42 @@ query getProjects(
...
+query getProjectById(
+  $projectId: ID!
+  $expiresAfter: BigInt
+  $minSupply: BigInt
+) {
   project(id: $projectId) {
     ...ProjectFragment
-    listings(where: { expiration_gt: $expiresAfter }) {
+    listings(
+      where: { expiration_gt: $expiresAfter, leftToSell_gt: $minSupply }
```
sprrwhwk, line 106: same as above, should be `gte`

carbonmark-api/src/graphql/digitalCarbon.gql
biwano: Thanks! But I noticed an issue with the updatedAt value. I cannot not filter directly in the query. So these got removed.

carbonmark-api/src/utils/helpers/listings.utils.ts
```diff
@@ -31,12 +22,27 @@ export const getCreditListings = async (
...
+    active?: boolean | null;
+    deleted?: boolean | null;
+    leftToSell?: string | null;
+  },
+  minSupply?: number
+) => !!l.active && !l.deleted && Number(l.leftToSell) >= (minSupply || 0);
+
+/**
+ * For marketplace subgraph projects
+ * Returns true if project has an active, unexpired listing
```
Atmosfearful, line 36: This comment is incorrect as `isActiveListing` is only checking for deletion, in other words a listing can be `active: true` and `deleted: false` but also be expired (expiration timestamp < current timestamp)

carbonmark-api/src/routes/users/[walletOrHandle]/get.ts
```diff
@@ -91,7 +92,10 @@ const handler = (fastify: FastifyInstance) =>
           };
         }) || [];
 
-      const listings = user?.listings?.map(formatListing) || [];
+      const listings = getActiveListings(
```
Comment (Atmosfearful) [inline carbonmark-api/src/routes/users/[walletOrHandle]/get.ts:95]: Just a note of caution:
We want to make sure we can still return unexpired listings or those with token dust, that way users can still manage their own listings.
At first glance it looks OK. Maybe we should check if we have a test case to cover these requirements just in case

Atmosfearful (commented):
I haven't reviewed it all yet, here are a few things to consider in the meantime:

Just double checking that you saw this:
https://github.com/KlimaDAO/klimadao/issues/1035#issuecomment-1865020339
And noted that the CMS is not reliable (may not always have content in it for every project)

Other thoughts:
- I have not put this on the roadmap formally yet, but you should be aware we will need to do an even more dramatic refactor to these endpoints. This is because the `Project` model should actually encapsulate _many vintages_. Right now we are actually returning `Credits` of a _specific vintage_ from this endpoint, like `/projects/VCS-191-2009`, so the endpoint is not even named correctly right now -> In the future it will just be `/projects/VCS-191` or perhaps `/credits/VCS-191-2009`
We have designs completed that rework the /projects view and /projects/[id] pages to "merge vintages together" so to speak.

- Another big workstream is Wallet Abstraction which will force us to separate "users" from "wallets". Currently we treat them as the same, because the wallet was the authentication mechanism. My gut feeling is we will have to split them into separate endpoints:
`/users/{uuid}`, `/profiles/{handle}` and `/wallets/{address}`

I'd leave it up to you if you still think this is worth merging, or if we should hold off until the bigger refactors!

biwano:
> I haven't reviewed it all yet, here are a few things to consider in the meantime:
> 
> Just double checking that you saw this: [#1035 (comment)](https://github.com/KlimaDAO/klimadao/issues/1035#issuecomment-1865020339) And noted that the CMS is not reliable (may not always have content in it for every project)

Yes, I am testing against this ticket too

> 
> Other thoughts:
> 
> * I have not put this on the roadmap formally yet, but you should be aware we will need to do an even more dramatic refactor to these endpoints. This is because the `Project` model should actually encapsulate _many vintages_. Right now we are actually returning `Credits` of a _specific vintage_, so the endpoint is not even named correctly right now -> In the future it will just be `/projects/VCS-191` or perhaps `/credits/VCS-191-2009`
>   We have designs completed that rework the /projects view and /projects/[id] pages to "merge vintages together" so to speak.
> * Another big workstream is Wallet Abstraction which will force us to separate "users" from "wallets". Currently we treat them as the same, because the wallet was the authentication mechanism. My gut feeling is we will have to split them into separate endpoints:
>   `/users/{uuid_or_handle}` and `/wallets/{address}`
> 
> I'd leave it up to you if you still think this is worth merging, or if we should hold off until the bigger refactors!

Yes I think we should merge this. From my point of view the biggest benefit of this PR is to merge the algorithms that produces the Project instances and thus reduces maintenance headaches. It has to be done anyways even if the algorithm itself changes later on and applies to another class.

carbonmark-api/src/utils/helpers/listings.utils.ts
biwano: I updated the comment in b46e71311
Expired listings are already filtered at gql query level

carbonmark-api/src/routes/users/[walletOrHandle]/get.ts
Comment (biwano) [reply carbonmark-api/src/routes/users/[walletOrHandle]/get.ts:95]: Yep, I created a test case on my account. It made sure that:
 - I could see my dust listing.
 - Other users could not see it (Done at UI level in 071e9f6bf)

jabby09:
@0xtapi this one is now ready for QA, see the notes above in the PR description

carbonmark-api/src/routes/projects/get.ts
```diff
@@ -86,18 +92,26 @@ const handler = (fastify: FastifyInstance) =>
...
-          poolProjectData: { ...project, carbonCredits: [credit] },
-          key,
-        });
+        // Create the entry if does not exist
+        if (!ProjectMap.has(key)) {
+          ProjectMap.set(key, {
+            poolProjectData: { ...project, carbonCredits: [] },
+            key,
+          });
+        }
```
0xemc, line 112: Are we hiding duplicate or multiple credits with the same key here? This goes back to the multiple credits for the same project problem from earlier.

0xemc (changes requested):

carbonmark-api/src/routes/projects/[id]/get.schema.ts
```diff
@@ -21,6 +21,13 @@ const querystring = Type.Object({
       examples: ["1620000000"],
     })
   ),
+  minSupply: Type.Optional(
+    Type.Number({
+      description:
+        "Only consider projects listings and pools that contains at least this amount of credits (Tonnes)",
+      examples: ["0"],
+    })
+  ),
```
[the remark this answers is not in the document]
Comment (biwano) [reply carbonmark-api/src/routes/projects/[id]/get.schema.ts]: Yes nice!
Done in a5168a113273ad0e40b74aa0a0088ecde353dda4

carbonmark-api/src/routes/projects/get.schema.ts
```diff
@@ -32,6 +32,13 @@ export const querystring = Type.Object({
       examples: ["1620000000"],
     })
   ),
+  minSupply: Type.Optional(
+    Type.Number({
+      description:
+        "Only return projects listings and pools that contains at least this amount of credits (Tonnes)",
+      examples: ["0"],
+    })
```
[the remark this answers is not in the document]
biwano: Done in a5168a113273ad0e40b74aa0a0088ecde353dda4

carbonmark-api/src/routes/projects/[id]/get.ts
```diff
@@ -51,49 +58,38 @@ const handler = (fastify: FastifyInstance) =>
...
     }
 
-    if (!projectDetails) {
+    const minSupply = request.query.minSupply || 0;
+
+    if (!cmsProject) {
       // only render pages if project details exist (render even if there are no listings!)
       return reply.notFound();
     }
+    const poolProject = digitalCarbonCredits?.carbonProjects?.at(0);
```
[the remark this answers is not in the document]
Comment (biwano) [reply carbonmark-api/src/routes/projects/[id]/get.ts:67]: In this PR. I try not to change the endpoints input and outputs. So the user still queries one credit at the time eg. VCS-191-2009

carbonmark-api/src/routes/projects/get.ts
biwano: Hmm... Each credit should have a different key eg. `VCS-191-2009` so there would not be hidden credits if I understand this credit thing right.
There is surely a naming issue there. `projectMap` should be `creditMap` but maybe we should have a larger refactor to rename everything properly.

0xtapi:
So, pricing issue on the marketplace seems to be fixed. I checked maybe ~40 different projects and none of them had unexpected pricing or dust issues. The projects which do show weird pricing behavior on production (like https://www.carbonmark.com/projects/VCS-981-2017: $1.00 on the marketplace, $0.78 in project details) show the right values everywhere. Prices across the marketplace now also correspond with what https://data.klimadao.finance/overview/price-of-digital-carbon is showing. No more silly prices!

Could also not find any dust left on the marketplace. There are projects like this one: https://carbonmark-git-2020hidedust-klimadao.vercel.app/projects/VCS-1883-2018. This has something like 0.001 in the BCT pool on production, so that dust has been removed and only a seller listing remains. If I understand the related tickets correctly, that's how it's supposed to be.

Found 2 issues, though:

1. When I pick a project on the marketplace and RETIRE NOW, this pops up:
retbug

The link to PolygonScan currently points to Mumbai. Need to change to mainnet.

2. Small QOL issue here:
numbererr

It doesn't accept 0.5, only 0.50. Can this be changed so it just assumes an additional 0 if no value is specified?

I did not test whether I can see other users' dust in their profiles, but @biwano has basically done this already:

> Yep, I created a test case on my account. I made sure that:
> 
> * I could see my dust listing
> * Other users could not see it when viewing my profile (Done at UI level in [071e9f6](https://github.com/KlimaDAO/klimadao/commit/071e9f6bfd50aaee1a6a61b7c989830db90b68d0))

@jabby09 do you wanna go over this as well? It's a pretty big one, so... But if you trust my judgement, you can set this to QA done as soon as you see this 🫡

carbonmark/1
0xemc, line 1: I'm not sure what this file is? Please remove

0xemc (changes requested):

carbonmark-api/src/routes/projects/get.ts
biwano: Ha Ok!!!! I understand now.
Indeed, my code is hiding assets. But yeah lets do that in another PR.

carbonmark-api/src/routes/projects/get.utils.ts
```diff
@@ -54,88 +71,6 @@ export const getDefaultQueryArgs = async (
...
- * For a given project from polygon-bridged-carbon, returns array of pool price strings
- * Chooses between default and selective price.
- * Sorted cheapest first.
- */
-type ProjectBalances = Record<
-  string,
-  { balance: string; creditTokenAddress: string }
->;
-
-export const getDigitalCarbonTokenPrices = (
```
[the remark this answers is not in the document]
biwano: So, this was the part where the prices where computed for the `/projects` endpoint. It was doing the same thing that was done in the `/projects/:id` endpointwith the `getProjectPoolPricesAndStats` but (i think) less accurately (hence the discrepencies in prices throughout the site). So I deleted it, refactored `getProjectPoolPricesAndStats` so it could accomodate both endpoints.

carbonmark-api/src/utils/helpers/utils.ts
```diff
@@ -275,3 +276,12 @@ export function asResponse<ReplyType>(
...
 }
+
+/**
+ * Formats tonnes for a subgraph query
+ */
+export function formatTonnesForSubGraph(qty = 0, decimals = 18) {
+  return ethers.BigNumber.from(qty)
+    .mul(ethers.BigNumber.from(10).pow(decimals))
+    .toString();
+}
```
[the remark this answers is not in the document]
biwano: Yes nice catch!

carbonmark/1
biwano: oups
done in 68ecae5ff889e189752b3e2b7d492d684c2abb26

carbonmark-api/src/routes/projects/get.utils.ts
```diff
@@ -250,53 +255,27 @@ const pickBestPrice = (
...
+      creditId,
+      marketplaceProject: data.marketplaceProjectData,
+      poolProject: data.poolProjectData,
+      cmsProject: carbonProject,
+      allPoolPrices,
+      network,
+      minSupply,
+    });
+    // Only return projects with supply
+    // TODO: Maybe this should be controlled via a query parameter
```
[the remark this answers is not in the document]
biwano: Yeah, but 
- the isActiveListing method is called within `buildProjectEntry` so `buildProjectEntry` may still return a project with empty listings
- same for the pools

carbonmark-api/src/utils/helpers/fetchPoolPricesAndStats.ts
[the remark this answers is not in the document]
biwano: Yes it was refactored as `getProjectPoolPricesAndStats.ts` because the function no longer does the `fetch` part.

biwano:
> So, pricing issue on the marketplace seems to be fixed. I checked maybe ~40 different projects and none of them had unexpected pricing or dust issues. The projects which do show weird pricing behavior on production (like https://www.carbonmark.com/projects/VCS-981-2017: $1.00 on the marketplace, $0.78 in project details) show the right values everywhere. Prices across the marketplace now also correspond with what https://data.klimadao.finance/overview/price-of-digital-carbon is showing. No more silly prices!
> 
> Could also not find any dust left on the marketplace. There are projects like this one: https://carbonmark-git-2020hidedust-klimadao.vercel.app/projects/VCS-1883-2018. This has something like 0.001 in the BCT pool on production, so that dust has been removed and only a seller listing remains. If I understand the related tickets correctly, that's how it's supposed to be.
> 
> Found 2 issues, though:
> 
> 1. When I pick a project on the marketplace and RETIRE NOW, this pops up:
> 
> retbug
> The link to PolygonScan currently points to Mumbai. Need to change to mainnet.

@0xtapi I think it is the same on all preview environments.
I suppose the application should auto detect the network behing used independently of the execution environment.
I made a separate ticket: https://github.com/KlimaDAO/klimadao/issues/2072

> 
> 2. Small QOL issue here:
> 
> numbererr
> It doesn't accept 0.5, only 0.50. Can this be changed so it just assumes an additional 0 if no value is specified?

Created a separate issue for this one too: https://github.com/KlimaDAO/klimadao/issues/2073

> 
> I did not test whether I can see other users' dust in their profiles, but @biwano has basically done this already:
> 
> > Yep, I created a test case on my account. I made sure that:
> > 
> > * I could see my dust listing
> > * Other users could not see it when viewing my profile (Done at UI level in [071e9f6](https://github.com/KlimaDAO/klimadao/commit/071e9f6bfd50aaee1a6a61b7c989830db90b68d0))
> 
> @jabby09 do you wanna go over this as well? It's a pretty big one, so... But if you trust my judgement, you can set this to QA done as soon as you see this 🫡

carbonmark-api/src/utils/helpers/cms.utils.ts
```diff
@@ -25,28 +24,45 @@ const projectKey = ({
...
+  };
+
+/**
+ * Merges a carbon project with carbon content
+ */
+const mergeCarbonProjectAndContent = (
+  project?: CmsProject,
+  content?: CmsProjectContent
+) => {
+  if (isNil(project) || isNil(content)) return null;
```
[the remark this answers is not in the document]
biwano: Note: in staging we do not return projects with no cmsData in the `/projects/:id` route https://github.com/KlimaDAO/klimadao/blob/f5844195ddaf2384d142f2f6f0f2a866c5cb4f35/carbonmark-api/src/routes/projects/%5Bid%5D/get.ts#L54
So I enforced the same in the `/projects` route in 4a7a29bb6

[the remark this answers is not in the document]
0xemc: Ok we need to document this behaviour somewhere I think. Along with all the other requirements Atmos listed [here](https://github.com/KlimaDAO/klimadao/issues/1035#issuecomment-1865020339)

0xemc (approved):
Approved pending the tick from @jabby09 or @0xtapi.

Atmosfearful (approved):
More notes as I compare old responses to new responses:

**On /projects endpoint:**
- Only 1 backwards-incompatible change, very nice 👏 (was it intentional?)
image

Noting some cleanups that are outside the scope of this PR, or can be addressed when we do bigger refactors
- Remove `project.listing.project` as this is redundant info
- If `long_description` or other fields are really blowing up the json size we can mark them opt-in via `?include` param
- As we add support for more registries we will really feel the need for pagination...

Everything else looks great! 💯

carbonmark-api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@klimadao/carbonmark-api",
-  "version": "4.1.3",
+  "version": "4.2.0",
```
Atmosfearful, line 3: oh, due to that breaking change with the country field, this needs to be a Major bump actually.

Atmosfearful (changes requested):

biwano:
> More notes as I compare old responses to new responses:
> 
> **On /projects endpoint:**
> 
> * Only 1 backwards-incompatible change, very nice 👏 (was it intentional?)
>   image

Yes I had to. For both endpoints to use the same algorithm. I had to adapt the country field for `/projects`

> 
> Noting some cleanups that are outside the scope of this PR, or can be addressed when we do bigger refactors
> 
> * Remove `project.listing.project` as this is redundant info
> * If `long_description` or other fields are really blowing up the json size we can mark them opt-in via `?include` param

Yes very good idea, I think we could even do that in a generic way for all endpoints.

> * As we add support for more registries we will really feel the need for pagination...

That is for sure.

> 
> Everything else looks great! 💯

carbonmark-api/package.json
biwano: Done in 2b927577f

carbonmark-api/src/models/Project.model.ts
```diff
@@ -13,22 +15,30 @@ export const ProjectModel = Type.Object({
...
   price: Type.String(),
+  prices: Type.Array(TokenPriceModel),
+  stats: Type.Object({
+    totalBridged: Type.Number(),
+    totalRetired: Type.Number(),
+    totalSupply: Type.Number(),
+  }),
   listings: Nullable(Type.Array(ListingModel)), // null when listings are empty
-  images: Nullable(
+  images: Type.Optional(
```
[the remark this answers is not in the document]
biwano: Done in baebdf8f0

jabby09 (approved):

PR #2052, opened 2023-12-20
[APP] ICR FE Integration

psparacino:
## Description

Carbonmark Integration for ICR credits

**Notes for reviewers:**

1) The retirement data for receipts in dependent on merge of #2048 . This will eventually all be moved to an endpoint.
2) `serialization` is one of three possible query parameters for an individual on the ICR api. `serialization` is used over the other two options (`contractAddress` and ICR's version of a `projectID`) whenever possible for consistency.
3) ICR quantities are regular integers, so the amounts are handled separately for the ICR registry [here](https://github.com/KlimaDAO/klimadao/compare/APP_feature/icr-api-integration?expand=1#diff-1a13cfc031e74278375852de1ca0bf4b8d421b8ed5125f837cd8cc799016d9dbR62). 
4) This PR uses a mainnet testing setup: mainnet RA, carbonmark contract, ICR Carbon Registry contract, ICR subgraph, digital-carbon and marketplace subgraphs. These all need to be changed to production versions before merge.
5) The API is currently hardcoded to the API PR Preview #2051 . This should be [removed](https://github.com/KlimaDAO/klimadao/pull/2052/files#diff-a1be0ef50959b412feee17a3bcc720124c83ccf3e1ab62f240d5ad0d669a1aa4R38) before merge.

To change before merge:
- [x] AggregatorV2 contract
- [x] Carbonmark contract 
- [x] ICR Carbon Registry contract
- [x] ICR subgraph
- [x] Digital-carbon subgraph
- [x] Marketplace subgraph
- [x] Hardcoded Preview URL

Other improvements were noted along the way that weren't necessarily ICR specific but were rolled into this PR:
1) Removing duplicate prices or quantities when only editing one option in a listing: [here](https://github.com/KlimaDAO/klimadao/compare/APP_feature/icr-api-integration?expand=1#diff-7c25eb7f0d621afca61781b09f378ef2b304e011f6bdfe6a1b8785d2b1c616b5R92)
2) Fix recurring nesting error because of title construction in Text component [here](https://github.com/KlimaDAO/klimadao/compare/APP_feature/icr-api-integration?expand=1#diff-4d85183ccd99ed8e794114df235358835b17f2e288ac462bda3c5625383179c0R37)

## Related Ticket

Also related to #2051 

## Notes For QA

Specific pages, components or journeys that might be affected:
- All need to be tested

Relevant preview URLs:
- 

Other notes:
- This is a large PR(!). Please feel free to reach out for any clarification or to pair review any specific section.

PR #1843, opened 2023-11-07
DRAFT: Feature/icr-api-integration

psparacino:
Base working branch for ICR app and api PRs

APP PR: #2052 

API PR: #2051 

All changes in the PRs are reflected here

PR #2053, opened 2023-12-21
revised approval modal copy for ICR tokens

### LINKED ISSUE #1989: [ICR] ERC1155 Approval Copy changes
The ERC1155 approval mechanism differs from the ERC20 and the approval modal copy should reflect that change as the signing message will be different in MM or other wallets.

The ERC1155 uses setApprovalForAll which allows all amounts for all tokenIds in the contract.

ChazSchmidt:
## Description

Revises the approval modal text to add context regarding ICR tokens setApprovalForAll 

## Related Ticket

Closes #1989

### DIFF
--- a/app/locale/en-pseudo/messages.po
+++ b/app/locale/en-pseudo/messages.po
@@ -6,65 +6,110 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "X-Generator: @lingui/cli\n"
 "Language: en-pseudo\n"
+"Project-Id-Version: \n"
+"Report-Msgid-Bugs-To: \n"
+"PO-Revision-Date: \n"
+"Last-Translator: \n"
+"Language-Team: \n"
+"Plural-Forms: \n"
 
-#: components/DisclaimerModal/index.tsx:84
-msgid "Acknowledge and Accept"
-msgstr ""
-
-#: components/TransactionModal/Approve.tsx:93
-#: components/views/Bond/index.tsx:454
-#: components/views/Stake/index.tsx:252
-#: components/views/Wrap/index.tsx:230
-msgid "Approve"
-msgstr ""
-
+#. js-lingui-explicit-id
 #: components/CarbonBalancesCard/index.tsx:126
 msgid "Balances"
 msgstr ""
 
-#: components/views/Redeem/index.tsx:25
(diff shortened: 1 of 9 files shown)


psparacino:
Content-wise looks fine to me but will wait for @jabby09 to finalize

Just one comment re: PR pointing given that the ICR branch workflow is somewhat convoluted:

This is the base working ICR branch: https://github.com/KlimaDAO/klimadao/pull/1843. Ideally you can point PRs here which I can then merge (the PR branches are updated from this base branch).

This is the FE PR: https://github.com/KlimaDAO/klimadao/pull/2052

Staging will still work; it will just display a note about ICR when no credits yet exist on mainnet. The ICR copy should probably also be conditional.

I'll just copy this over for now, so just a note for the other copy changes

jabby09:
> Content-wise looks fine to me but will wait for @jabby09 to finalize

yea I assigned QA on this one to @0xtapi

0xtapi:
The content looks fine, but I wonder if we can do a little better context wise. Some thinkings:

1.) The current text might be confusing for a user who doesn't know what ICR credits are:

"The allowance below reflects the sum of all your listings for this specific token. For ICR credit tokens **(what are these tokens?)**, the allowance will reflect the sum of all your ICR token listings **(how do I know whether my listings are ICR token listings?)**."

I assume that by the time ICR credits go live, we would have an accompanying blog post ready (knowing Peter, I'm pretty sure this is a marketing opportunity he would happily take). Could we, in the future, extend the current message with a "Click here to learn more" text that links to that post?

2.) Is it possible to make this text conditional, i.e. it only shows if the user actually intends to list ICR credits? I'm not sure if a user who lists other types of credits would need to be confronted with this message. Could be confusing and cause hesitation.

For now, this current iteration is fine and I'm gonna slap my seal of approval on this PR. But we should consider holding the user's hand a little more in the future once this goes live to make it a 100% smooth, no-confusion experience. After all, Carbonmark exists to make Web3 carbon retirement as easy as possible without leaving any room for head scratching, right?

Atmosfearful:
I am going to pass this to @psparacino since it's pointed at his branch

0xemc (approved):

psparacino (approved):

PR #2061, opened 2024-01-02
Rejigger CMS entities to support ICR projects and add subcategories and methodology links

0xAurelius:
## Description

Supports additional registries (including ICR) in the CMS, plus adds subcategories for projects and methodology links

Should not have any breaking impact on the frontend's API as only new fields were added (existing fields untouched)


Atmosfearful (approved):
Looks gooood 🚢 Nice work!

PR #2066, opened 2024-01-03
fix activity designs

### LINKED ISSUE #2039: [activity] UI inconsistency in activity feed
There is a small inconsistency in the Carbonmark activity feed, for example at https://www.carbonmark.com/projects/VCS-981-2017

Screenshot of the issue below:
incon

When you create a listing, it shows "1t", including the symbol that indicates tonnes. When you change the amount, it just shows the numerical value. For consistency's sake, this should show the "t" symbol as well, so "You updated quantity 1t -> 2t" for example.

sprrwhwk:
## Description
This PR fixes general problems with the activity feed:

* Adds the `t` to quantities for quantity updates
* Makes the UI look more like the figma design in general

|Before|After|
|---|---|
|grafik|grafik|
|grafik|grafik|

## Related Ticket

Closes #2039

### DIFF
--- a/carbonmark/components/Activities/Activity.tsx
+++ b/carbonmark/components/Activities/Activity.tsx
@@ -1,6 +1,6 @@
 import { useWeb3 } from "@klimadao/lib/utils";
 import { t } from "@lingui/macro";
-import { East } from "@mui/icons-material";
+import EastIcon from "@mui/icons-material/East";
 import { Text } from "components/Text";
 import { createProjectLink } from "lib/createUrls";
 import { formatToPrice, formatToTonnes } from "lib/formatNumbers";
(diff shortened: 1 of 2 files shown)


sprrwhwk:
cc @theprofessora

jabby09 (changes requested):
@sprrwhwk very nice, looks much better

Only think I could see (compared to the designs) is the light blue frame should alternate, so in this example activity 1, 3, and 5 would have the blue backing but activity 2 and 4 would not.

Screenshot 2024-01-03 at 2 59 38 PM

sprrwhwk:
> @sprrwhwk very nice, looks much better
> …

Fixed in the latest commit

jabby09 (approved):

0xemc (approved):

PR #2074, opened 2024-01-05
reduce min fill amount (debug)

Atmosfearful:
Reduce the min fill amount to 0.001 for testing purposes


sprrwhwk:
@Atmosfearful I believe we can close this, right?

PR #2077, opened 2024-01-08
[API] Refactor /projects and /projects/:id routes and change /user/:walletorHandle routes

biwano:
## Description

API PR for the prices issues (see https://github.com/KlimaDAO/klimadao/pull/2049)

## Related Ticket

Also related to #2049

## Notes For QA

- 

Relevant preview URLs:
- 

Other notes:
- Technical PR. Only consider #2049 for QA


carbonmark-api/src/models/Project.model.ts
```diff
@@ -1,34 +1,46 @@
...
-  country: Type.Object({ id: Type.String() }),
+  country: Type.String(),
   region: Type.String(),
   price: Type.String(),
+  prices: Type.Array(TokenPriceModel),
+  stats: Type.Object({
+    totalBridged: Type.Number(),
+    totalRetired: Type.Number(),
+    totalSupply: Type.Number(),
+  }),
```
0xemc, line 33: I see we've aligned this model with the `DetailedProjectModel` was this so you could consolidate the `buildProjectEntry` function?

The original intent for having the two models was that some of this data is not required in the case of `/projects` given that the payload size is getting prohibitively large we should only provide the data that is necessary.

**Staging**
Screenshot 2024-01-08 at 11 19 05 am

**2020_hide_dust_api**
Screenshot 2024-01-08 at 11 21 35 am

It looks like we might have doubled that payload size?

I agree that a seperate model just to cater for a specific use case is not great however I'm not sure if the solution here is to add paging or even graphql so that the consumer can decide what data is necessary for their use case..

Thoughts?

@Atmosfearful @biwano

0xemc (approved):

Atmosfearful (approved):

PR #2081, opened 2024-01-08
Emc/chore/2079-kubb-v2

0xemc:
## Description

Upgrade Kubb to v2.x

* Removes `useEffect` hack to in place of correct keys

## Notes For QA

Relevant preview URLs:
- `/projects`
- `/projects/[id]`
- Filter Modal in /projects

Other notes:
-


jabby09:
@0xtapi FYI

carbonmark/kubb.config.ts
```diff
@@ -29,10 +29,22 @@ export default defineConfig(async () => {
...
+          importPath: "../client",
+        },
       }),
       createSwaggerSWR({
-        client: "./lib/api/client.ts",
+        client: { importPath: "../client" },
+        /**
+         * Unfortunately we need to disable hook generation for PUT and POST calls.
+         * This should be temporary until the issue can be resolved, see:
+         *  */
```
biwano, line 41: You forgot the link to the issue there?

carbonmark/package.json
```diff
@@ -10,7 +10,7 @@
     "lingui:compile": "lingui compile",
     "lingui:extract:dev": "node ../scripts/lingui.js extract --clean --overwrite --locale=en && node ../scripts/lingui.js extract --clean --overwrite --locale=en-pseudo",
     "postinstall": "npm run lingui:compile",
-    "generate:types": "tsx ./scripts/fetch-api-schema.ts && kubb"
+    "generate:types": "tsx ./scripts/fetch-api-schema.ts & kubb && cp lib/api/client.ts .generated/carbonmark-api-sdk"
```
biwano, line 13: Did you change `&&` into `&` on purpose there?

biwano (commented):
Just 2 questions.

carbonmark/package.json
0xemc: I did yea, we need to wait for the schema fetch before running the generator. This didnt cause any issues yet but it could run the gen on an old schema (possibly)

carbonmark/kubb.config.ts
0xemc: Good pickup, updated

carbonmark/lib/utils/string.utils.ts
```diff
@@ -0,0 +1,15 @@
+import { isStringArray } from "./types.utils";
+
+/**
+ * Joins an array of strings into a single string.
```
sprrwhwk, line 4:
```suggestion
 * Joins an array of strings into a single comma-separated string.
```

carbonmark/lib/utils/string.utils.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+export const joinArray = (value: string | string[]): string =>
+  isStringArray(value) ? value.join(",") : value;
+
+/**
+ * Checks if a string is empty and returns undefined if true.
+ * If the string is not empty, it returns the string as is.
+ */
+export const emptyToUndefined = (value: string): string | undefined =>
+  value === "" ? undefined : value;
```
sprrwhwk, line 15: I think what we are trying to do here in the scope of this PR is filter empty strings right? Mapping them to undefined seems like a detour that also gets us there but I believe filter is what we should use.

sprrwhwk (changes requested):

0xtapi:
Vintage filters and Blue Carbon filter now work 🫡

carbonmark/lib/utils/string.utils.ts
0xemc: Sort of, this utility on it's own can be useful but in the context of the query params we're trying to remove attributes from the object that have the "" value. Perhaps we should be using `omitBy` instead?

0xemc: Ok I've updated to hopefully more clearly sanitise the query params:

1. Remove nullish values from arrays & nullish values from the object
2. Join any arrays using ","

This should cover more cases and be a little clearer

sprrwhwk (approved):

PR #2082, opened 2024-01-08
Remove google tag manager when not in production

### LINKED ISSUE #2057: Disable `google-analytics` and `UserTracker` components when not in production
If `IS_PRODUCTION` from `carbonmark/lib/constants.tsx` is false, the GA and LO analytics scripts should not execute.
See `_app.tsx`

Stretch goals:
- Any smarter abstractions we can use?
  - Note that Plausible is called from `PageHead` where the other scripts are called from _app.tsx
- Any nextjs builtins we can use?

biwano:
## Description

Removes google tag manager when not in production. For app, site and carbonmark
It does not specifically disable UserTracker, but since lucky orange is not loaded. The traces set up by user tracker are never picked up.

I noticed that app and site share the same google tag manager id. Is it normal?

## Related Ticket

Closes #2057
Also related to <issue-number>

## Notes For QA

- None

Relevant preview URLs:
- None

Other notes:
- To see if it works, visit Lucky orange and google analytics backends

### DIFF
--- a/app/pages/_app.tsx
+++ b/app/pages/_app.tsx
@@ -1,3 +1,4 @@
+import { GTMScript } from "@klimadao/lib/components";
 import "@klimadao/lib/theme/globals.css";
 import "@klimadao/lib/theme/normalize.css";
 import "@klimadao/lib/theme/variables.css";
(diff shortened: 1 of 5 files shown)


biwano:
@Atmosfearful @0xemc Extra notes:
- I noticed that app and site share the same google tag manager tag. Is it normal?
- I checked GTM doc :
image
It seems we should have two scripts (one in head, one at the start of body). Ours is at the end of the body. Is this for performance reasons?
- Concerning stretch goals. I did not find any nextJS builtins that could make things easier.

Atmosfearful (approved):
Looks good ty. I have no idea what the implications are of Body vs Head vs both... Good question for GPT :) I'll leave the decision to you!

jabby09 (approved):

PR #2127, opened 2024-01-15
Fix Retirement Quantities

### LINKED ISSUE #2123: [retirements] retirement quantity doesn't display correctly
Issue:
Screenshot 2024-01-15 115639

Go to https://www.carbonmark.com/retire to reproduce.

While it would be nice to retire a few quintillion tonnes of carbon, we still have some way to go before we get there, so let's fix this bug in the meantime.

I also checked other areas of the site whether they display numbers in some undesired way, but couldn't find anything. Seems like only the Retire page is affected.

psparacino:
## Description

Fixes retirement quantities bug for overview and total carbon retired.

Creating issue to track improvements discovered by this issue.

## Related Ticket

Closes #2123

### DIFF
--- a/carbonmark/components/pages/Retire/Activity/Overview.tsx
+++ b/carbonmark/components/pages/Retire/Activity/Overview.tsx
@@ -1,13 +1,11 @@
 import { cx } from "@emotion/css";
-import {
-  getTotalCarbonRetired,
-  getTotalRetirements,
-  trimWithLocale,
-  useWeb3,
-} from "@klimadao/lib/utils";
+import { formatUnits, trimWithLocale, useWeb3 } from "@klimadao/lib/utils";
 import { t } from "@lingui/macro";
 import { Card } from "components/Card";
 import { Text } from "components/Text";
+import { ethers } from "ethers";
+import { parseUnits } from "ethers/lib/utils";
+import { queryKlimaRetiresByAddress } from "lib/retirementDataQueries/retirementDataViaPolygonDigitalCarbon";
 import { useRouter } from "next/router";
 import { FC, useEffect, useState } from "react";
 import * as styles from "./styles";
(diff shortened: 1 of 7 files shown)


biwano (commented):
Looks good to me,

0xtapi:
Retire page is fixed, but there is an issue remaining with the PDF certificates, see:
Screenshot 2024-01-15 122548

To reproduce, acccess https://carbonmark-git-fix-retirement-summary-2123-klimadao.vercel.app/retirements/0xca2b532Dd6fBff14430b4FF2cFB2823eeB1245c8/9 and download PDF.

carbonmark/lib/retirementDataQueries/retirementDataViaPolygonDigitalCarbon.ts
```diff
@@ -121,15 +121,15 @@ export const queryKlimaRetireByIndex = async (
...
   try {
     const json: QueryKlimaRetires = await fetchGraphQL(
       generateKlimaRetireQuery(beneficiaryAddress)
     );
-    return !!json.data.klimaRetires.length && json.data.klimaRetires;
+    return json.data.klimaRetires || [];
   } catch (e) {
     console.error("Failed to query KlimaRetiresByAddress", e);
-    return Promise.reject(e);
+    return [];
```
0xemc, line 132: If we're not going to be rejecting the promise we probably need to be throwing an error here instead.

If it's expected behaviour "sometimes" then we should be using `console.info`

carbonmark/pages/api/certificates/[beneficiaryAddress]/[retirementIndex]/index.ts
```diff
@@ -45,7 +45,9 @@ export default async function handler(
     if (!retirement) {
       return res.status(404).send("Not found");
     }
-
+    if (retirement.retire.credit.project.registry !== "ICR") {
```
Comment (0xemc) [inline carbonmark/pages/api/certificates/[beneficiaryAddress]/[retirementIndex]/index.ts:48]: Is this something we should be cleaning up at the api? If the number of decimals for the amount depends on the registry we should be standardising this.

carbonmark/pages/api/retirements/[beneficiary_address]/index.ts
```diff
@@ -8,6 +9,17 @@ export interface APIDefaultResponse {
   message: string;
 }
 
+const formatAmount = (r: KlimaRetire["retire"]) => {
```
Comment (0xemc) [inline carbonmark/pages/api/retirements/[beneficiary_address]/index.ts:12]: Can you give this a more descriptive and put it in a `.utils.ts` file

carbonmark/components/pages/Retire/Activity/Overview.tsx
```diff
@@ -34,14 +32,25 @@ export const RetireOverview: FC = () => {
...
-        const [totalRetirements, totalCarbonRetired] = await Promise.all([
-          getTotalRetirements({ beneficiaryAddress: address }),
-          getTotalCarbonRetired({ beneficiaryAddress: address }),
-        ]);
+        const totalRetiredTonnesWei = klimaRetires.reduce((acc, r) => {
+          const amount = r.retire.amount;
 
-        setTotalRetirements(totalRetirements);
-        setTotalCarbonRetired(totalCarbonRetired);
+          if (r.retire.credit.project.registry === "ICR") {
```
0xemc, line 40: it might be better to do this "edge" case formatting inside the `queryKlimaRetiresByAddress` function before we move all this in to the API

Issue #2062, opened 2024-01-03
[tech] Migrate to Lingui V4


Atmosfearful:
Lingui v4 has better support for NextJS and the SWC plugin seems to be a bit more stable now.
https://lingui.dev/blog/2023/04/26/announcing-lingui-4.0

### Context & related issues
We previously tried to remove Babel and replace it with SWC in #927 
Migrating to SWC will allow us to ship #1485
Related to / potential solution for https://github.com/KlimaDAO/klimadao/issues/1961

# Migration plan
Many of these changes can be made in advance, in smaller PRs, before upgrading the package version.

Here is the official migration guide please read carefully.
https://lingui.dev/releases/migration-4

Start with Carbonmark for now (ignore `app`, and ignore `site`, site is deprecated).
Try to break it up into many PRs for fast reviews. These are the two major changes I am aware of (please double check the guide to see if I missed any others)
* [ ] Convert these objects, to interpolated strings. This can be 1 or more PRs.
```ts
import { t } from "@lingui/macro"
// old
const message = t({ id: "my.message", message: `My Message` });
// new
const message = t`My Message`;
```
* [ ] Fix [translations outside of react components](https://lingui.dev/releases/migration-4#translation-outside-react-components-migration) to use one of the recommended patterns (see guide). Can also be 1 or more PRs.

PR #2138, opened 2024-01-17
2062 converted objects to interpolated strings

sidneyarida:
## Description

1st part of #2062 

Converted translation objects to interpolated strings

Other notes:
-

Translation keys are changed and will need to re-translate.

From the the [migration guide](https://lingui.dev/releases/migration-4)  (screenshot below) they suggest to add this comment **#. js-lingui-explicit-id** ontop of the key, but from I understood from the project the keys are handled with translations.io and the language files are overwritten. 
So I think it's not doable and may need to re-insert the translations
How do we want to manage it?

Screenshot 2024-01-17 at 15 29 40


0xemc (approved):

0xemc:
@biwano I'm not sure I understand if this will impact us or not

>From the the [migration guide](https://lingui.dev/releases/migration-4) (screenshot below) they suggest to add this comment #. js-lingui-explicit-id ontop of the key, but from I understood from the project the keys are handled with translations.io and the language files are overwritten.
So I think it's not doable and may need to re-insert the translations
How do we want to manage it?

Any ideas?

biwano:
> @biwano I'm not sure I understand if this will impact us or not
> 
> > From the the [migration guide](https://lingui.dev/releases/migration-4) (screenshot below) they suggest to add this comment #. js-lingui-explicit-id ontop of the key, but from I understood from the project the keys are handled with translations.io and the language files are overwritten.
> > So I think it's not doable and may need to re-insert the translations
> > How do we want to manage it?
> 
> Any ideas?

Hi 

If memory serves right, the PO files are updated via the `update_translations.yml` with `npm run extract-strings`. That command
- extracts strings into PO files 
- gets translations from 
- sends new keys to translation.io
The new keys would have no translations. So everything would need to be translated again.

There is maybe one option to not have to retranslate everything: 
- Create a new project
- Initialize the project translations with this https://translation.io/docs/create-library#initialization. It would require a script to convert existing translations into segments
- Switch carbonmark translation.io configuration to use the new project

0xemc:
This PR keeps getting kicked further down the road and I think we should just bite the bullet.

@brunoilponseisae @Atmosfearful If you both agree I think we should merge this and then deal with the consequences (updating translations, shifting to machine translations etc).

0xemc:
@sidneyarida can you resolve the conflicts please 🙏

Atmosfearful (approved):
LFG let's get this in before we switch to the private repo. We can suffer the broken translations. I'll let @0xtapi know we should brace for impact

Issue #2129, opened 2024-01-16
[marketplace] vintage filter showing impossible dates


0xtapi:
Issue:
![Screenshot 2024-01-16 134248](https://github.com/KlimaDAO/klimadao/assets/128430614/06f2623b-ff8c-40fb-8f00-ef1d81f646bc)

https://www.carbonmark.com/projects and select Filters -> Vintage to reproduce.

Currently lists all years between 1902 and 2038, with a few large gaps in between. I think last time we encountered a vintage filter bug this was related to credits which have been bridged but were not part of the KlimaDAO ecosystem. Obviously, this can't be the case here, as there never have been credits from 1902.

Issue #1054, opened 2023-04-19
[filter] Add "contains seller listing" filter option


jabby09:
When filtering projects, given that a project may have multiple listing (pool and seller) it would be helpful to be able to filter for projects that have a seller listing.

jabby09:
@theprofessora I know we will have this as a filter option in CM 2.5 UI design updates but I'd like to get this implemented in the current CM version. 

Can you please add an option in the current filter UI to allow users to select projects in the marketplace that have a seller listing?

theprofessora:
https://www.figma.com/file/5GewbdVaM1Q06CwHmF7d3F/Projects?type=design&node-id=0-1&mode=design

jabby09:
@0xMakka here is an enhancement to the project filtering UI you built.

- see [design here](https://www.figma.com/file/5GewbdVaM1Q06CwHmF7d3F/Projects?type=design&node-id=518-35453&mode=design&t=NGopalMIJn98dp7L-4)
- add a toggle option to filter marketplace projects by those that "have a seller listing"
![Screenshot 2023-11-30 at 1 36 43 PM](https://github.com/KlimaDAO/klimadao/assets/103535195/1ebdcdc9-433c-4f2c-bf89-9bcb1cbd071d)
- should be applied to desktop and mobile screens
- also note the info ? that should be added; this should leverage our standard info pop up component
- _[note] Brice to provide content/copy for the info pop up_
- no requirement to surface this filter option on the marketplace page itself, the user will need to open the filter modal to view or change the toggle setting
- CLEAR FILTERS button in the modal and the marketplace controls should set the toggle to off (i.e. show all projects)

jabby09:
> * [note] Brice to provide content/copy for the info pop up

@theprofessora any suggested text for the info pop up?

PR #2152, opened 2024-01-18
Improve filtering of vintages

biwano:
## Description

Only return polygon-digital-carbon vintages for credits which have a pool with a non zero balance

## Related Ticket

Closes <issue-number>
Also related to #2129

## Notes For QA

Preview URL: https://carbonmark-476cw0noh-klimadao.vercel.app/projects


psparacino (approved):

carbonmark-api/src/utils/helpers/utils.ts
```diff
@@ -47,7 +47,11 @@ export async function getAllVintages(
   digitalCarbonProjects.forEach((project) => {
     project.carbonCredits.forEach((credit) => {
       if (credit.vintage) {
-        uniqueValues.add(credit.vintage.toString());
+        if (
+          credit.poolBalances?.some((balance) => Number(balance.balance) > 0)
+        ) {
+          uniqueValues.add(credit.vintage.toString());
+        }
```
psparacino, line 54: Just a note that this will filter out all the ICR vintages as there are no pool balances for those, just direct seller to buyer listings. 

This isn't currently an issue as the ICR filters are fetched separately from their API at the moment, but the eventual goal is get everything from CM subgraphs and not outside sources.

0xtapi:
Looks good on the frontend. Only vintages remaining that we actually offer 👍

carbonmark-api/src/utils/helpers/utils.ts
jabby09: @psparacino this project is not part of a pool and does show up when filtering https://www.carbonmark.com/projects/VCS-1764-2020

also note we are in the process of adding a "contains seller listing" toggle to the filtering tools
https://github.com/KlimaDAO/klimadao/issues/1054

jabby09 (approved):

carbonmark-api/src/utils/helpers/utils.ts
psparacino: Oh, I just meant the vintage from the project not the project itself. Given that there is so much overlap in the common vintages it's probably a non-issue

sprrwhwk (approved):

carbonmark-api/src/utils/helpers/utils.ts
```diff
@@ -47,7 +47,11 @@ export async function getAllVintages(
   digitalCarbonProjects.forEach((project) => {
     project.carbonCredits.forEach((credit) => {
       if (credit.vintage) {
-        uniqueValues.add(credit.vintage.toString());
+        if (
+          credit.poolBalances?.some((balance) => Number(balance.balance) > 0)
```
0xemc, line 51: Is this something that could be done in gql?

.github/workflows/deploy_carbonmark_api.yml
0xemc, line 1: Is this just for testing purposes?

If you're finding that the deploy is failing please let me know

0xemc (changes requested):

0xemc (approved):

PR #2162, opened 2024-01-22
[api] Support both images and externalMedia media from cms projects

psparacino:
## Description

Project media can appear in two different sources from the cms. 

In order of preference, from `images` on the allProject query. 

The second option is from `externalMedia` on allProjectContent. This PR fetches from both in that order of preference. 

**This PR also filters out any youtube or video links as CM does currently not support them.**

ICR project page with images:

Screen Shot 2024-01-22 at 2 04 33 PM

## Notes For QA
- Individual project pages

### DIFF
--- a/carbonmark-api/package.json
+++ b/carbonmark-api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@klimadao/carbonmark-api",
-  "version": "5.3.0",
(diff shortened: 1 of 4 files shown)


jabby09 (approved):

biwano (approved):
Nice!

Issue #2093, opened 2024-01-10
[tech] Debt and Improvements from ICR epic


psparacino:
In the course of adding support for ICR projects, there many different areas of improvement that came to light to facilitate adding additional registries (ICR debt), clean up previous debt, and create central sources of truth for both registry and project data.

### API
**1. Move `/retirements` data to endpoint:**
a. Currently queries to the digitalCarbon subgraph for retirements data take place in the FE. This should all be moved to a retirements endpoint so there are no direct subgraph queries from the app.

### Subgraph
1. **Project Info on `marketplace` and `digital-carbon` subgraphs**
    - Both subgraphs need current project info for all projects. This is currently manually updated and can only be updated on redeployments.  As some subgraphs (`digital-carbon`) have a significant sync time of around a day, we can dynamically update the project info using ipfs and ipfs fileDatasources on the subgraph.  
    - To go along with the above, a script that assembles the required subgraph project info data from the central sources of cms project data and on-chain project data will facilitate this process.
2. **Add tokenId to marketplace subgraph project info type:**
    - This will negate the need for a modified url (to add vintage vintage as a query param) if retiring ICR. Not a big issue at all, but this makes for one less area to update when adding registries.

### App and/or Api
1. Central source of truth for supported networks
    - Most likely a map in the api constants that NetworkParam uses, as the frontend uses the generated NetworkParam for almost all network references
2. Conditional amount handling
    -  All amounts coming from the subgraph are in 18 decimals at the moment, except for ICR which are regular integers (0 decimals). Rather than updating the amounts individually based on project ids or key, we should have a central source of truth for decimals relating either to registry or projects.

PR #2163, opened 2024-01-22
[api] Remove ICR artifacts 

psparacino:
## Description

In the original version of the ICR integration, ICR's project credit`serialization` was the only reliable way to query the ICR api for a specific project.

After the refactor to use the cms for project data, the serialization is only needed to create the `registry-registryProjectID` format to identify the project when querying the ICR subgraph for ICR holdings. The serialization can be removed everywhere else.

In the app, besides generated files, the serialization only exists as a prop in carbonmark/lib/createUrls and will be deleted there for the APP PR for this api version. 

The end goal is not to use the serialization anywhere in the CM codebase. Once we can query ICR holdings from a CM subgraph this last requirement can be removed.

## Related Ticket

related to [2093](https://github.com/KlimaDAO/klimadao/issues/2093)

### DIFF
--- a/carbonmark-api/package.json
+++ b/carbonmark-api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@klimadao/carbonmark-api",
-  "version": "5.3.1",
+  "version": "5.3.2",
   "description": "An API for exploring Carbonmark project data, prices and activity.",
   "main": "app.ts",
   "scripts": {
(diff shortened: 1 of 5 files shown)


biwano (approved):
Nice!
You will need to change the api version before merging.

0xemc (approved):

Issue #2159, opened 2024-01-19
[fiat] Support ICR and tokenIds in fiat-to-retirement tooling


Atmosfearful:
- add support for a `tokenId` parameter on quotes and checkout requests
- token id should be an integer
  - added to the "listing struct"  as shown in the new ABI for `retireCarbonmarkListing()` method
- if there is a token id, assumes it's an ICR contract (ERC1155 multi-token) that has zero decimals
- ask peter for example code on what retirements look like for ICR listings, we should have some in our frontend.
Stretch goal:
- ICR will not be pooled into any toucan or c3 liquidity pools (yet), so assume its only listings
- do preliminary work / thinking on what if a ICR was present in a pool... (this will eventually need to be supported)

- [x] Remove special conditional logic for ICR in `RetireForm.tsx`

Atmosfearful:
No reason we cant start on this now since ICR is coming very soon ™️

sprrwhwk:
@Atmosfearful 
>       * added to the "listing struct"  as shown in the new ABI for `retireCarbonmarkListing()` method

It's a new function in the ABI so we need to call a different function for ICR retirements instead of tweaking the params.

sprrwhwk:
@psparacino how can we get a token id for the api here? Say I need the token id for the following: https://staging-api.carbonmark.com/projects/ICR-4-2020

Atmosfearful:
Since we've got the tokenId added to the API and the relevant Offsetra changes put on staging, we should be able to get a PR up on the frontend and close this out

PR #2188, opened 2024-01-29
add tokenId to listings

sprrwhwk:
## Description

This PR adds a `tokenId` field to our listings endpoint. This is needed for ERC1155 retirements which we're introducing alongside support for the IC registry (ICR).

## Related Ticket

Implements the api part of #2159


carbonmark-api/test/fixtures/tokens.ts
```diff
@@ -34,6 +34,7 @@ const prices: GetPoolPricesQuery["prices"] = [
 
 export const token = {
   symbol: "TCO2-VCS-1764-2020",
+  id: "1",
 };
```
psparacino, line 38: This token isn't an ERC1155 and thus won't have a tokenId.

carbonmark-api/test/routes/listings/[id]/get.test.ts
```diff
@@ -15,10 +15,25 @@ describe("GET /listings/:id", () => {
       console.error("/listings/:id get.test.ts setup failed", e);
     }
   });
+
```
Comment (psparacino) [inline carbonmark-api/test/routes/listings/[id]/get.test.ts:18]: Might better to abstract this test to 'is an ERC1155' as there could be other non-ICR 1155 projects in the future

psparacino (changes requested):

carbonmark-api/test/fixtures/tokens.ts
sprrwhwk: it's tricky because the source of this is called "token" in the subgraph, whereas the other comes from carbon credits but it refers to the tokenId.
I have moved it out in [645c800](https://github.com/KlimaDAO/klimadao/pull/2188/commits/645c800b440a1bb93ae34fdf5773725e6e9d8c10)

sprrwhwk: this is now obsolete due to the subgraph changes where we now have tokenId as part of a listing

carbonmark-api/test/routes/listings/[id]/get.test.ts
Comment (sprrwhwk) [reply carbonmark-api/test/routes/listings/[id]/get.test.ts]: this is now obsolete due to the subgraph changes where we now have tokenId as part of a listing

carbonmark-api/src/app.constants.ts
```diff
@@ -12,7 +12,7 @@ const GRAPH_API_ROOT_ID = "https://api.thegraph.com/subgraphs/id";
  * This is also the case for SANITY_URLS
  */
 const POLYGON_URLS = {
-  marketplace: `${GRAPH_API_ROOT_ID}/QmXrzcwG5b31hE1nDzT5NCAiHfM9stwMLqs8uk9enJiyPf`,
+  marketplace: `${GRAPH_API_ROOT_ID}/QmaHkN47zUB3b1xnfeqCwatG5DRtMuitEDf9FdFkq5V3Fr`,
```
biwano, line 15: We have to change this again?

sprrwhwk: Yes, we have added the tokenId to listings

0xemc (approved):

biwano (approved):

Issue #2136, opened 2024-01-16
[Quantites/Amounts] Standardize amount/quantity handling discovery


psparacino:
**Is your feature request related to a problem? Please describe.**
All registries, up until ICR, have only supported ERC20s (18 decimals). ICR credits are currently regular integers and thus the amounts need to be handled differently by both api and UI.

The current method for ERC20s has been to keep amounts in wei as long as possible until necessary to display to the UI. Currently although ICR quantities are regular integers, they are also kept in their original state as long as possible, to facilitate approvals and contract interactions, unless necessary to interact with other retirement data from ERC20s.

Additionally, ICR is considering switching to an 18 decimal format. It's also possible if using ERC1155s that different tokens within one project or registry have different decimals (not 'natively' supported by an 1155 but stored in the token uri), although this more of an edge case.

**Describe the solution you'd like**
Amounts can either be standardized per registry or per credit. Per credit token would probably head off any debt down the road although it might address edge cases that will never exist. 

If decimals were included with every credit token address, then it'd be possible to always associate with amounts from a credit with a decimal value. This could be possible on the digital-carbon subgraph by creating a token entity for 1155s as well and associating that token entity with a credit.

**Describe alternatives you've considered**
Assuming all registries only ever use one token type a basic mapping could also be used in the api and fe to format amounts based on registry.

**Additional context**

PR #2189, opened 2024-01-29
[api] Standardize quantity/amount handling

psparacino:
## Description

Clean up from the ICR integration: 

ICR handles their quantities in regular integers (0 decimals). This was handled in the previous integration by ternaries where applicable. This PR uses a util function for that purpose instead to increase maintainability (ICR is considering going to the standard 18 decimals).

## Related Ticket

Addresses one half of this issue. The FE needs this cleanup as well: [2136](https://github.com/KlimaDAO/klimadao/issues/2136)

## How to Test

1.  Any buy/sell/retirement quotes
2. `purchases/`
3. `activities/`

## Notes For QA

There should be no change in behavior from the current staging; this clean-up refactoring only

### DIFF
--- a/carbonmark-api/package.json
+++ b/carbonmark-api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@klimadao/carbonmark-api",
-  "version": "6.1.0",
+  "version": "6.1.1",
   "description": "An API for exploring Carbonmark project data, prices and activity.",
   "main": "app.ts",
   "scripts": {
--- a/carbonmark-api/src/utils/marketplace.utils.ts
+++ b/carbonmark-api/src/utils/marketplace.utils.ts
@@ -1,5 +1,7 @@
 import { utils } from "ethers";
+import { formatUnits } from "ethers/lib/utils";
 import { compact } from "lodash/fp";
+import { REGISTRIES } from "../../src/app.constants";
 import {
   GetProjectsQuery,
   Listing,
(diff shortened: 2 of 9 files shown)


carbonmark-api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@klimadao/carbonmark-api",
-  "version": "5.3.5",
```
Atmosfearful, line 3: about to merge 5.3.6 so this will need to be incremented again

Atmosfearful (changes requested):

carbonmark-api/src/utils/marketplace.utils.ts
```diff
@@ -32,6 +34,17 @@ export type GetProjectListing = NonNullable<
...
+/** Currently all registries use 18 decimals except ICR, which uses 0  */
+
+export const formatAmountByRegistry = (registry: string, quantity: string) => {
+  if (registry === REGISTRIES.ICR.id) {
+    return formatUnits(quantity, REGISTRIES.ICR.decimals);
+  } else {
+    return formatUnits(quantity, 18);
+  }
+};
+
```
sprrwhwk, line 47: This seems a little unintuitive to me. Can we do something like this instead?

```suggestion
export const formatAmountByRegistry = (registryId: string, quantity: string) => {
  return formatUnits(quantity, REGISTRIES[registryId].decimals);
};
```

When we later add another registry, we won't have to update this function. We just need to add the registry with it's decimals to `REGISTRIES`.

sprrwhwk (changes requested):
The conditional logic for ICR can easily be overlooked when adding another registry in the future.

carbonmark-api/src/utils/marketplace.utils.ts
psparacino: Good call, revised

sprrwhwk (approved):
Nice 💪

0xemc (approved):

biwano (approved):

PR #2245, opened 2024-02-07
[API] Update subgraph version to include `UpdatedExpiration` event

### LINKED ISSUE #2243: [listings] Expired (?) listings not showing on Profile page
See this [discord thread](https://discord.com/channels/1135827610434289727/1204433909060730900) for further context,

logged in as https://www.carbonmark.com/users/jabby

No active listings??
Screenshot 2024-02-06 at 2 14 25 PM

In my Portfolio I have 3 assets that show Listed tonnes:
Screenshot 2024-02-06 at 2 15 43 PM
Screenshot 2024-02-06 at 2 15 56 PM
Screenshot 2024-02-06 at 2 16 08 PM

Another side note/issue is the Stats component says I have 4 Active listings??
image

@psparacino did a quick investigation and it looked like the listings were expired. However, an expired listing should not just disappear, they should have an EXPIRED label on them.

Related issues from initial implementation:
#1525 
#1655

psparacino:
## Description

The listings issue described below appears to be a casing issue in the frontend.  

This updates the subgraph (current version breaks on activities query as there are null projects for some entities due to an uncaught case) to handle expiration date only updates.

## Related Ticket

Related to #2243 

## How to Test

1. got to https://carbonmark-git-tufnel-fix-listings-not-showing-klimadao.vercel.app/users/jabby
2. confirm expired listings are there


jabby09 (commented):
@psparacino expired listings are now showing 

> listings will show for 90 days after they expire
Don't totally understand this.
Do you mean they will show as expired listing on my Profile page?
Why only for 90 days and what happens after that?

0xemc (changes requested):
@psparacino @biwano 

I'm not sure this fixes the issue. My understanding is that the user should always see their own expired listings even after 90 days. 

Can we not just provide a large value for expiresAfter on the call to the api from the consumer (app)

jabby09:
> @psparacino @biwano
> 
> I'm not sure this fixes the issue. My understanding is that the user should always see their own expired listings even after 90 days.
> 
> Can we not just provide a large value for expiresAfter on the call to the api from the consumer (app)

@0xemc correct this is not completely fixed, Peter is working on it https://discord.com/channels/1135827610434289727/1204433909060730900/1204862568926281848

carbonmark-api/src/routes/users/[walletOrHandle]/get.utils.ts
```diff
@@ -114,7 +114,8 @@ export const getUserByWallet = async (params: {
 }) => {
   const sdk = gql_sdk(params.network);
   const expiresAfter =
-    params.expiresAfter || Math.floor(Date.now() / 1000).toString();
+    params.expiresAfter ||
+    Math.floor((Date.now() - 90 * 24 * 60 * 60 * 1000) / 1000).toString();
   const { listings, activities } = await sdk.marketplace.getUserByWallet({
     wallet: params.address.toLowerCase(),
     expiresAfter,
```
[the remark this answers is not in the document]
Comment (psparacino) [reply carbonmark-api/src/routes/users/[walletOrHandle]/get.utils.ts]: I adjusted back to the earlier expiresAfter default as the hooks in the components are passing the right values for expiresAfter to show vs hide expired listing based on user connection. This version was showing expired listings even if an unconnected user. 

Not sure what happened that cause the initial error, if it was a subgraph issue or something else

biwano (approved):

carbonmark-api/src/.generated/types/marketplace.types.ts
```diff
@@ -1214,6 +1223,11 @@ export type SubscriptionUsersArgs = {
   where: InputMaybe<User_Filter>;
 };
 
+export enum TokenStandard {
+  ERC20 = 'ERC20',
+  ERC1155 = 'ERC1155'
+}
+
```
biwano, line 1230: Is it used anywhere?

psparacino: not yet. but I included it on the subgraph so we have a more reliable way of checking the tokenStandard in the future to handle versus by ids or registry

0xemc (approved):

carbonmark-api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@klimadao/carbonmark-api",
-  "version": "6.1.2",
+  "version": "6.2.1",
```
0xemc, line 3: please run `npm version minor` from the carbonmark-api dir (or run `npm install`)

0xemc (commented):
Is this PR still fixing the missing listing just by upgrading the target subgraph (rather than changing the expired listings limit)?

If so we should update the title and description

0xemc:
@jabby09 @0xtapi can we give this PR the greenlight?

jabby09:
> @jabby09 @0xtapi can we give this PR the greenlight?

hey @0xemc I _believe_ this is working now but I'm doing some final FE testing on this PR https://discord.com/channels/1135827610434289727/1204433909060730900/1206684271616598057

psparacino:
@jabby09 The reverting calls on those expired listing occurred because those listings had actually already been deleted by the contract due to having less than the minimum fill amount. There was no event emitted so the subgraph didn't pick it up. The issue was fixed and new events emitted to update the subgraph so that should be separate from this issue

jabby09 (approved):

0xemc:
Closing after ticks from all parties

PR #2248, opened 2024-02-08
Fix API Preview Deployments

0xemc:
## Description

Ensure preview deployments still go ahead even if the health check fails


biwano (approved):
Perfect!

PR #2250, opened 2024-02-09
Remove ICR API from filters (and carbonmark-api)

psparacino:
## Description

Now that all ICR projects are on the subgraph, the countries, vintages and methodologies for those projects are on the subgraph and we no longer need to query ICR's api for the available filters.

This PR removes the last remaining piece of the ICR api from `carbonmark-api` so that all ICR sources, (except querying user ICR holdings) are either from the CM cms or subgraphs.

## How to Test

1. On the preview link go to `/projects`
2. Filter by 'Spain'
3. The only result is an ICR project

## Notes For QA

Other notes:
- There shouldn't be any change in behavior, as the ICR api calls for filters had become redundant


carbonmark-api/src/app.constants.ts
0xemc, line 1: As @biwano often mentions we should use the `npm version patch` command or run `npm install` after changing this to make sure that the package-lock is updated correctly

biwano (approved):
Nice.
Can you update the version package-lock.json too?
To do so when you update you can use the command `npm version patch` in the `carbonmark-api` folder. It updates both the `package.json` and the `package-lock.json`.

biwano (changes requested):
I connected carbonmark to the preview API in vercel and it made the PR crash.

carbonmark-api/src/app.constants.ts
psparacino: hm that's only updating the package.json and not package-lock for me.

biwano (approved):

PR #2237, opened 2024-02-06
Fix retirement amount formatting

psparacino:
draft for comparision

PR #2261, opened 2024-02-13
[ICR listing] Fix update listing minFillAmount for different decimals

psparacino:
## Description

The updateListing function was parsing the default min fill amount (1) for every update. This is correct for ERC20s but incorrect for ICR credits. 

This issue has now been addressed on the contract level as well and breaking minFillAmounts (if a min fill is > than the actual amount) will revert.  To correct this issue on the frontend so that the correct params are passed the default min fill needs to be adjusted for different decimal amounts.

The next iteration will format these amounts by registry similar to #2237 for the api versus using ternaries.

## How to Test

1. Update an ICR credit listing with either a price or quantity change
2. The update transaction should successfully update without any contract errors or reverts

### DIFF
--- a/carbonmark/lib/actions.ts
+++ b/carbonmark/lib/actions.ts
@@ -274,6 +274,9 @@ export const updateListingTransaction = async (params: {
     const amount = params.projectId.startsWith("ICR")
       ? params.newAmount
       : parseUnits(params.newAmount, 18);
+    const minFillAmount = params.projectId.startsWith("ICR")
+      ? DEFAULT_MIN_FILL_AMOUNT.toString()
+      : parseUnits(DEFAULT_MIN_FILL_AMOUNT.toString(), 18);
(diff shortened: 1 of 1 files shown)


biwano (commented):
Are there other contract calls where this should be done too?

psparacino:
@biwano `createListing` is the only other marketplace contract function that requires a minFill, where this conversion is already included. The other two marketplace functions that require more than the id to call (ie. registry dependent decimals) both format amounts already:

`createListing`: https://github.com/KlimaDAO/klimadao/blob/staging/carbonmark/lib/actions.ts#L328-L339 
&
`fillListing`: https://github.com/KlimaDAO/klimadao/blob/staging/carbonmark/lib/actions.ts#L328-L339

jabby09 (approved):

biwano (approved):

0xemc (approved):

PR #2264, opened 2024-02-14
cm: web3 hook returns lowercase address

psparacino:
## Description

The useWeb3 hook is returning a checksum address while the subgraph returns only lowercase addresses.

This is causing a mismatch in several areas where hooks are not returning the correct listings, which leads to allowance errors etc.

There should either be consistency to checksum or lowercase.  Given the subgraph only uses lowercase it would probably be better to have that be the st

## Related Ticket

Closes <issue-number>
Also related to <issue-number>

## How to Test

1. _step1_
2. _step2_
3. _step3_

## Notes For QA

- 

Relevant preview URLs:
- 

Other notes:
-

### DIFF
--- a/lib/utils/useProvider/index.tsx
+++ b/lib/utils/useProvider/index.tsx
@@ -150,7 +150,7 @@ export const useProvider = (
       const newState: ConnectedWeb3State = {
         provider,
         signer,
-        address,
+        address: address.toLowerCase(),
         network,
         networkLabel: isTestnetChainId(network.chainId) ? "mumbai" : "polygon",
         isConnected: true,


carbonmark/components/pages/Users/SellerConnected/index.tsx
```diff
@@ -41,7 +41,8 @@ export const SellerConnected: FC<Props> = (props) => {
     props.userAddress,
     {
       network,
-      expiresAfter: address === props.userAddress ? "0" : undefined,
+      expiresAfter:
+        address?.toLowerCase() === props.userAddress ? "0" : undefined,
```
0xemc, line 45: for safety's sake please convert both to lowercase explicitly

```suggestion
      expiresAfter:
        address?.toLowerCase() === props.userAddress.toLowerCase() ? "0" : undefined,
```

0xemc (commented):
``

0xemc (approved):

jabby09 (approved):

PR #2265, opened 2024-02-14
[listings] Update to use subgraph version with updatedExpiration event

psparacino:
## Description

The marketplace subgraph previously didn't handle the situation where only the expiration time was updated without price or quantity changes. 

This PR uses api version 6.2.0 which uses the updated subgraph (`QmT2YkGwTqvek54VsGkfurGLrGGKgkwcK6TLotvkjpF4Mz`) that handles this case.

This PR should correctly display listings updated without price or quantity changes as current and for the activity sidebar to include this activity as Updated Expiration

## Related Ticket

Also related to #2243 

## How to Test

1. Check Profile for listings updated without price or quantity changes. They should be current listings.


carbonmark/components/Activities/Activity.tsx
```diff
@@ -84,6 +86,26 @@ export const Activity = (props: Props) => {
       props.activity.amount && `${formatToTonnes(props.activity.amount)}t`;
   }
 
+  /** Expiration Labels */
+  if (isUpdateExpiration) {
+    if (
+      props.activity.timeStamp !== null &&
+      props.activity.timeStamp !== undefined
+    ) {
+      // The expiration is on the listing entity, but not activity. In order to avoid more extensive refactoring, 90 days is added the expiration here.
```
0xemc, line 95: Hmm we need to be careful here. 

Is the 90 days an unlinked external dependency? (i.e is 90 days hard coded somewhere else?)

psparacino: revised with DEFAULT_EXPIRATION_DAYS in place of the hardcoded 90

carbonmark/.generated/carbonmark-api-sdk/client.ts
```diff
@@ -46,14 +46,21 @@ export const fetchClient = async <
...
+      body: JSON.stringify(request.data),
+      headers: {
+        "Content-Type": "application/json",
+        ...request.headers,
+      },
+    });
+  } catch (e) {
+    console.error(`Fetch error: ${url}`);
+    throw e;
+  }
```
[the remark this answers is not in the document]
psparacino: I generated these from api v6.2.0; the one with the updated subgraph. 

So this will match with that version?

jabby09 (commented):
@psparacino @0xemc 

this listing I created yesterday (Feb 14) so the Expiration would have been set to 5/14/2024

today when I edit the listing, the Expiration is showing 5/15/2024 which is what it would be updated to if I did update the listing

However this should be the actual expiry date of the listing from the last update

but now that I'm thinking about this, I wonder if that is how it was built

Screenshot 2024-02-15 at 12 43 45 PM

jabby09 (approved):

Atmosfearful (approved):
Code looks good, ran a create, an edit, and an update-expiration txn through and they all worked. 
image

PR #2274, opened 2024-02-15
[provenance] fix layout issues with GridContainer

### LINKED ISSUE #2273: [provenance] provenance page layout issue
Layout of provenance page needs some fixing after the Nav header PR 

See: [https://www.carbonmark.com/retirements/0x087a7afb6975a2837453be685eb6272576c0bc06/6/provenance](https://www.carbonmark.com/retirements/0x087a7afb6975a2837453be685eb6272576c0bc06/6/provenance)

image

0xMakka:
## Description

`GridContainer` is used for the page layouts and after the recent navigation changes, the page layout has changed. Grid container by default is setting `display: grid`, in this instance using `display: unset` resolves the page issues. 

Note: I'd like to relook at these pages that use GridContainer in a follow up ticket and see can they now be replaced with the `Container` component from material-ui. I've briefly checked if it works, it does apart from the footer needs some work to  span the full screen.

## Related Ticket

Closes #2273

## How to Test

1. Ensure the layout isn't breaking on desktop/mobile.
2. Ensure the retirements page layout is also working as expected.

## Screenshots:

Screenshot 2024-02-15 at 15 33 20

Screenshot 2024-02-15 at 15 33 34

### DIFF
--- a/carbonmark/components/pages/Retirements/RetirementProvenance/ProvenanceComponent/index.tsx
+++ b/carbonmark/components/pages/Retirements/RetirementProvenance/ProvenanceComponent/index.tsx
@@ -60,7 +60,7 @@ export const ProvenanceComponent = (props: ProvenanceComponentProps) => {
             <BalanceOutlined fontSize="large" />
             <Trans>Amount</Trans>
           </Text>
-          <Text t="h5">
+          <Text t="h5" align="end">
             <Trans>{formattedAmount} Tonnes</Trans>
           </Text>
         </div>
--- a/carbonmark/components/pages/Retirements/RetirementProvenance/ProvenanceTimeline/styles.ts
+++ b/carbonmark/components/pages/Retirements/RetirementProvenance/ProvenanceTimeline/styles.ts
@@ -1,4 +1,5 @@
 import { css } from "@emotion/css";
+import breakpoints from "@klimadao/lib/theme/breakpoints";
 
 export const timeline = css`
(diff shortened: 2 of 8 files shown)


0xtapi:
Almost there, but I did find something:
bug

Layout breaks (wrong alignments, text overflow) when the browser window is too small. To reproduce, make your browser window small enough so CM switches to mobile layout.

(This PR also fixes the retirements page btw, which is broken in the same way the provenance page is.)

0xMakka:
> Almost there, but I did find something: bug
> 
> Layout breaks (wrong alignments, text overflow) when the browser window is too small. To reproduce, make your browser window small enough so CM switches to mobile layout.
> 
> (This PR also fixes the retirements page btw, which is broken in the same way the provenance page is.)

Thanks @0xtapi I thought I had covered all bases with the initial, apologies. I've made some more changes to hopefully address the issues mentioned, it should be good to retest again.

0xtapi:
One more small fix and this is good to go:
Screenshot 2024-02-15 174846

"1 tonnes" should be aligned to the right side of that element, right? It does seem to do that for large numbers (like here: https://carbonmark-git-makka-2273-fix-page-layout-issues-klimadao.vercel.app/retirements/0x65a5076c0ba74e5f3e069995dc3dab9d197d995c/8/provenance) but not small ones.

0xemc (approved):

PR #2276, opened 2024-02-15
[Carbonmark] Fix retire from listing quantites

psparacino:
## Description

The original values for ICR were passing parsed values which led to the contract rejecting the quantities.

These values have been adjusted for ICR credits.

## How to Test

1. `Retire Now` from one of these listings: https://www.carbonmark.com/projects/ICR-112-2019
2.  The txn should complete successfully

### DIFF
--- a/carbonmark/lib/actions.retire.ts
+++ b/carbonmark/lib/actions.retire.ts
@@ -216,17 +216,24 @@ export const retireCarbonTransaction = async (params: {
 
     let txn;
     if (params.product.type === "listing") {
+      const isICR = params.product.project.id.startsWith("ICR");
+      const leftToSell = isICR
+        ? params.product.leftToSell
+        : parseUnits(params.product.leftToSell, 18);
+      const quantity = isICR
+        ? params.quantity
+        : parseUnits(params.quantity, 18);
+
       txn = await retireContract.retireCarbonmarkListing(
         [
           params.product.id,
           params.product.seller.id,
           params.product.tokenAddress,
-          parseUnits(params.product.leftToSell, 18),
+          leftToSell,
           parseUnits(params.product.singleUnitPrice, getTokenDecimals("usdc")),
         ],
         parsedMaxAmountIn,
-        // TODO: ICR tokens have zero decimal places
-        parseUnits(params.quantity, 18),
+        quantity,
         "",
         params.beneficiaryAddress || params.address,
         params.beneficiaryName,


Atmosfearful (approved):

psparacino:
Converted to draft; further changes needed on this

Atmosfearful (approved):

jabby09 (changes requested):
worked for a full tonne https://carbonmark-git-tufnel-fix-retire-from-listing-q-2f677a-klimadao.vercel.app/retirements/0x6476edf98Bdc5EE62f9D4E4BF8932d69Ed685483/47

but transaction breaks if you use decimals in quantity so we need to add conditional logic to retire forms

jabby09 (changes requested):
from @0xtapi testing

Some more info: This message seems to pop up only if the first value you enter after accessing the site or refreshing it is a 0. Like this:

Message does NOT appear when the first value you enter on the retirement page of that credit after refreshing the page is anything other than 0. Example: If you enter 2, delete that, enter 0.5, it automatically switches to 1 tonne and does not display the error message. (Good)

Message DOES appear if you enter a 0 first after refreshing the page. And then it won't disappear, even if you switch to a valid value, like 1, 2, 3, etc... It will still let you retire (the button works), but the message will also stay there. (Bad)
So basically... we need to make sure the message is correct (at least 1 tonne, not 0.001) and that it disappears after a valid value has been entered 🤔 It just does not like the user entering 0 after page refresh for some reason

jabby09 (approved):

carbonmark/components/pages/Project/Retire/Pool/RetireInputs.tsx
```diff
@@ -188,6 +188,10 @@ export const RetireInputs: FC<Props> = (props) => {
       : props.product.leftToSell
   );
 
+  const isICR =
+    props.product.type === "listing" &&
```
sprrwhwk, line 192: Why do we need this check? I get the id / registry check but why does pool vs listing matter?

psparacino: If this check is omitted a type error pops up because `project` doesn't exist on the "pool" `product` type.

0xemc: @psparacino @sprrwhwk following the prop drilling back up to /projects/retire/[lsitingId] or /projects/retire/[pool] it looks like project is coming from 

https://github.com/KlimaDAO/klimadao/blob/b784b79139e78b177db2dd6fd0dbbc772523f11e/carbonmark/.generated/carbonmark-api-sdk/clients/getProjectsId.ts#L13

Can we use the server types here instead of the handcrafted ones?

0xemc: Ok moving this conversation to a new ticket

https://github.com/KlimaDAO/klimadao/issues/2293

0xemc (approved):

PR #2312, opened 2024-02-20
Add country type to carbon projects CMS for tracking centroids and isRemoval boolean for methodologies

0xAurelius:
## Description

Add Country type to track centroids since Puro projects only specify country and we still want to map their locations (even if roughly)

Stretch goal would be to show full country boundaries, but lets start with centroids only first

### DIFF
--- a/carbon-projects/schemas/country.ts
+++ b/carbon-projects/schemas/country.ts
@@ -1,0 +1,42 @@
+import { defineField, defineType } from "sanity";
+
+export default defineType({
+  name: "country",
+  title: "Country",
+  description: "Country metadata",
+  type: "document",
+  preview: {
+    select: {
+      slug: "id",
+      name: "name",
+    },
+    prepare(selection) {
+      return {
+        title: selection.slug.current || "",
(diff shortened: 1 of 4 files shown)


carbon-projects/schemas/country.ts
```diff
@@ -0,0 +1,41 @@
...
+        title: selection.slug.current || "",
+        subtitle: selection.name || "",
+      };
+    },
+  },
+  fields: [
+    defineField({
+      name: "id",
+      type: "slug",
+      description: "Unique two-digit ISO code identifier for this country e.g. 'AU'",
```
sprrwhwk, line 24:
```suggestion
      description: "Unique two-letter ISO code for this country e.g. 'AU'",

```

sprrwhwk (commented):
Where do we host these type documentations?

0xAurelius:
> Where do we host these type documentations?

We don't really, they end up as in-line documentation in the Sanity interface like these: 
image

0xAurelius:
Closed in favor of #2327

PR #2318, opened 2024-02-21
2254 token usdc vs usdce clarificationsupport

### LINKED ISSUE #2254: [token] USDC vs USDC.e clarification/support
**Describe the bug**
Circle recently issued native usdc: https://www.circle.com/blog/native-usdc-now-available-on-polygon-pos

vs the previous bridged version. The frontend only supports the bridged version: [0x2791bca1f2de4661ed88a30c99a7a9449aa84174](https://polygonscan.com/token/0x2791bca1f2de4661ed88a30c99a7a9449aa84174) which is identified on Carbonmark as USDC. 

So if a user has the new native USDC they will see a zero balance for their USDC (as the balance is actually querying USDC.e).

Carbonmark should either specify USDC.e or support both tokens for payment.

**To Reproduce**
1. If you have native USDC, check your USDC balance in Carbonmark vs wallet balance

**Expected behavior**
Either both tokens should be supported or USDC clearly labeled USDC.e with an explanation.

**Screenshots**
Screenshot from user:
image

ChazSchmidt:
## Description
The original USDC token has been changed to USDC.e on the Polygon blockchain. This PR clarifies that we currently support USDC.e by changing the UIs of Carbonmark and Klima app to display USDC.e wherever USDC was previously displayed. Mentions of USDC in the API descriptions has also been changed to USDC.e for clarification.

## Related Ticket

Closes #2254

## Notes For QA

- Wherever USDC was displayed i.e. payment dropdown menus
usdce-issue2254


ChazSchmidt:
> Still USDC here in headline: https://carbonmark-git-2254-token-usdc-vs-usdce-clarifi-81c4f8-klimadao.vercel.app/blog/how-to-get-usdc-or-matic

:sweat_smile: oops, how'd I miss that? fixed.

> Search function is not working and it shows "resources.form.categories.subheader" to the left under "Categories". Not sure if this is related to this PR. Can changing text from USDC to USDC.e break a search function?

The string text errors in Resources are now fixed after a fresh lingui extract + compile. Correct me if I'm wrong. I don't think the search works on Vercel previews.

PR #2319, opened 2024-02-21
Reduce default min fill amount and validate with subgraph minFill

### LINKED ISSUE #2317: [form] Reduce and enforce DEFAULT_MIN_FILL_AMOUNT
**Describe the bug**
Change it from 1 to 0.001
Make sure ICR doesn't break (icr can't go below 1)
Make sure we are only using this value to set the min fill during listing creation, in other parts of the application we need to read the value provided for the actual listing (because old listings will still be 1, and future listings might someday be even more than that)
Use the minFillAmount value to validate the form here and prevent the user from trying to retire a value < the listing's min fill
https://www.carbonmark.com/projects/VCS-1764-2020/retire/0x734db5a54215c163b1372fb1cc860f40d6a90ccea89731433b9ca48cdc358039

psparacino:
## Description

Adjusts default minFill to .001 so new listings will have that default minFill for purchases or retires.

Retires and Purchases use the minFill amount on the subgraph instead of hardcoded version either in the component or a constant.

one minor issue: the format from the subgraph has on decimal place. so it used to show "1" but now will show "1.0". Should that be trimmed?

## Related Ticket

Closes #2317 

## How to Test

1. Test a fractional retire on a current listing (should not validate as default min is 1)
2. Create a new listing 
3. Do the same as above (new minFill amount should be .001 for new listing)
4. Do the same above for a purchase

### DIFF
--- a/carbonmark/components/pages/Project/Purchase/Listing/PurchaseInputs.tsx
+++ b/carbonmark/components/pages/Project/Purchase/Listing/PurchaseInputs.tsx
@@ -55,7 +55,6 @@ export const PurchaseInputs: FC<Props> = (props) => {
               <Trans>Available: {Number(props.listing.leftToSell)}</Trans>
             </Text>
           </div>
-
           <InputField
             id="amount"
             inputProps={{
@@ -70,8 +69,8 @@ export const PurchaseInputs: FC<Props> = (props) => {
                   message: t`Amount is required`,
                 },
                 min: {
-                  value: 1,
-                  message: t`The minimum amount to buy is 1 Tonne`,
+                  value: Number(props.listing.minFillAmount),
+                  message: t`The minimum amount to buy is ${props.listing.minFillAmount} Tonne`,
                 },
                 max: {
                   value: Number(props.listing.leftToSell),
--- a/carbonmark/components/pages/Project/Retire/Pool/RetireForm.tsx
+++ b/carbonmark/components/pages/Project/Retire/Pool/RetireForm.tsx
@@ -455,10 +455,10 @@ export const RetireForm: FC<Props> = (props) => {
   );
 };
 
-function isPool(product: Product): product is PoolProduct {
+export function isPool(product: Product): product is PoolProduct {
   return product.type === "pool";
 }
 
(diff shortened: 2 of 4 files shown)


0xtapi:
Spotted some odd behavior. These three links all point to listings of the same credit:

https://carbonmark-git-fix-reduce-default-min-fill-amount-klimadao.vercel.app/projects/VCS-191-2008/retire/0x9e0113c4826fcdeb1a7b12ea41c2441aa77d230444836d526b744da866724d3a

https://carbonmark-git-fix-reduce-default-min-fill-amount-klimadao.vercel.app/projects/VCS-191-2008/retire/0x808ff5bb0301bbfa99fe99713ec99be6f26d6848bff9b67bdb900092046d4a3d

https://carbonmark-git-fix-reduce-default-min-fill-amount-klimadao.vercel.app/projects/VCS-191-2008/retire/0x1b6b75dd7f00ff26d71f87e4c491820dee12277e39ab1f1a7ea7c40d9e8495a5

First listing is from early December, second from mid January, third from mid February, so they are all old listings.

First and third link return errors when I try to retire fractional tonnes (as they should) and the Retire button doesn't do anything. But the second link does not show the error message, and I can click Retire. When I try to approve the transaction, this happens:

Screenshot 2024-02-23 142351

But I shouldn't even get to that point because the UI should block my attempt to retire a fractional tonne...

Also, the error message isn't quite there yet:
Screenshot 2024-02-23 141211

psparacino:
good catch, for some reason that middle listing has a default min fill of .001 coming from the subgraph. 

Not sure how that could have slipped through i.e. you updated that on the 2/15 and it somehow set the minFill to 1 where the previous listing had a minFill of .001, which shouldn't have happened to begin with
https://polygonscan.com/tx/0xb4dedb60ab32c106999000d0c09b6d9c54eceda4688825f228566fc6d88d8135#eventlog

even though that's kind of weird that shouldn't be an issue as the subgraph should pick up that and use that as the minFill for the frontend

psparacino:
ah ok, the subgraph isn't handling minFill amounts for updating Listings

PR #2327, opened 2024-02-27
Add country type to carbon projects CMS for tracking centroids

0xAurelius:
## Description
Needed because Puro projects only have a country not a geolocation

Will need fallback logic to read from this new `countryDetails` field if the project location is missing


Atmosfearful (approved):
nice solution!

0xemc (approved):

PR #2348, opened 2024-04-23
Fix Puro registry ID

0xAurelius:
## Description

Should be "PUR" instead of "PURO" to match Toucan's labeling scheme


Atmosfearful (approved):
Ty we will need to fix this in the API and FE regex as well 👻

PR #2349, opened 2024-04-24
Migrate subgraph URLs to decentralized network

0xAurelius:
## Description

DO NOT MERGE UNTIL ALL SUBGRAPHS ARE UPGRADED!


0xAurelius:
So I actually believe `subgraphs` is unused on the main branch... it was only used for the offset interface which is now legacy code and being rebuilt as a sep app

I'm going to update this PR to simply remove the subgraph URLs entirely

PR #2366, opened 2024-05-27
Switch away from using old one-off subgraph

0xAurelius:
## Description
Should not be using this one-off subgraph, as of https://github.com/KlimaDAO/klima-subgraph/pull/130/files this functionality is now available in the main one


0xAurelius:
Apparently this functionality moved and now we just have cleanup to do, will fold that cleanup into the subgraph migration PR

PR #2373, opened 2024-06-24
Makka/deposit carbon view

### LINKED ISSUE #2368: [deposit] Dev implementation in the dApp
### Designs
https://www.figma.com/design/oA67X2DpUOjlpEktwaDZH7/Deposit?node-id=0-1&t=N3wez6P2Tw2so6xj-0

### Requirements

- [x] support for desktop and mobile
- [x] new url path https://app.klimadao.finance/#/deposit
- [x] need to show Risk Disclaimer if user enters site from the /deposit URL
- [x] new `Deposit Carbon` option in side menu
- [x] new [Deposit carbon form](https://www.figma.com/design/oA67X2DpUOjlpEktwaDZH7/Deposit?node-id=23-1249&t=N3wez6P2Tw2so6xj-4)
- only available for TC02 --> BCT 
- only editable field is Deposit TC02
- MAX button fills in field with value from Available balance
- need to [LOGIN to CONTINUE](https://www.figma.com/design/oA67X2DpUOjlpEktwaDZH7/Deposit?node-id=23-1293&t=N3wez6P2Tw2so6xj-4)
- [x]  Select token [dropdown modal](https://www.figma.com/design/oA67X2DpUOjlpEktwaDZH7/Deposit?node-id=23-1425&t=N3wez6P2Tw2so6xj-4)
- lists all tokens in wallet
- indicates which tokens are compatible with BCT and greys out those as selectable
- user clicks token and that closes the modal and populates token to deposit on the form
- [x] [Error states](https://www.figma.com/design/oA67X2DpUOjlpEktwaDZH7/Deposit?node-id=23-1337&t=N3wez6P2Tw2so6xj-4)
- form field checks to ensure user has enough available balance for deposit quantity entered

### Notes
- noticing now we do not have designs for what happens after user clicks CONTINUE; but I assume there would be wallet interactions
- we can follow up on this once Codey is back from vacation
- for now can we follow the pattern used in other transactions (retire and buy) in the dApp?

0xMakka:
## Description

New view added to the dApp which allows users to deposit TCO2 tokens in exchange for BCT tokens 1-1.

## Related Ticket

Closes #2368 

## How to Test

1. Navigate to the deposit view. Connect a wallet which contains some TCO2 tokens which you wish to exchange for BCT.
2. The first valid token in the list of user tokens should be shown on the UI and the amount available next to it. 
3. Click on the token in the UI and you should see a popup/modal with all available tokens visible. If a token is not eligible i.e. is a C3T token, you should still see it in the popup but should not be able to select it.
4. Enter the amount you wish to exchange, if the user enters more than what is available you shouldn't be able to continue and the text should change to the color red like the figma designs.
5. If the amount entered is valid, you should be able to proceed to the approval step. Approve the token and submit (similar to the other flows across the dapp).


0xMakka:
@0xtapi @jabby09 this is ready for QA whenever you guys have a chance to take a look.

0xtapi:
Works well, could not find anything. Tried the conversion with two different TCO2s, worked as expected. Tested on Desktop and Mobile.

Comparison to Figma designs checks out as well 👍

jabby09 (approved):

app/actions/deposit.ts
```diff
@@ -0,0 +1,98 @@
...
+}) => {
+  try {
+    const contract = new Contract(
+      params.selectedToken.token.id,
+      IERC20.abi,
+      params.provider.getSigner()
+    );
+    const value: BigNumber = await contract.allowance(
+      params.address,
+      addresses["mainnet"]["bct"]
```
RastogiAbhijeet, line 23: Can the network & "bct" be parameterised?

must-be-carbon, line 23: Can "mainnet" & "bct" be parametrised?

0xMakka: Was keeping with the general convention already used across the dApp. Although it should actually be `addresses["mainnet"].bct` to keep with the current naming convention.

app/components/views/Deposit/index.tsx
```diff
@@ -0,0 +1,334 @@
...
+  const insufficientTokens = Number(formattedTokenBalance) < Number(quantity);
+  const hasAllowance = !!Number(allowance.value);
+  const fullStatus: AppNotificationStatus | null = useSelector(
+    selectNotificationStatus
+  );
+  const status = fullStatus?.statusType;
+
+  useEffect(() => {
+    if (!address) return;
+    (async () => await getTokensHoldings())();
```
must-be-carbon, line 68:
```suggestion
    getTokensHoldings(); 
```

IMO, this should work just fine.

Atmosfearful (approved):
Nice!!

app/components/views/Deposit/index.tsx
```diff
@@ -0,0 +1,334 @@
...
+            </div>
+          </div>
+        </div>
+        <ButtonPrimary className={styles.depositButton} {...getButtonProps()} />
+      </div>
+      {showModal && selectedToken && props.isConnected && (
+        <CarbonTokenModal
+          holdings={holdings}
+          onHide={() => setShowModal(false)}
+          onSelect={(token: CarbonToken) => setSelectedToken(token)}
```
must-be-carbon, line 304: nit: 
```suggestion
          onSelect={setSelectedToken}
```

app/components/views/Deposit/styles.ts
```diff
@@ -0,0 +1,186 @@
...
+`;
+
+export const depositButton = css`
+  width: 100%;
+  ${breakpoints.desktop} {
+    width: max-content;
+  }
+
+  &:disabled {
+    color: #767676;
```
must-be-carbon, line 158: maybe it's too late for this, but I genuinely think there should be a 
const color = { } ;  atleast from where all colors can be referenced or 
to take it on step forward a neatly defined theme.. ( Maybe something to think about in future. )

app/components/CarbonTokenModal/styles.ts
```diff
@@ -0,0 +1,49 @@
...
+  border-radius: 0.4rem;
+  font-size: 1.2rem;
+  font-weight: 700;
+  padding: 0.5rem 0.9rem;
+`;
+
+export const tokenText = css`
+  font-weight: 700;
+  font-size: 1.6rem !important;
+  line-height: 2rem !important;
```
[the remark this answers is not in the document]
0xMakka: This part is quite annoying tbh, the base styles have !important on them, meaning we can't just overwrite the styles with css specificity.

app/components/views/Deposit/index.tsx
```diff
@@ -0,0 +1,334 @@
...
+import { AccountBalanceWalletOutlined } from "@mui/icons-material";
+import KeyboardArrowDown from "@mui/icons-material/KeyboardArrowDown";
+import {
+  approveDepositToken,
+  depositTokens,
+  getAllowance,
+} from "actions/deposit";
+import { BalancesCard } from "components/BalancesCard";
+import { CarbonTokenModal } from "components/CarbonTokenModal";
+import { DisclamerModal } from "components/DisclaimerModal";
```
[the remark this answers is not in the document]
0xMakka: Nice catch, didn't even spot this typo.

app/components/views/Deposit/index.tsx
0xMakka: Nice

app/components/views/Deposit/styles.ts
0xMakka: There is already a themes file in the app, but these variables don't exist. They are currently one off values for this screen and this is the reason I didn't create new variables for them. You are right though, these should really be css variables.

PR #2382, opened 2024-07-25
must-be-carbon/feat/2377 add supply endpoint for KLIMA & BCT

must-be-carbon:
## Description

This pull request will re-introduce the /supply API URL for KlimaDAO. 
This end-point can then be used in platforms like CoinMarketCap/Coingecko

## Related Ticket

Closes 2377

## How to Test
1. Open terminal 
2. Run `npm run dev-app` from root dir
3. Open a new terminal
4. Run the following commands to test all cases & update port-no before testing 
port-no. can be seen as an output of step(2) 
```
# fetches total supply for a token
curl -X "GET" "localhost:<port-no>/api/supply?type=total-supply&token=<klima|bct>"

# fetches circulating-supply for a token
curl -X "GET" "localhost:<port-no>/api/supply?type=circulating-supply&token=<klima|bct>"
```

Screenshot below
image

### DIFF
--- a/app/pages/api/supply.ts
+++ b/app/pages/api/supply.ts
@@ -19,13 +19,26 @@ type APIResponse =
       error: string;
     };
 
+// This can be simplified to be just a string[]
+// but IERC20 ABI doesn't have a decimal exposed
+const DEFAULT_TOKENS = [
+  {
+    name: "klima",
+    decimals: 9,
+  },
+  {
+    name: "bct",
+    decimals: 18,
+  },
+];
+
 /**
- *
+ * @description fetch total supply for a token
  * @param req
  * @param res
- * @example GET /api/supply Returns {APIResponse}
- * @example GET /api/supply?type=total-supply Returns number
- * @example GET /api/supply?type=circulating-supply Return number
+ * @example GET /api/supply?token=klima Returns {APIResponse}
+ * @example GET /api/supply?type=total-supply&token=klima Returns number
+ * @example GET /api/supply?type=circulating-supply&token=klima Return number
  */
 export default async function handler(
   req: NextApiRequest,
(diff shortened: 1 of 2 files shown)


app/pages/api/supply.ts
```diff
@@ -0,0 +1,111 @@
...
+  | {
+      totalSupply: number;
+      circulatingSupply: number;
+    }
+  | {
+      error: string;
+    };
+
+// This can be simplified to be just a string[]
+// but IERC20 ABI doesn't have a decimal exposed
```
0xAurelius, line 23: really? weird... the ERC-20 spec includes the decimal function so this should be readable from the contract, maybe we need a more complete IERC20 ABI

for now you can just use this util to get the decimals: https://github.com/KlimaDAO/klimadao/blob/staging/lib/utils/getTokenDecimals/index.ts

app/pages/api/supply.ts
```diff
@@ -0,0 +1,111 @@
...
+    decimals: 18,
+  },
+];
+
+/**
+ * @description fetch total supply for a token
+ * @param req
+ * @param res
+ * @example GET /api/supply?token=klima Returns {APIResponse}
+ * @example GET /api/supply?type=total-supply&token=klima Returns number
```
0xAurelius, line 40: these examples are still using the older long `type` values with the `-supply` suffix - should be fixed to shorter names

app/pages/api/supply.ts
```diff
@@ -0,0 +1,111 @@
...
+            case "total-supply":
+            case "circulating-supply":
+              return res.status(200).send(totalSupplyNum);
+            default:
+              return res.status(400).send({ error: "Invalid type" });
+          }
+        }
+
+        return res.status(200).send({
+          totalSupply: totalSupplyNum,
```
0xAurelius, line 98: I'm confused why we're returning a dict with two keys

This endpoint should always return a flat, single, integer number

0xAurelius (changes requested):

app/pages/api/supply.ts
must-be-carbon: Believe it or not I didn't push the latest changes I made last night. Will do that as well.

app/pages/api/supply.ts
0xAurelius: nah let's just keep it simple - these endpoints only exist for CMC and CG and they always expect a single integer value

must-be-carbon: all g. Will update.

app/pages/api/supply.ts
```diff
@@ -0,0 +1,99 @@
...
+        if (!totalSupply) {
+          return res.status(500).send({ error: "Failed to get total supply" });
+        }
+
+        const totalSupplyNum = Number(
+          formatUnits(totalSupply, getTokenDecimals(targetToken))
+        );
+
+        switch (type) {
+          case "total":
```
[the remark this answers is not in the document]
must-be-carbon: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#taking_advantage_of_fall-through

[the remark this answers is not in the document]
must-be-carbon: Given total & circulating supply return the same value hence .. I have used fall through.

0xAurelius (approved):

Atmosfearful (approved):

PR #2384, opened 2024-08-02
use ditial-carbon-id and not bridged-carbon-id

0xMakka:
## Description

Use correct graph id for digital carbon.


0xMakka:
Bumping this @Atmosfearful - when you have a chance to approve and merge.

PR #2390, opened 2024-09-24
remove v2 from squid router url

0xMakka:
## Description

V2 on squid router no longer looks to be working - this PR updates the url to lowercase token addresses and removes v2 from the url.


0xMakka:
Not required any longer - got included as part of the autocompounder release.

PR #2391, opened 2024-10-02
Restructure schema to catch captions in GraphQL

0xAurelius:
## Description

Captions are not showing up in GraphQL because they are defined inline

Defining a custom type resolves this

### DIFF
--- a/carbon-projects/schemas/projectContent.ts
+++ b/carbon-projects/schemas/projectContent.ts
@@ -76,50 +76,21 @@ export default defineType({
       name: "coverImage",
       description: "Primary cover image to be shown on project page",
       group: "media",
-      type: "image",
-      fields: [
-        {
-          name: "caption",
-          description:
-            "English language caption to show below the image. Can include image attribution if needed.",
-          type: "string",
-        },
-      ],
+      type: "captionImage",
     }),
     defineField({
       name: "satelliteImage",
       description:
         "Mapbox satellite image based on registry lat/long to be shown on project page",
       group: "media",
-      type: "image",
-      fields: [
-        {
-          name: "caption",
-          description:
-            "English language caption to show below the image. Can include image attribution if needed.",
-          type: "string",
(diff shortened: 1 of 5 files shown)


carbon-projects/schemas/projectContent.ts
```diff
@@ -76,50 +76,21 @@ export default defineType({
...
-      type: "image",
-      fields: [
-        {
-          name: "caption",
-          description:
-            "English language caption to show below the image. Can include image attribution if needed.",
-          type: "string",
-        },
-      ],
+      type: "captionImage",
```
biwano, line 79: Will this not break previous coverImages

carbon-projects/schemas/projectContent.ts
```diff
@@ -76,50 +76,21 @@ export default defineType({
...
-      fields: [
-        {
-          name: "caption",
-          description:
-            "English language caption to show below the image. Can include image attribution if needed.",
-          type: "string",
-          placeholder: "Satellite image of the project location",
-        },
-      ],
+      type: "captionImage",
```
biwano, line 86: Will this not break previous satelliteImages?

biwano (approved):
LGTM. Just want to be sure that the other image fields wont break.

carbon-projects/schemas/projectContent.ts
0xAurelius: No, these existing images "just work" because the structure of the underlying data is the same

The only thing that breaks is the array of images - we need to recreate each one as a CaptionImage to get the schema to pick them up

0xAurelius:
I'm not sure why the formatting check is failing... I ran format-all locally, it seems to be an issue with the alphabetical ordering of imports in lib/components/index.ts...

Perhaps related to inconsistent dependency versions between my local setup and GitHub Actions

0xAurelius:
DO NOT merge as-is, breaks the schema

PR #2397, opened 2024-11-21
add auto-compounder button

0xMakka:
## Description

- [X] Remove base logo from buttons and group under a title Base Network and logo (open to suggestions on this if we want to change it).
- [X] Fix the squid router link - was previously pointing to v2 which no longer exists.

## Screenshot

Screenshot 2024-11-21 at 14 24 03


jabby09 (approved):
@0xMakka looks good
autocompounder option 404s at the moment but I assume that will fix itself when we deploy the app

0xMakka:
> @0xMakka looks good autocompounder option 404s at the moment but I assume that will fix itself when we deploy the app

Yep @jabby09 - currently not showing because auto-compounder route doesn't exist on prod, but as soon as it's released this button will link to it.

PR #2413, opened 2025-01-19
Simplify accreditation structure

0xAurelius:
## Description

We don't actually need a reference field for Accreditation, just a simple name and URL

No entities use this yet so shouldn't be a big deal to migrate


0xMakka (approved):
LGTM 👍

PR #2418, opened 2025-01-27
Fix file category values

0xAurelius:
## Description

Per request from @0xMakka to make FE work simpler


0xMakka (approved):
Nice, thanks @0xAurelius

PR #2420, opened 2025-01-29
[CM] align file categories with figma designs

0xMakka:
## Description

After a discussion with @jabby09 - the file categories need to be aligned with the current[ figma designs ](https://www.figma.com/design/5GewbdVaM1Q06CwHmF7d3F/Projects?node-id=4191-40749&t=ExoFkw01gVPMFvEP-0)

### DIFF
--- a/carbon-projects/lib/fileCategories.ts
+++ b/carbon-projects/lib/fileCategories.ts
@@ -1,9 +1,5 @@
 export const fileCategories = [
-  { title: "Methodology", value: "Methodology" },
-  {
-    title: "Project Description Document (PDD)",
-    value: "Project Description Document (PDD)",
-  },
-  { title: "Verifications", value: "Verifications" },
+  { title: "Methodology and PDD", value: "Methodology and PDD" },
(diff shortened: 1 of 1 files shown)


0xAurelius:
So my concern with lumping Methodology and PDD this way is that methodology documents are typically (in the existing external registry projects) associated with the Methodology entity itself in the CMS as an external file URL, rather than the individual project as a hosted file

we can reupload the methodology document on the individual project as well, but I though it would be "cleaner" to have this be a separate category so that way if a given project uses a standard methodology, whose document is linked from the Methodology section of the project page, the user isn't confused why there is a Documents section called "PDD and Methodology" but then it doesn't contain a Methodology document - and similarly if there is a unique project-specific methodology doc, then it gets the `Methodology` category shows up under `Documents`

I'm also concerned that `PDD` is jargon and some users may not know what the acronym stands for...

My apologies for not communicating better about these minor issues, let's discuss this

0xAurelius (approved):

PR #2434, opened 2025-10-24
Adds CarbonClass and CarbonClass group schemas

biwano:
## Description

Adds the Carbon Class and Carbon Class Croup schemas to the sanity CMS.

Those entities would be queried by the dApp to display supplemental information about Carbon Classes.

###  A Carbon Claass Group:
image

### A Carbon Class
image

## Related Ticket

NA

## How to Test

Done on a personal project. I can deploy and give access if necessary.

### DIFF
--- a/carbon-projects/package.json
+++ b/carbon-projects/package.json
@@ -5,7 +5,7 @@
   "main": "package.json",
   "license": "UNLICENSED",
   "scripts": {
-    "dev": "sanity dev",
+    "dev": "sanity dev --host 0.0.0.0",
     "start": "sanity start",
     "build": "sanity build",
     "deploy": "sanity deploy",
--- a/carbon-projects/schemas/carbonClass.ts
+++ b/carbon-projects/schemas/carbonClass.ts
@@ -1,0 +1,50 @@
+import { defineField, defineType } from "sanity";
+
+export default defineType({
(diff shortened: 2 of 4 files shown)


Atmosfearful:
At this stage I would rather not tightly couple the new dapp to sanity, since it's just for mapping labels to addresses and we don't have a more complex use case or a lot of classes to manage yet...
- can the label be fetched from on-chain? (stretch goal for abhijeet?)
- can we stick with a hardcoded JSON in the FE?

jabby09:
> At this stage I would rather not tightly couple the new dapp to sanity, since it's just for mapping labels to addresses and we don't have a more complex use case or a lot of classes to manage yet...
> 
> * can the label be fetched from on-chain? (stretch goal for abhijeet?)
> * can we stick with a hardcoded JSON in the FE?

biwano:
> At this stage I would rather not tightly couple the new dapp to sanity, since it's just for mapping labels to addresses and we don't have a more complex use case or a lot of classes to manage yet...
> 
> * can the label be fetched from on-chain? (stretch goal for abhijeet?)
cc @must-be-carbon
> * can we stick with a hardcoded JSON in the FE?
Yes. We can hardcode. Freezing this now.

